PR #2, opened 2026-04-07
fix(fast-checker): remove stdout.log size check from isAgentActive() — fixes permanent typing indicator

grandamenium:
Fixes #219

## Problem

`isAgentActive()` had two signal paths:

1. **stdout.log file size growth** — if log grew between polls, return true immediately
2. **Hook-based** — use `lastMessageInjectedAt` vs `last_idle.flag` (Stop hook)

Claude Code writes ANSI escape sequences (spinner, cursor movement, status line) to stdout constantly, even when sitting idle at the prompt. This means stdout.log always grows, path 1 always returns true, and users see a permanent "typing..." indicator on every agent 24/7.

## Fix

Remove the stdout.log size check entirely. The hook-based path is correct and accurate:

- `lastMessageInjectedAt`: set when fast-checker injects a Telegram message
- `last_idle.flag`: written by the Stop hook when Claude finishes its turn
- Result: typing indicator shows only while Claude is responding to a message, clears the moment it finishes

The Stop hook (`hook-idle-flag.ts`) is confirmed working — `last_idle.flag` is being written correctly.

## Impact

- Typing indicator now shows only when agent is actively responding
- Clears immediately when Claude's turn ends (Stop hook fires)
- No false positives from Claude Code's internal spinner output

### DIFF
--- a/src/daemon/agent-manager.ts
+++ b/src/daemon/agent-manager.ts
@@ -19,6 +19,9 @@ import { processMediaMessage } from '../telegram/media.js';
 export class AgentManager {
   private agents: Map<string, { process: AgentProcess; checker: FastChecker; poller?: TelegramPoller }> = new Map();
   private workers: Map<string, WorkerProcess> = new Map();
+  // Tracks agents that received a start request while still stopping.
+  // stopAgent() honors these after cleanup completes so restart-all is race-free.
+  private pendingRestarts: Set<string> = new Set();
   private instanceId: string;
   private ctxRoot: string;
   private frameworkRoot: string;
@@ -50,7 +53,10 @@ export class AgentManager {
    */
   async startAgent(name: string, agentDir: string, config?: AgentConfig): Promise<void> {
     if (this.agents.has(name)) {
-      console.log(`[agent-manager] Agent ${name} already running`);
+      // Agent is registered but may be mid-stop (race: restart-all sends stop+start simultaneously).
+      // Queue the start so stopAgent() will re-launch it once cleanup finishes.
+      this.pendingRestarts.add(name);
+      console.log(`[agent-manager] Agent ${name} is stopping — queued restart`);
       return;
     }
 
@@ -306,6 +312,15 @@ export class AgentManager {
     entry.checker.stop();
     await entry.process.stop();
     this.agents.delete(name);
+
+    // Honor any restart that was queued while we were stopping.
+    if (this.pendingRestarts.has(name)) {
+      this.pendingRestarts.delete(name);
+      console.log(`[agent-manager] Honoring queued restart for ${name}`);
+      this.startAgent(name, '').catch(err =>
+        console.error(`[agent-manager] Queued restart failed for ${name}:`, err),
+      );
+    }
   }
 
   /**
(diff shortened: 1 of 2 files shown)


johnreitano:
## Mode-4 added — plus a near-miss worth recording

Two commits, and **the order between them is load-bearing**: `6cb49a1` (heartbeats create the paths they name) must land before `ee546be` (missing source = exit 1). Reversed, this is a fleet outage.

### What Mode-4 fixes
`kb-ingest` named a file, printed `NOT FOUND: <path>`, then printed `Done!` and exited 0. Ingesting *only* a missing file was indistinguishable from a real ingest. The signal existed and died in the rendering — the summary contradicted the body, and the exit code sided with the summary.

Fixed in `mmrag.py`, not the TS wrapper, because `bus/kb-ingest.sh` calls the python directly and would otherwise keep the old silent behaviour.

### The near-miss (this is the part I'd want reviewed)
The gate I inherited was "every heartbeat naming `learnings.md` has it" — 15 naming, 0 violations, re-verified. **That gate was necessary but not sufficient**, and I nearly shipped on it.

Step 10 also ingests `./memory/$(date -u +%Y-%m-%d).md` — *today's* memory file, which does not exist until today's first write. It is named-but-legitimately-absent **by design**, and no static check catches it because the violation only exists at certain times of day. Checked live at 00:19 UTC: **3 of 4 leadio agents had no file for the current UTC day.** Strictness alone would have failed their next heartbeat.

Worse, it was briefly live: `mmrag.py` is not compiled, so the change took effect the moment I saved it. I reverted, fixed the heartbeats, then re-applied in the correct order.

The failure mode it would have caused is the ironic one: a guard that cries wolf every morning trains everyone to ignore it, destroying the signal it exists to provide.

### Scope widened deliberately (flagging, not sneaking)
`errors` had the identical defect — counted, then discarded at the exit code. An ingest that hit exceptions now also exits 1. **This is a live behaviour change**: a transient Gemini failure surviving retry will now surface as a failed heartbeat ingest rather than a green one. I believe that's correct (it's true information), but it's beyond the task as written — say the word and I'll narrow it to NOT-FOUND only.

### Verification
- Missing only → exit 1; mixed real+missing → exit 1 **and the real source still ingests** (2 of 3, not all-or-nothing); all present → exit 0; directory arg → exit 0.
- Verbatim replay of a live agent's Step 10 with no daily memory file → exit 0, `3 of 3 named source(s)`.
- Empty file (what `touch` creates) → `SKIP (empty)`, exit 0 — the assumption the touch fix rests on.
- 5 new unit tests; mutation-tested (2 fail against the old code, so they aren't vacuous).
- Full suite: 1985 passed. The 33 failures are **pre-existing** — proven by stashing my changes and reproducing them identically on a clean tree (hooks/crons/upgrade-cron-teaching, untouched here).

### Separate findings, not fixed here
1. `bus/kb-ingest.sh` is a **second, divergent entry point**: it genuinely supports `--collection` while the TS CLI now rejects it as never having existed. Its header still advertises the flag. Same name, two contracts.
2. `community/agents/agentic-crm-assistant/HEARTBEAT.md` still has **no ingest step at all** (Mode 2) — the one file the sweep couldn't match, and unguardable by construction.
3. The `orgs/` heartbeat fixes are **unversioned** (`.gitignore:13`) — live on disk, no version record. Same gap already flagged on this PR.

Still **yours to merge** — framework repo, fleet blast radius.

PR #8, opened 2026-04-08
fix(daemon): respect enabled-agents.json in discoverAndStart + listAgents

grandamenium:
## Summary

The daemon's agent discovery loop and the CLI's `list-agents` command were reading from completely different sources of truth, so user-initiated `enable`/`disable` operations had no effect across daemon restarts and `cortextos list-agents` could show a fictional view that didn't match what the daemon was actually running.

This PR aligns both sides on the same logical model with two surgical changes (~70 lines net).

## Root cause

There were TWO parallel "agent enabled" registries that didn't talk to each other:

1. **Per-agent `config.json`** at `${frameworkRoot}/orgs/${org}/agents/${name}/config.json` with `enabled` field — the daemon respects this (`agent-manager.ts:43`)
2. **Instance-level `enabled-agents.json`** at `${ctxRoot}/config/enabled-agents.json` — the CLI's `enable`/`disable` commands, the dashboard's lifecycle API, `list-agents`, `metrics`, and various bus subcommands all read/write this file

The daemon completely ignored `enabled-agents.json` (verified: `grep -rn "enabled-agents" src/daemon/` returns ONLY references to its own internal `discoverAgents` function).

Symmetrically, `bus/agents.ts:listAgents` treated `enabled-agents.json` as authoritative — if the file existed, the directory scan was skipped entirely. This caused `cortextos list-agents` to miss any agent that the daemon had discovered but that wasn't in the file.

## Repro on plain `main` (`6421635`)

```bash
curl -fsSL https://raw.githubusercontent.com/grandamenium/cortextos/main/install.mjs | node
# ... add an agent dir to ~/cortextos/orgs/<org>/agents/<name> ...
cortextos start
# verify the daemon discovers and runs the agent
cortextos disable <name>
# the file is updated, IPC stop sent
pm2 restart cortextos-daemon
# the agent is RUNNING AGAIN — daemon ignored the disable
```

### Change 1 — `src/daemon/agent-manager.ts`

`discoverAndStart` now also reads `${ctxRoot}/config/enabled-agents.json` and respects `enabled: false` entries. The existing per-agent `config.enabled === false` check is preserved unchanged. A new private helper `readInstanceEnableList()` encapsulates the file read with safe fallback on missing/corrupt file.

### Change 2 — `src/bus/agents.ts:listAgents`

[...]

PR #7, opened 2026-04-07
fix(install): chmod 600 the instance .env file

grandamenium:
## Summary

The instance `.env` file written by `cortextos install` (at `~/.cortextos/<instance>/.env`) is being created with the default mode (644 — world-readable). The parent directory is 700 so cross-user access is blocked, but `.env` files conventionally hold secrets, and any future feature writing a token or credential there would silently expose it to any local process running as the same user.

## Repro (on plain `main`)

```bash
curl -fsSL https://raw.githubusercontent.com/grandamenium/cortextos/main/install.mjs | node
stat -f '%Mp%Lp' ~/.cortextos/default/.env
# → 0644 (current behavior)
```

## Fix

One-line addition mirroring the existing pattern at `src/cli/install.ts:269` (`bus-signing-key`) and `:306` (`dashboard.env`):

```ts
try { chmodSync(envPath, 0o600); } catch { /* ignore on Windows */ }
```

`chmodSync` was already imported at the top of the file. The chmod is inside the `if (!existsSync(envPath))` block so existing `.env` files are unaffected — only new installs get the fix.

## Test plan

- [x] `npm run build` — clean
- [x] `npm test` — 382/382 passed
- [ ] Reproduce bug on plain main: `curl ... main ... | node && stat -f '%Mp%Lp' ~/.cortextos/default/.env` → expect `0644`
- [ ] Apply fix, reinstall, verify: `stat -f '%Mp%Lp' ~/.cortextos/default/.env` → expect `0600`

## Context

Found during a fresh-install audit on a clean macOS install. Filed internally as `BUG-001`. First in a series of small upstream fixes coming out of a comprehensive cortextOS production-readiness sweep.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #9, opened 2026-04-08
feat(install): add CORTEXTOS_BRANCH env var for testing pre-merge fixes

grandamenium:
## Summary

Add a `CORTEXTOS_BRANCH` env var to `install.mjs` so we can test pre-merge fixes via the canonical curl install path. ~10 lines, fully backward compatible.

## Why

The canonical install command is:
```bash
curl -fsSL https://raw.githubusercontent.com/grandamenium/cortextos/main/install.mjs | node
```

The curl URL fetches `install.mjs` from a specific branch, but `install.mjs` itself unconditionally clones `main` (line 392 today). So if you want to test a fix in `fix/some-branch` via curl:

- Option A — clone `main`, get the unfixed code, defeats the purpose
- Option B — manually `git clone --branch fix/some-branch && node ~/cortextos/install.mjs`, which works but isn't the "real user experience" the curl command represents
- Option C — wait until your fix is merged to test via curl, which means publishing unverified fixes

This PR adds option D:

```bash
CORTEXTOS_BRANCH=fix/some-branch \
  curl -fsSL https://raw.githubusercontent.com/grandamenium/cortextos/fix/some-branch/install.mjs | node
```

Both URL components point at the same branch — `install.mjs` is fetched FROM that branch (containing this fix) and then clones FROM that branch via `git clone --branch ${REPO_BRANCH}`. Self-bootstrapping.

## Change

Mirrors the existing pattern for `CORTEXTOS_REPO` and `CORTEXTOS_DIR` (lines 15-16):

```js
// CORTEXTOS_BRANCH lets you install a specific branch instead of `main`. Useful
// for testing fixes before they merge:
//   CORTEXTOS_BRANCH=fix/foo curl -fsSL .../fix/foo/install.mjs | node
// Branch name is restricted to standard git ref characters to avoid shell injection.
const REPO_BRANCH_RAW = process.env.CORTEXTOS_BRANCH || 'main';
if (!/^[a-zA-Z0-9._/-]+$/.test(REPO_BRANCH_RAW)) {
  console.error(`Invalid CORTEXTOS_BRANCH value: ${REPO_BRANCH_RAW}`);
  console.error('Branch names may only contain letters, digits, dot, underscore, slash, or dash.');
  process.exit(1);
}
const REPO_BRANCH = REPO_BRANCH_RAW;
```

And one line in the clone block:
```js
runVisible(`git clone --branch ${REPO_BRANCH} ${REPO_URL} ${JSON.stringify(INSTALL_DIR)}`);
```

## Backward compatibility

- `CORTEXTOS_BRANCH` unset → defaults to `main` → identical behavior to today
- `CORTEXTOS_BRANCH=main` → explicitly clones main → identical behavior
- Branch name validation: shell-injection-safe regex (matches git's own ref naming rules conservatively)
- The existing-install path (`git pull upstream main --ff-only`) is intentionally NOT touched in this PR — it still pulls from `main`. If you want to switch an existing install to a branch, manually delete `~/cortextos` first, then re-run with `CORTEXTOS_BRANCH=foo`.

## Test plan

- [x] `npm test` — 389/389 passing (install.mjs has no unit tests; this just confirms nothing else broke)
- [ ] Manual: `./reset.sh && CORTEXTOS_BRANCH=fix/install-branch-flag curl -fsSL https://raw.githubusercontent.com/grandamenium/cortextos/fix/install-branch-flag/install.mjs | node` → verify install lands on commit 5a043b1
- [ ] Sanity: `./reset.sh && curl -fsSL https://raw.githubusercontent.com/grandamenium/cortextos/main/install.mjs | node` (canonical, unchanged) → still works

Will follow up with proof comments after testing.

### DIFF
--- a/install.mjs
+++ b/install.mjs
@@ -14,6 +14,18 @@ import { homedir, platform } from 'os';
 
 const REPO_URL = process.env.CORTEXTOS_REPO || 'https://github.com/grandamenium/cortextos.git';
 const INSTALL_DIR = process.env.CORTEXTOS_DIR || join(homedir(), 'cortextos');
+
+// CORTEXTOS_BRANCH lets you install a specific branch instead of `main`. Useful
+// for testing fixes before they merge:
+//   CORTEXTOS_BRANCH=fix/foo curl -fsSL .../fix/foo/install.mjs | node
+// Branch name is restricted to standard git ref characters to avoid shell injection.
+const REPO_BRANCH_RAW = process.env.CORTEXTOS_BRANCH || 'main';
+if (!/^[a-zA-Z0-9._/-]+$/.test(REPO_BRANCH_RAW)) {
+  console.error(`Invalid CORTEXTOS_BRANCH value: ${REPO_BRANCH_RAW}`);
+  console.error('Branch names may only contain letters, digits, dot, underscore, slash, or dash.');
+  process.exit(1);
+}
+const REPO_BRANCH = REPO_BRANCH_RAW;
 const IS_WINDOWS = platform() === 'win32';
 const IS_MAC = platform() === 'darwin';
 const IS_LINUX = platform() === 'linux';
@@ -389,8 +401,8 @@ if (existsSync(INSTALL_DIR)) {
     fail(`${INSTALL_DIR} exists but is not a git repo. Remove it or set CORTEXTOS_DIR to a different path.`);
   }
 } else {
-  log(`Cloning cortextOS to ${INSTALL_DIR}...`);
-  runVisible(`git clone ${REPO_URL} ${JSON.stringify(INSTALL_DIR)}`);
+  log(`Cloning cortextOS (branch: ${REPO_BRANCH}) to ${INSTALL_DIR}...`);
+  runVisible(`git clone --branch ${REPO_BRANCH} ${REPO_URL} ${JSON.stringify(INSTALL_DIR)}`);
   ok('Cloned');
 
   // Rename origin → upstream so check-upstream and upstream-sync work out of the box


grandamenium:
## ✅ Self-bootstrapping verification

Tested by running the new env var on a fully cleaned system:

\`\`\`bash
./reset.sh --yes
curl -fsSL https://raw.githubusercontent.com/grandamenium/cortextos/fix/install-branch-flag/install.mjs \
  | CORTEXTOS_BRANCH=fix/install-branch-flag node
\`\`\`

Result:

\`\`\`
$ git -C ~/cortextos log --oneline -3
5a043b1 feat(install): add CORTEXTOS_BRANCH env var for testing pre-merge fixes
159907c fix(daemon): respect enabled-agents.json in discoverAndStart + listAgents (#8)
6421635 fix(install): chmod 600 the instance .env file (#7)

$ git -C ~/cortextos branch --show-current
fix/install-branch-flag

$ grep -n "CORTEXTOS_BRANCH" ~/cortextos/install.mjs
18:// CORTEXTOS_BRANCH lets you install a specific branch instead of `main`. Useful
20://   CORTEXTOS_BRANCH=fix/foo curl -fsSL .../fix/foo/install.mjs | node
22:const REPO_BRANCH_RAW = process.env.CORTEXTOS_BRANCH || 'main';
\`\`\`

The clone landed on `5a043b1` (this branch's tip), not `159907c` (main's tip). Self-bootstrapping confirmed: install.mjs fetched from the branch URL → new code reads CORTEXTOS_BRANCH from env → clones the branch directly.

## ⚠️ Documentation gotcha discovered

Bash pipe semantics trap: when documenting this command, **do NOT** use the inline-prefix form:

\`\`\`bash
# WRONG — env var is set for curl, not node
CORTEXTOS_BRANCH=foo curl ... | node
\`\`\`

In a pipeline, env var assignments only apply to the immediate command. The `node` process on the right side of the pipe gets a fresh environment without the var. Always use one of:

\`\`\`bash
# ✅ env var on the right side of the pipe (shortest)
curl ... | CORTEXTOS_BRANCH=foo node

# ✅ export first (more readable)
export CORTEXTOS_BRANCH=foo
curl ... | node
\`\`\`

Worth adding a usage example to the README so users don't hit this trap. Will follow up in a doc-only PR if you want.

Ready to merge.

PR #5, opened 2026-04-07
fix(cli): require --all to stop every agent (foot-gun fix)

grandamenium:
## Summary

`cortextos stop` with no arguments silently iterated `list-agents` and stopped every running agent in sequence, taking down the entire fleet. The command's description claimed it would "stop the daemon," which is doubly wrong: it did not stop the daemon, and it did stop every agent. This caused a real cascade in production: an autonomous orchestrator agent ran `cortextos stop` expecting to stop the daemon process and instead killed all six running agents within seconds.

This PR makes the dangerous behavior require an explicit `--all` flag and rewrites the help text to be accurate.

## Behavior before

```
$ cortextos stop
Stopping all agents...
  alice: stopped
  bob: stopped
  carol: stopped
  ...
```

The entire fleet went down. The daemon kept running. Description said "Stop the daemon or a specific agent."

## Behavior after

```
$ cortextos stop
Refusing to stop all agents without an explicit target.

  To stop one agent:    cortextos stop <agent>
  To stop every agent:  cortextos stop --all
  To stop the daemon:   pm2 stop cortextos-daemon

(Previously `cortextos stop` with no argument silently stopped every running agent. That behavior was a foot-gun and now requires --all.)
$ echo $?
2
```

- `cortextos stop <agent>` — unchanged, still stops one agent.
- `cortextos stop --all` — new, explicit opt-in to fleet-wide stop.
- `cortextos stop` — now a hard error with a helpful message; exits 2; performs no IPC.
- `cortextos stop <agent> --all` — rejected as ambiguous; exits 2.
- Description and `[agent]` help text rewritten to accurately describe behavior and explicitly disclaim that this command does not stop the daemon process.

## Test plan

No existing tests cover `src/cli/*.ts`, so this PR follows the project precedent of manual verification:

- [ ] `cortextos stop` (no args) prints the safety error and exits 2; no agent is stopped.
- [ ] `cortextos stop --all` stops every running agent and exits 0; daemon still running.
- [ ] `cortextos stop <agent>` still stops the named agent and exits 0.
- [ ] `cortextos stop <agent> --all` errors out with the mutual-exclusion message and exits 2.
- [ ] `cortextos stop --help` shows the `--all` flag and the `pm2 stop cortextos-daemon` guidance.

## Out of scope

- Cleaning up the corrupt `enabled-agents.json` entry (separate task).
- Adding source logging to the IPC server's `stop-agent` handler (separate task).
- Touching v1 LaunchAgents (separate task).
[...]

PR #4, opened 2026-04-07
fix(daemon): rebuild Telegram poller on restartAgent

grandamenium:
## Summary
`restartAgent()` in `src/daemon/agent-manager.ts` stopped and restarted the agent process and fast checker, but never touched the `TelegramPoller`. After every IPC `restart-agent` call the old poller kept running in stale state forever and no new poller was ever created — agents silently stopped receiving Telegram messages until the daemon itself restarted.

## Fix
Delegate to `stopAgent` + `startAgent('')` so every per-agent resource is rebuilt from fresh `.env` credentials on restart:
- `AgentProcess`
- `FastChecker` (incidentally picks up any credential changes)
- `TelegramAPI` and `TelegramPoller`
- `AgentProcess.onStatusChanged` crash-notification callback
- Slash-command registration

`agentDir` is auto-discovered by `startAgent` from `frameworkRoot/orgs/{org}/agents/{name}`, which is the same path used for queued restarts in `stopAgent` (line 320). The new implementation participates in the `pendingRestarts` race protection added in #1 for free.

## Scope
Single-method rewrite. No helpers, no new tests, no signature changes, no telemetry beyond two log lines.

## Explicitly out of scope
- `FastChecker` credential staleness (incidentally fixed, not expanded)
- SIGUSR1 double-registration window in `FastChecker.stop`/`start` (sidestepped by instance replacement)
- `TelegramPoller.getUpdates` AbortController (separate known issue)
- Dashboard passing `mode` to restart-agent IPC (ignored, separate issue)

## Test plan
- [x] `npx tsc --noEmit` clean
- [x] `npm run build` clean
- [x] `npm test` — all existing tests pass
- [ ] End-to-end: fresh-boot a live agent, send it a Telegram message (baseline), `cortextos bus restart <agent>`, send a second Telegram message — agent must reply. This is executed by the orchestrator after review.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #11, opened 2026-04-08
fix(daemon): close PTY race in AgentProcess.stop() (BUG-011)

grandamenium:
## Summary

`AgentProcess.stop()` had a race window where the PTY's exit handler could fire AFTER `stopping=false` was set, falling into the crash recovery branch and resurrecting an agent we just intentionally stopped. This PR closes the window by making `stop()` await the actual PTY exit before resolving, and also delegates `sessionRefresh()` to `stop() + start()` (eliminating an additional duplicate-pty bug).

## The race (verified in source)

`src/daemon/agent-process.ts:106-137` (before this PR):

```ts
async stop(): Promise<void> {
  if (this.stopping) return;
  this.stopping = true;          // race window opens
  // ... write Ctrl-C, /exit, sleep, kill ...
  this.stopping = false;          // race window closes — but exit may not have fired yet
  this.status = 'stopped';
  // ...
}
```

The PTY's `kill()` is asynchronous on macOS. The exit callback (set in `start()` at line 81) fires whenever the underlying process actually dies — could be before, during, or AFTER `stop()` returns. The exit handler at line 250-256 has a guard:

```ts
private handleExit(exitCode: number): void {
  this.pty = null;
  this.clearSessionTimer();
  if (this.stopping) return;    // ← this check is the race
  // ... crash recovery: increment crashCount, status='crashed', schedule restart ...
}
```

If the exit fires AFTER `stop()` set `stopping=false`, the guard fails and the function falls through to crash recovery. Result: an agent we *intentionally* stopped gets resurrected ~5-10 seconds later under the wrong assumption that it crashed.

## Empirical evidence

PR #4's cycle 2 daemon log captured the AgentManager-level `pendingRestarts` race protection firing TWICE during a single restart sequence:

```
[agent-manager] Restarting testbot
[testbot] Stopping...
[testbot] Exited with code 0 signal 0
[testbot] Stopped
[agent-manager] Honoring queued restart for testbot          ← race window 1
[testbot] Starting in continue mode
[testbot] Running (pid: ...)
[agent-manager] Agent testbot is stopping — queued restart   ← race window 2
[agent-manager] Restart complete for testbot
```

Two `startAgent` calls effectively raced through the same restart. The end state was correct that time, but the underlying race is real and non-deterministic. The `pendingRestarts` mechanism in AgentManager was a workaround for the underlying race fixed in this PR.

### Change 1 — `stop()` awaits the PTY exit before resolving

Add `exitPromise` and `resolveExit` fields. `start()` creates a fresh promise; the existing `onExit` handler resolves it AFTER calling `handleExit`. `stop()` awaits the promise (with a 5-second safety timeout for hung PTYs) AFTER `pty.kill()` and BEFORE `stopping=false`.

```ts
async start(): Promise<void> {
  // ... existing setup ...
  this.exitPromise = new Promise<void>((resolve) => {
    this.resolveExit = resolve;
  });
  this.pty.onExit((exitCode, signal) => {
    this.log(`Exited with code ${exitCode} signal ${signal}`);
    this.handleExit(exitCode);
    this.resolveExit?.();        // signal anyone awaiting the exit
    this.resolveExit = null;
  });
}

async stop(): Promise<void> {
  if (this.stopping) return;
  this.stopping = true;
  // ... clearSessionTimer, capture pty, write Ctrl-C, kill ...
  if (exitPromise) {
    await Promise.race([exitPromise, sleep(5000)]);  // BUG-011 fix
  }
  this.stopping = false;
  // ...
}
```

By the time `stopping=false` runs, the exit handler has already fired and seen `stopping=true`, so it skipped crash recovery. Race closed.

### Change 2 — `sessionRefresh()` delegates to `stop() + start()`

The previous inline implementation duplicated the stop logic AND had a separate bug where the OLD pty's exit handler could fire AFTER the NEW pty was set up, nulling out the wrong reference. The cleanest fix is the same lesson PR #4 applied to `restartAgent`:

```ts
async sessionRefresh(): Promise<void> {
  this.log('Session refresh (--continue restart)');
  await this.stop();
  await this.start();  // shouldContinue() returns true → continue mode
  this.log('Session refreshed');
}
```

Inherits the BUG-011 fix automatically AND eliminates the duplicate-pty bug for free.

## Tests

- [x] `npm test`: **395/395 passing** (391 baseline + 4 new in `tests/unit/daemon/agent-process.test.ts`)
- [x] New tests:
  - `stop() awaits the PTY exit handler before resolving`
  - `stop() does NOT trigger crash recovery on intentional stop` ← regression test for the bug
  - `handleExit DOES trigger crash recovery on UNINTENTIONAL exit` ← regression check, make sure real crash recovery still works
  - `sessionRefresh() delegates to stop() then start()` ← pins the new delegation
- [ ] Manual: 3-cycle test (reset → main without fix → reset → branch with fix → reset → main with merge), follow up in comments

## What this PR does NOT change

- The `stopping` flag's existence or its semantics — the only change is WHEN it gets reset (after the exit fires, not before)
- The `handleExit` function — left untouched, just gets `resolveExit` called AFTER it
- The crash recovery branch — left untouched, regression test pins it in place
- The AgentManager-level `pendingRestarts` logic — separate concern. With BUG-011 fixed, we can revisit whether `pendingRestarts` is still needed (or just dead code paying the price for the absent BUG-011 fix). Will track as BUG-031 if it doesn't go away.

## Refs

- BUG-011 in our internal tracker (the PTY race itself)
- Likely also closes BUG-010 (boris2 SIGHUP code 129 on restart) — that was the symptom of spurious crash recovery firing on an intentional stop, then `start()` running into a half-cleaned-up PTY state
- Sixth PR in tonight's stability sweep (#7, #8, #9, #10, #4, #5, this one)

### DIFF
--- a/src/daemon/agent-process.ts
+++ b/src/daemon/agent-process.ts
@@ -26,6 +26,12 @@ export class AgentProcess {
   private sessionStart: Date | null = null;
   private status: AgentStatus['status'] = 'stopped';
   private stopping: boolean = false;
+  // BUG-011 fix: stop() awaits this promise (resolved by the onExit handler in start())
+  // to guarantee the PTY exit has fired before stopping=false is reset. Without
+  // this, the exit handler can fire after stopping=false and trigger spurious
+  // crash recovery for an agent we just stopped intentionally.
+  private exitPromise: Promise<void> | null = null;
+  private resolveExit: (() => void) | null = null;
   private dedup: MessageDedup;
   private log: LogFn;
   private onStatusChange: ((status: AgentStatus) => void) | null = null;
@@ -77,10 +83,20 @@ export class AgentProcess {
     this.log(`Log path: ${logPath}`);
     this.pty = new AgentPTY(this.env, this.config, logPath);
 
+    // BUG-011 fix: create a fresh exit signal for this run. resolveExit is
+    // called from the onExit handler below; stop() awaits exitPromise to
+    // guarantee the exit handler has fired before clearing stopping.
+    this.exitPromise = new Promise<void>((resolve) => {
+      this.resolveExit = resolve;
+    });
+
     // Handle exit
     this.pty.onExit((exitCode, signal) => {
       this.log(`Exited with code ${exitCode} signal ${signal}`);
       this.handleExit(exitCode);
+      // Signal anyone awaiting this PTY's exit (e.g. stop() — BUG-011 fix)
+      this.resolveExit?.();
+      this.resolveExit = null;
     });
 
     try {
@@ -113,6 +129,9 @@ export class AgentProcess {
     // shutdown doesn't race with us and trigger crash recovery or a double-kill.
     const pty = this.pty;
     this.pty = null;
+    // Capture the exit promise before any awaits — we'll wait on this AFTER
+    // pty.kill() to guarantee the exit handler has run before stopping=false.
+    const exitPromise = this.exitPromise;
 
     if (pty) {
       try {
@@ -128,6 +147,15 @@ export class AgentProcess {
       } catch {
         // PTY may have already exited — ignore
       }
+
+      // BUG-011 fix: AWAIT the exit handler before resolving stop().
+      // Without this, stop() returns while the PTY exit is still in flight,
+      // and the exit handler may fire AFTER stopping=false (set below),
+      // triggering spurious crash recovery for an agent we just stopped.
+      // 5-second timeout guards against a hung PTY.
+      if (exitPromise) {
+        await Promise.race([exitPromise, sleep(5000)]);
+      }
     }
 
     this.stopping = false;
@@ -138,45 +166,18 @@ export class AgentProcess {
 
   /**
    * Restart with --continue (session refresh).
+   *
+   * Delegates to stop() + start() so it inherits the BUG-011 race fix
+   * automatically. This also eliminates a separate bug in the previous
+   * inline implementation where the OLD pty's exit handler could fire
+   * AFTER the NEW pty was set up, nulling out the wrong reference.
+   * `start()` will pick up `continue` mode automatically because the
+   * conversation directory still has .jsonl files (shouldContinue() is true).
    */
   async sessionRefresh(): Promise<void> {
     this.log('Session refresh (--continue restart)');
-    this.stopping = true;
-    this.clearSessionTimer();
-
-    // Capture and null out pty before awaits (same race-condition guard as stop()).
-    const pty = this.pty;
-    this.pty = null;
-
-    if (pty) {
-      try {
-        pty.write('\x03'); // Ctrl-C
-        await sleep(1000);
-        pty.write('/exit\r');
-        await sleep(3000);
-      } catch {
-        // Ignore
-      }
-      try {
-        pty.kill();
-      } catch {
-        // PTY may have already exited — ignore
-      }
-    }
-
-    this.stopping = false;
-
-    // Relaunch with continue mode
-    const logPath = join(this.env.ctxRoot, 'logs', this.name, 'stdout.log');
-    this.pty = new AgentPTY(this.env, this.config, logPath);
-
-    this.pty.onExit((exitCode) => {
-      this.handleExit(exitCode);
-    });
-
-    const prompt = this.buildContinuePrompt();
-    await this.pty.spawn('continue', prompt);
-    this.startSessionTimer();
+    await this.stop();
+    await this.start();
     this.log('Session refreshed');
   }
 
--- a/tests/unit/daemon/agent-process.test.ts
+++ b/tests/unit/daemon/agent-process.test.ts
@@ -1,0 +1,146 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+
+// Capture the PTY exit handler so tests can simulate exits at controlled times
+let capturedOnExit: ((exitCode: number, signal?: number) => void) | null = null;
+
+const mockPty = {
+  spawn: vi.fn().mockResolvedValue(undefined),
+  kill: vi.fn(),
+  write: vi.fn(),
+  getPid: vi.fn().mockReturnValue(12345),
+  onExit: vi.fn().mockImplementation((cb: (exitCode: number, signal?: number) => void) => {
+    capturedOnExit = cb;
+  }),
+};
+
+vi.mock('../../../src/pty/agent-pty.js', () => ({
+  AgentPTY: function AgentPTY() { return mockPty; },
+}));
+
+vi.mock('../../../src/pty/inject.js', () => ({
+  injectMessage: vi.fn(),
+  MessageDedup: class { isDuplicate() { return false; } },
+}));
+
+vi.mock('../../../src/utils/atomic.js', () => ({
+  ensureDir: vi.fn(),
+  atomicWriteSync: vi.fn(),
+}));
+
+vi.mock('../../../src/utils/env.js', () => ({
+  writeCortextosEnv: vi.fn(),
+  resolveEnv: vi.fn().mockReturnValue({ instanceId: 'test', ctxRoot: '/tmp/test' }),
+}));
+
+vi.mock('../../../src/bus/reminders.js', () => ({
+  getOverdueReminders: vi.fn().mockReturnValue([]),
+}));
+
+vi.mock('../../../src/utils/paths.js', () => ({
+  resolvePaths: vi.fn().mockReturnValue({}),
+}));
+
+vi.mock('fs', async () => {
+  const actual = await vi.importActual<typeof import('fs')>('fs');
+  return {
+    ...actual,
+    mkdirSync: vi.fn(),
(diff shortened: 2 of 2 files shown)


grandamenium:
## ✅ Verified end-to-end (3-cycle test) — fix works, BUG-011 closed

### Cycle 1 — plain main (\`e3f3b2f\`), bug reproduced

Set up testbot, ran 3 consecutive \`cortextos bus soft-restart-all\` cycles. Daemon log captured the bug live:

\`\`\`
[agent-manager] Restarting testbot
[testbot] Stopping...
[testbot] Stopped
[agent-manager] Honoring queued restart for testbot
[testbot] Starting in continue mode
[testbot] Running (pid: 76225)
[agent-manager] Agent testbot is stopping — queued restart
[testbot] Starting. Waiting for bootstrap...
[agent-manager] Restart complete for testbot
[testbot] Exited with code 129 signal 0       ← exit fires AFTER restart completes
[testbot] Crash recovery: restart in 5s (crash #1)   ← SPURIOUS crash recovery — THE BUG
[agent-manager] Restarting testbot
... (cycle repeats with crash #2 in 10s) ...
\`\`\`

**Race fired 2 times in 3 restart cycles**: 4 distinct testbot PIDs in rapid succession (75992, 76225, 76469, 76479), 2 spurious crash-recovery schedules. The agent we *intentionally* stopped was being resurrected by the crash-recovery branch of \`handleExit\` because the PTY's exit fired AFTER \`stopping=false\` was set.

### Cycle 2 — branch via CORTEXTOS_BRANCH, fix verified

Same 3-cycle scenario. Daemon log:

\`\`\`
[agent-manager] Restarting testbot
[testbot] Stopping...
[testbot] Exited with code 129 signal 0       ← exit fires DURING stop's await window
[testbot] Stopped
[agent-manager] Honoring queued restart for testbot
[testbot] Starting in continue mode
[testbot] Running (pid: 78833)
[agent-manager] Agent testbot is stopping — queued restart
[testbot] Starting. Waiting for bootstrap...
[agent-manager] Restart complete for testbot
\`\`\`

**Note the exit-129 line moved earlier** — it now fires DURING \`stop()\`'s await window (so \`stopping=true\`) instead of AFTER (so \`stopping=false\`). The handleExit guard catches it correctly and skips crash recovery.

### Results

| Metric | Cycle 1 (no fix) | Cycle 2 (with fix) | Verdict |
|---|---|---|---|
| **\`Crash recovery\` scheduled events** | **2** | **0** | ✅ **FIXED — the BUG-011 race is closed** |
| Distinct testbot PIDs in 3 restart cycles | **4** (race chaos) | **3** (one per intentional restart) | ✅ FIXED |
| Exit-129 events | 2 | 2 | unchanged — see below |
| AgentManager pendingRestarts race events | 6 | 6 | unchanged — separate concern, BUG-031 |

### About the exit-129 events

The PTY genuinely exits with code 129 (SIGHUP) — that's how Claude Code dies when sent SIGTERM via \`pty.kill()\`. **The bug was never about the exit code itself; it was about the daemon's REACTION to it.** Before this PR, the exit fired AFTER \`stopping=false\` and triggered crash recovery. With this PR, \`stop()\` awaits the exit before clearing \`stopping\`, so the handler always sees \`stopping=true\` and silently completes the cleanup. The chaotic restart spiral is gone.

If we ever want code 0 instead of 129, that's a separate teardown investigation (probably about how Ctrl-C / \`/exit\` is delivered) — but functionally the fix is complete.

### About the remaining pendingRestarts race

The 6 queued-restart events / 3 honoring events come from the AgentManager-level \`pendingRestarts\` mechanism in \`agent-manager.ts:88, 348-354\`. This was a workaround for the underlying BUG-011 race, and with BUG-011 now fixed, it may be redundant — but removing it is a separate concern (tracked as BUG-031 in our internal log). It doesn't cause incorrect behavior, just extra log noise.

### Test gate

- \`npm test\`: **395/395 passing** (391 baseline + 4 new BUG-011 regression tests)

Ready to merge. **BUG-010 (boris2 SIGHUP code 129 on restart) is also closed by this PR** — that bug was the symptom of the spurious crash recovery firing on a SIGHUP exit, which no longer happens.

grandamenium:
## ✅ Cycle 3: final verification on main

Reset, ran the canonical curl install on main now at \`82e06cc\` (this PR squash-merged). Same 3-restart-cycle scenario:

\`\`\`
=== smoking gun checks ===
Crash recovery scheduled events: 0 (expected: 0)
Restart complete events: 3 (expected: 3)

✓ END-TO-END VERIFIED — BUG-011 fix lands cleanly via canonical curl install
\`\`\`

Sequence summary:
1. Cycle 1 — main without fix → 2 spurious crash recoveries, 4 distinct PIDs in 3 restart cycles, restart spiral chaos
2. Cycle 2 — branch via CORTEXTOS_BRANCH → 0 crash recoveries, 3 PIDs (one per restart), clean
3. Cycle 3 — main with merge → identical to cycle 2, end-to-end verified via canonical curl

**BUG-011 closed. BUG-010 closed (it was the symptom of the same race).** 🎉

The chaotic restart spiral observed in the prior session is now structurally impossible.

InocuousCabbage:
## /local-ultrareview — PR #11 — 2026-04-25

**3 Opus agents ran in parallel (correctness · security/perf · architecture), synthesized, implementation plan produced.**

---

### Overall Assessment

Safe to merge. No critical issues from any reviewer. Core fix motivation is correct — the PTY exit race (`stop()` not awaiting exit before resolving) and `sessionRefresh()` delegation are sound. Biggest concern is a cluster of **8 warnings** around daemon lifecycle re-entrancy gaps (`stop()` re-entrancy, `start()` failure path state leak, `exitPromise`/`resolveExit` asymmetry) — several have already been addressed at HEAD by BUG-040/BUG-032/BUG-048 upstream follow-ups.

**Security reviewers: no issues.** Posture unchanged — no injection surface, no auth changes, no sensitive data.

---

### Critical Issues
**None.**

---

### Warnings (8 — should fix, most partially mitigated at HEAD)

1. **`sessionRefresh()` re-entrancy** — concurrent `stop()` returns immediately on `stopping=true`, leaving `start()` running against a tearing-down PTY. Fix: track `stopPromise: Promise<void> | null`, return in-flight promise on re-entry.
2. **`start()` failure leaves `exitPromise`/`pty` half-initialized** — if PTY spawn throws, `resolveExit` is never called; subsequent `stop()` burns full 15s timeout. Fix: in catch block, call `resolveExit?.()`, null both fields, null `this.pty`.
3. **`sessionRefresh()` swallows `start()` failures** — "Session refreshed" logs even when relaunch crashed. Fix: check `this.status` after `start()` or rethrow.
4. **`exitPromise`/`resolveExit` overwrite race** — old PTY's `onExit` resolves the new run's promise; masked by BUG-040 generation guard but fragile. Fix: capture `resolveExit` in closure at spawn time (`const myResolve = this.resolveExit`). **Flagged by 2 reviewers.**
5. **`onExit` nulls `resolveExit` but not `exitPromise`** — stale resolved promise on instance between lifecycles. **Flagged by 2 reviewers.**
6. **Missing test: timeout-wins-the-race branch** — the actual BUG-040 regression vector is untested. **Flagged by 2 reviewers.**
7. **Missing test: post-timeout late-exit case** — no test for what happens when PTY exits after the 15s timeout arm fires.
8. **`sessionRefresh()` contract not documented** — undocumented dependency on `shouldContinue()` filesystem heuristic; future change could silently turn refreshes into fresh boots.

### Suggestions (8 — encapsulation, magic numbers, test gaps)

- Introduce `Deferred<void>` primitive to replace two-field `exitPromise`/`resolveExit` handshake
- Name `EXIT_AWAIT_TIMEOUT_MS` constant (already bumped to 15s at HEAD)
- Add log-on-timeout when `Promise.race` timeout arm wins
- Replace 3 real-time tests with fake-timer equivalents
- Extract shared `mockPty` fixture to `__fixtures__` module
- Test fresh-Deferred-per-run and start-failure cleanup paths
- Document `shouldContinue()` dependency in `sessionRefresh()` JSDoc
- Clear `this.exitPromise = null` symmetrically with `resolveExit`

---

### Points of Agreement (highest confidence — flagged by 2+ reviewers)
- `exitPromise`/`resolveExit` null asymmetry (W4/W5)
- Missing timeout-wins test (W6/W7)
- 5s timeout magic number needs naming + logging (W8 / S2)
- `sessionRefresh()` contract leakage (W8)

---

### What Looks Good
- Security posture: unchanged, no new attack surface
- `Promise.race` is the correct pattern for the exit-await
- Local-variable capture before `await` is correct
- No performance hot-path concerns

---

### Implementation Estimate
~150–200 LOC across `src/daemon/agent-process.ts` + test file + new `src/utils/deferred.ts`. ~3–4h. Most warnings are partially mitigated at HEAD already; this PR is safe to merge as-is, warnings can land as a follow-up hardening PR.

**Artifacts:** `reviews/11-2026-04-25/` — review-correctness.md · review-security-perf.md · review-architecture.md · synthesis.md · implementation-plan.md

🤖 /local-ultrareview via Claude Code (analyst agent)

PR #14, opened 2026-04-08
fix: close-out batch (BUG-015, 021, 031, 032 + verify-close 008, 009, 026, 027)

grandamenium:
## Summary

Tenth and final PR of tonight's stability sweep. Closes 8 bugs in one PR: 4 small/defensive code fixes plus 4 verify-and-close confirmations of bugs that were transitively fixed by earlier PRs. BUG-003 (mystery SIGTERM cascade) deferred to a future soak session.

| Bug | Sev | Type | What |
|---|---|---|---|
| BUG-015 | P2 | code | IPC source logging — \`source\` field on IPCRequest, daemon logs every call |
| BUG-021 | P2 | code | pm2 startup mid-flow blocking — defer sudo paste to end-of-onboarding optional section |
| BUG-031 | P2 | code | pendingRestarts regression detector — replaces dormant queue with warn lines |
| BUG-032 | P3 | code | PTY SIGHUP code 129 — CRLF + 5s wait (defensive, not aggressive) |
| BUG-008 | P1 | verify | Telegram 409 — likely closed by PR #4, cycle 1 confirms |
| BUG-009 | P2 | verify | "Agent <orgname> not found" — likely closed by PR #8+#10, cycle 1 confirms |
| BUG-026 | P1 | verify | agent-manager path missing org — already correct in source |
| BUG-027 | P2 | verify | dashboard SQLite filename hardcoded — already templated in source |

## Changes (10 files, +86/-26)

- src/types/index.ts — IPCRequest.source field
- src/daemon/ipc-server.ts — log incoming requests with source
- src/cli/{enable-agent,start,stop,status,bus}.ts — instrument all 12 IPC callsites with source field
- src/daemon/agent-manager.ts — replace pendingRestarts queue with regression-warning console.warn lines
- src/daemon/agent-process.ts — CRLF + 5s wait in stop()
- .claude/commands/onboarding.md — defer pm2 startup sudo paste to end-of-flow OPTIONAL section

## Test plan

- [x] npm run build clean
- [x] npm test 414/414 passing (no regressions)
- [ ] Cycle 1 (canonical main): 4 verify-and-close bugs confirmed, 4 code-fix bugs reproduced
- [ ] Cycle 2 (this branch): all 4 fixes verified, all 4 verifications still hold
- [ ] Cycle 3 (canonical merged main): final end-to-end

## Out of scope (deliberate)

- BUG-003 — needs 30-min soak, deferred to future session
[...]

PR #15, opened 2026-04-09
fix(daemon): close PTY exit/stop timing race (BUG-040, closes BUG-038)

grandamenium:
## Summary

Closes BUG-040 (P0) — the actual root cause of BUG-038 (the pendingRestarts regression detector firing under restart-all).

## The bug

PR #11 (BUG-011) introduced \`exitPromise\` so \`stop()\` would await the PTY exit before clearing the \`stopping\` flag. This worked great until PR #14 (BUG-032) bumped the post-/exit wait from 3s to 5s. With the longer wait, total elapsed time before \`pty.kill()\` is ~6s, then the PTY can take additional time to actually exit. If that delayed exit fires AFTER stop()'s 5-second \`Promise.race\` timeout, \`stopping\` is already cleared and \`handleExit\` triggers spurious crash recovery — exactly what PR #11 was supposed to prevent.

**Surfaced live in this session's fastloop test**: a single \`pm2 restart cortextos-daemon\` produced a crash recovery cascade with two crash counts incrementing during what should have been a clean shutdown.

## Fix

1. **\`stopRequested\` flag** that persists across \`stop()\`'s return until \`handleExit\` clears it. Survives the timeout. Cleared either by handleExit consuming an intentional exit, or by \`start()\` beginning a new lifecycle.

2. **\`lifecycleGeneration\` counter** incremented on each \`start()\`. Each PTY's onExit closure captures the generation at spawn time and bails out if it doesn't match — prevents an old PTY's late exit from polluting a new lifecycle.

3. **Bumped Promise.race timeout from 5s to 15s** for less log noise (functional correctness no longer depends on it).

4. **\`handleExit\` now checks \`stopRequested || stopping\`** — either flag short-circuits crash recovery.

## Test plan

- [x] \`npm run build\` clean
- [x] \`npm test\` 414/414 passing (no regressions)
- [x] **Live verification via fastloop** (no full reinstall needed):
  - Daemon SIGTERM (\`pm2 restart cortextos-daemon\`): clean shutdown, zero crash recovery
  - 5× \`cortextos bus soft-restart-all\`: 0 \"Crash recovery\", 0 \"REGRESSION CHECK\", 0 \"Ignoring late exit\", 5× \"Restart complete\"
  - **Telegram screenshot evidence**: 5× correct \"🔄 commander restarted by user\" at 14:23, zero false 🚨 CRASH alarms
- [x] BUG-038 (the symptom) closes as a side effect — pendingRestarts regression detector stays silent

## Closes

- BUG-040 (root cause)
- BUG-038 (symptom — closed as side effect)

## Note on BUG-032

BUG-032 (PTY exits with SIGHUP code 129) is still present — agents still exit with 129 instead of 0 after the graceful shutdown sequence. But with BUG-040 fixed, the daemon correctly recognizes 129 as intentional and does NOT fire crash recovery. BUG-032 is now purely cosmetic log noise and can be addressed in a separate aggressive-fix follow-up if desired.

## Methodology note

This is the **first PR to use the new fast-loop iteration cycle** (\`helper-fastloop.sh\` instead of \`helper-install-and-onboard.sh\`). The bug was identified, the fix written, deployed, and verified in ~15 minutes total — vs the ~30 min cycle a full reinstall would have required. The fix doesn't touch any install-pipeline files (install.mjs, onboarding.md, ecosystem.ts, init.ts), so the fast-loop is sufficient.

### DIFF
--- a/src/daemon/agent-process.ts
+++ b/src/daemon/agent-process.ts
@@ -26,6 +26,22 @@ export class AgentProcess {
   private sessionStart: Date | null = null;
   private status: AgentStatus['status'] = 'stopped';
   private stopping: boolean = false;
+  // BUG-040 fix: persists across stop() return until handleExit clears it.
+  // Required because BUG-032's CRLF + 5s wait can cause graceful shutdown to
+  // exceed the 5s Promise.race timeout in stop(), which would otherwise reset
+  // `stopping=false` BEFORE the PTY actually exits, then handleExit would fire
+  // with stopping=false and trigger spurious crash recovery (a partial regression
+  // of BUG-011). stopRequested survives the timeout and is only cleared either
+  // by handleExit when an intentional exit fires, or by start() at the beginning
+  // of a new lifecycle.
+  private stopRequested: boolean = false;
+  // BUG-040 fix: monotonic generation counter incremented on each successful
+  // start(). Each PTY's onExit closure captures the generation at spawn time
+  // and bails out if the generation doesn't match — i.e. a NEW PTY has been
+  // spawned since this old one was created. Without this guard, a late exit
+  // from an old PTY can race past stopRequested and trigger crash recovery on
+  // the new agent.
+  private lifecycleGeneration: number = 0;
   // BUG-011 fix: stop() awaits this promise (resolved by the onExit handler in start())
   // to guarantee the PTY exit has fired before stopping=false is reset. Without
   // this, the exit handler can fire after stopping=false and trigger spurious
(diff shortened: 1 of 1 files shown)


grandamenium:
## Cycle 3 — verified on canonical merged main ✅

Used `helper-fastloop.sh main` to deploy + restart (~10 sec). Then 5× `cortextos bus soft-restart-all` + 8s sleeps. Filtered to events after the post-merge daemon PID:

- **Crash recovery**: 0
- **REGRESSION CHECK warnings**: 0
- **"Ignoring late exit" (generation guard)**: 0
- **Restart complete**: 5/5
- **Telegram receive verification**: 5× correct "🔄 commander restarted by user: soft-restart-all requested" notifications, zero false 🚨 CRASH alarms (verified visually by user)
- **Exit codes**: 5× code 129 (BUG-032 still cosmetic, but BUG-040 fix correctly handles it as intentional — no crash recovery fires)

BUG-040 fully closed end-to-end. BUG-038 closed as side effect.

This was the **first PR shipped using the new fast-loop iteration** — total time from bug discovery to merged + verified was ~30 minutes, vs the ~90 minutes a full reinstall cycle would have taken.

🤖 Posted via Claude Code

PR #16, opened 2026-04-09
fix(cli): validate agent name in add-agent to match resolveEnv (BUG-041)

grandamenium:
## Summary

BUG-041 (P0) validation mismatch: `cortextos add-agent` accepted any string as an agent name, but `resolveEnv()` in `src/utils/env.ts` strictly validates via `validateAgentName()` at runtime. Result: mixed-case names like \`CortextDesigner\` got written to disk, the daemon spawned them, then every `cortextos bus *` command failed. Affected agents were half-functional — daemon-managed fine, Telegram receiving worked, but sending, inbox reading, task creation, approvals, and every other bus call were broken.

## Discovery

Hit live this session: an ephemeral worker agent 'CortextDesigner' was created (mixed-case) and spawned successfully, but couldn't reply to user messages on Telegram because the bus command rejected its own name. Workaround was calling the Telegram API directly via curl using BOT_TOKEN from .env — not sustainable.

## Fix

Call \`validateAgentName()\` at the start of the add-agent action, BEFORE any filesystem operations. Same function that resolveEnv uses, so add-agent and resolveEnv now agree.

## Changes

- \`src/cli/add-agent.ts\` — import + call \`validateAgentName(name)\` at action entry, with clear error + valid name examples
- \`tests/unit/utils/validate.test.ts\` — new regression test for mixed-case / PascalCase names
- \`tests/unit/cli/add-agent-validation.test.ts\` (NEW) — 4 integration tests using Commander parseAsync + process.exit spy

## Test plan

- [x] \`npm run build\` clean
- [x] \`npm test\` 419/419 passing (414 baseline + 5 new)
- [x] **Live verification via manual dist copy** (no pm2 restart needed — CLI-only fix):
  - \`cortextos add-agent TestMixedCase --template agent --org testorg\` → rejected with clear error, exit 1, no directory created ✓
  - \`cortextos add-agent test-valid-name --template agent --org testorg\` → succeeds, normal flow works ✓
  - \`CTX_AGENT_NAME=test-valid-name cortextos bus list-agents\` → succeeds, proves fix makes add-agent and resolveEnv consistent ✓
  - \`CTX_AGENT_NAME=CortextDesigner cortextos bus list-agents\` → still fails (expected — fix prevents NEW bad agents, existing stay broken until recreated) ✓

## Why this was a good fast-loop PR

- Single source file change + 2 test files
- CLI-only fix → deploys via manual \`cp -r dist\` with no pm2 restart
- No install pipeline changes
- Deterministic repro (anyone can type a mixed-case name to trigger)
- Security-positive: closes validation gap
- Immediately user-visible: affected agents can now be recreated with valid names and actually work

## Out of scope (follow-ups)

- Inline regex duplicates in \`src/cli/setup.ts\`, \`src/cli/goals.ts\`, \`src/bus/agents.ts\`, \`src/bus/system.ts\` — hygiene cleanup, separate PR
- Existing CortextDesigner agent — stays broken, will be manually recreated as valid name
- Rename-agent CLI command — doesn't exist, would be nice, separate feature

### DIFF
--- a/src/cli/add-agent.ts
+++ b/src/cli/add-agent.ts
@@ -3,6 +3,7 @@ import { existsSync, readdirSync, readFileSync, writeFileSync, mkdirSync, copyFi
 import { join, resolve } from 'path';
 import { homedir } from 'os';
 import { OrgContext } from '../types';
+import { validateAgentName } from '../utils/validate';
 
 export const addAgentCommand = new Command('add-agent')
   .argument('<name>', 'Agent name')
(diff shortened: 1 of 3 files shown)


grandamenium:
## Cycles 1-3 verified ✅

**Cycle 1** (reproduce on plain main via existing CortextDesigner): mixed-case name was already present and broken — 'cortextos bus send-telegram' failed with 'Invalid agent name CortextDesigner' ✓

**Cycle 2** (fix on branch via manual dist copy, no pm2 restart):
- add TestMixedCase → rejected at add-agent time, exit 1, clear error, no dir created ✓
- add test-valid-name → normal flow works, dir + enabled-agents.json entry ✓
- CTX_AGENT_NAME=test-valid-name cortextos bus list-agents → succeeds, proves resolveEnv + add-agent are now consistent ✓
- CTX_AGENT_NAME=CortextDesigner cortextos bus list-agents → still fails (expected — fix prevents NEW bad agents, existing stays broken until recreated)

**Cycle 3** (canonical merged main at 2c36ab4 via manual dist copy):
- add CortextDesigner → rejected with clean error ('Invalid agent name CortextDesigner. Must contain only lowercase letters, numbers, underscores, and hyphens.') ✓

Special methodology note: because I'm an ephemeral worker agent running as a cortextOS daemon-managed process, running the standard helper-fastloop.sh would have pm2 restarted the daemon and killed me mid-cycle. Instead I manually copied dist/ without restarting — works because BUG-041 is a CLI-only fix (src/cli/add-agent.ts → dist/cli.js), the running daemon's already-loaded code is untouched.

BUG-041 closed. 419/419 tests passing. Out-of-scope cleanup (inline regex duplicates) filed as hygiene follow-up.

🤖 Posted via Claude Code (CortextDesigner ephemeral worker agent)

PR #56, opened 2026-04-14
fix(pty): redact JWT-shaped tokens from OutputBuffer

ClintMoody:
# fix(pty): redact JWT-shaped tokens from OutputBuffer

**Branch**: `pr/pty-redact-jwt-tokens`
**Commit**: `a4adae0`
**Priority**: SECURITY

---

## Problem

Anything an agent's shell writes to stdout gets captured by the cortextOS PTY and streamed both to the in-memory ring buffer (for bootstrap detection) and to a persisted log at `~/.cortextos/<instance>/logs/<agent>/stdout.log`. When an agent runs `curl -v` against an authenticated endpoint, curl's verbose output prints the response `Set-Cookie` header + its own `* Added cookie` trace line to stdout. If the response contains a session cookie (JWT, OAuth bearer, etc.), that token ends up persisted on disk with 600 perms — but still present indefinitely.

A baseline gitleaks audit of agent stdout logs found 16 JWTs this way (Telegram NextAuth `authjs.session-token` cookies, ~500 bytes each, eyJhb... shape). Zero were from cortextOS code paths logging tokens — 100% were from agent-run curl commands the PTY captured faithfully.

## Root cause

`src/pty/output-buffer.ts` `OutputBuffer.push(data)` is the single chokepoint where PTY output reaches both the ring buffer and the disk log. No redaction layer existed — the `push()` contract was byte-faithful capture, which is correct for bootstrap-detection needs but wrong for any token-bearing output that happens to flow through.

## Fix

- New `src/pty/redact.ts` exporting a single `redactSecrets(data)` helper.
- JWT pattern: requires the `eyJ` base64 header prefix + three dot-separated base64url segments of at least 10 characters each. Length qualifier prevents false positives on version strings (`v1.2.3`) or short alphanumeric fragments (`eyJab.x.y`). Matches replaced with the literal string `[REDACTED_JWT]`.
- `OutputBuffer.push()` applies `redactSecrets` once at the top; the scrubbed string is used for BOTH the ring buffer AND the disk log. Redacting both views keeps them consistent — no future footgun from the two paths diverging.
- Defense-in-depth design: this catches ANY tool that happens to print a JWT on stdout (curl, wget, openssl, dumps of a cookie jar, etc.), not just curl. Protects against secrets from tools the agent adopts later without the contributor needing to remember to scrub.

## Tests

6 new unit tests in `tests/unit/pty/output-buffer.test.ts`:

1. Single JWT in a single chunk → redacted in both the disk log AND the in-memory ring buffer
2. Multiple JWTs in one chunk → all replaced
3. Non-JWT PTY data (ANSI escapes + regular stdout + version strings) passes through unchanged — regression guard
4. `isBootstrapped()` still works after redaction — guards the ring buffer search path
5. **Chunk-boundary limitation**: JWT split across two `push()` calls is NOT redacted. Locked in as a regression guard so any future refactor that adds buffer-aware redaction has to be an explicit decision (test fails if the behavior silently changes).
6. Short alphanumeric resembling a truncated JWT (`eyJab.x.y`) is NOT redacted — length-qualifier false-positive guard.

- **Chunk-boundary edge case**: PTY data arrives in OS-buffered chunks (typically 4KB). If a chunk boundary falls inside a JWT, neither chunk matches the regex and the token slips through across two `push()` calls. JWTs are typically 300-500 bytes so they fit in one chunk in the overwhelming majority of real cases. Buffer-aware redaction (carry a partial-match buffer across chunks) is the follow-up if this edge case ever surfaces in production. The test explicitly asserts this behavior so it cannot regress without notice.
- **JWT-only pattern today**: only eyJ-prefixed JWTs are redacted. Other token shapes (bare OAuth bearer tokens, API keys with provider-specific prefixes, PATs, etc.) are not covered by this helper. Additive — adding more patterns to `redactSecrets` is a trivial extension without structural change, and I'd expect those to accumulate as observed leaks surface.
- **Historical log scrub NOT included**: operators with existing leaked JWTs in their stdout logs should rotate the affected credentials and optionally `rm` the log. This patch prevents future leaks only.
- **In-memory buffer redaction is belt-and-suspenders**: bootstrap detection searches for the literal string `permissions`, which is never going to collide with JWT redaction. I chose to redact both views for consistency rather than only the disk path; if a reviewer prefers disk-only redaction, that's a 2-line change.

(None found in this patch — the code is generic; no framework-internal terms or paths are hardcoded.)

---

**Test suite**: full suite green pre-prep (524 tests). Cherry-pick is clean against upstream/main a608a5d — no conflicts.

### DIFF
--- a/src/pty/output-buffer.ts
+++ b/src/pty/output-buffer.ts
@@ -1,4 +1,5 @@
 import { appendFileSync } from 'fs';
+import { redactSecrets } from './redact.js';
 
 // Dynamic import for strip-ansi (ESM module)
 let stripAnsi: (text: string) => string;
@@ -27,9 +28,18 @@ export class OutputBuffer {
   /**
    * Push new output data into the buffer.
    * Also streams to log file if configured.
+   *
+   * Secret redaction runs once at the top via `redactSecrets` and the
+   * scrubbed string is used for BOTH the in-memory ring buffer AND the
+   * disk log. Without this, any JWT or session cookie an agent's shell
+   * happens to print (e.g. curl -v against an authenticated endpoint)
+   * would end up persisted to stdout.log verbatim. See src/pty/redact.ts
+   * for the rationale + the known chunk-boundary limitation.
    */
   push(data: string): void {
-    this.chunks.push(data);
+    const safe = redactSecrets(data);
+
+    this.chunks.push(safe);
     if (this.chunks.length > this.maxChunks) {
       this.chunks.shift();
     }
@@ -37,7 +47,7 @@ export class OutputBuffer {
     // Stream to log file (replaces tmux pipe-pane)
     if (this.logPath) {
       try {
-        appendFileSync(this.logPath, data, 'utf-8');
+        appendFileSync(this.logPath, safe, 'utf-8');
       } catch {
         // Ignore log write errors
       }
--- a/src/pty/redact.ts
+++ b/src/pty/redact.ts
@@ -1,0 +1,52 @@
+/**
+ * PTY output redaction.
+ *
+ * Secret-bearing output can reach the PTY capture stream whenever an agent
+ * runs a shell command that prints credentials — curl -v against an
+ * authenticated endpoint, wget --debug, openssl s_client, dumping a cookie
+ * jar, etc. The PTY's OutputBuffer ring captures everything the child
+ * process emits and also streams it verbatim to a persisted stdout.log.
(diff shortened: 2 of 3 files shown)


grandamenium:
**Pipeline Stage 1–2 Passed** (static analysis). Score: 9/10. Recommendation: MERGE. Full report in `pr-reviews/56/result.md`. Sandbox execution deferred — overnight batch volume. Ready for James's approval.

Issue #27, opened 2026-04-11
BUG-047: Daemon has no persistent log file — crash root causes are invisible


grandamenium:
## Summary

The cortextOS daemon writes all operational logs (crash recovery decisions, exit codes, session timer events, restart triggers) to its own `stdout` only. No log file is created. When a crash loop occurs, the only evidence available is in agent-level `stdout.log` files which contain raw PTY output with terminal escape sequences — hard to parse, contains no daemon-level context.

## Impact

Discovered during RCA for Apr 10 crash loop (20:14–22:29 UTC, hit all 6 agents simultaneously). Without daemon logs, root cause could not be fully confirmed. Key questions that required guessing:
- What exit code did each agent return?
- Which code path triggered restart (session timer vs crash vs stop)?
- Was the session timer involved?

## Evidence

The daemon process (`node dist/daemon.js`) runs with no log file argument. Agent `stdout.log` files contain ANSI escape sequences and rendered TUI output, not structured daemon decisions. The `crashes.log` per-agent only records clean stop/disable events, not unclean exits during crash loops.

## Proposed Fix

Add a daemon-level log file at `~/.cortextos/<instance>/daemon.log` (or similar). Log:
- Agent start/stop/crash events with exit codes
- Session timer fires 
- Crash recovery decisions (backoff, halt)
- sessionRefresh triggers

Low noise — these are already being called via `this.log()` in `agent-process.ts`. Just route them to a file.

## Related

Companion issue: BUG-048 (session timer not re-reading config after changes)

grandamenium:
**Triage finding (cortext-designer, 2026-05-07): still open AND the codebase now has misleading state.**

`src/cli/start.ts:117` computes a `daemon.log` path:

```ts
const logDir = join(ctxRoot, 'logs');
const logFile = join(logDir, 'daemon.log');
```

…but the very next spawn block uses `stdio: ['ignore', 'ignore', 'ignore']` (line 121), so daemon stdout/stderr is black-holed. The `logFile` variable is referenced only in a console.log error message ("Check logs if agents do not appear: ${logFile}") — so the user is told to check a file that is never written.

[...]

PR #57, opened 2026-04-14
fix(daemon): classify PM2 shutdown PTY exits as planned stops

ClintMoody:
# fix(daemon): classify PM2 shutdown PTY exits as planned stops, persist crash audit to restarts.log

**Branch**: `pr/daemon-classify-pm2-shutdown-exits`
**Commit**: `0fc31c5`
**Closes**: #27

---

## Problem

Every `pm2 restart cortextos-daemon` (or `pm2 stop`) was silently inflating each running agent's `.crash_count_today` by one, even though the shutdown was planned and the agents had exited cleanly. Scheduled restarts with `crash #N` attached were logged to the daemon's stdout but never actually ran (the daemon itself was going down), leaving an audit trail that was both wrong AND incomplete.

Additionally: when a REAL crash DID happen, there was no persistent record of it anywhere the operator would look. `restarts.log` only received planned restarts from `src/bus/system.ts` (SELF-RESTART, HARD-RESTART). Daemon-classified crashes (handleExit's crash-recovery path + HALTED path) only emitted to the daemon's stdout, which rotates through PM2 and gets lost.

## Root cause

When PM2 stops the daemon, it sends SIGTERM to the entire process group. Every PTY's Claude Code child receives SIGTERM directly from the OS process group AT THE SAME TIME the daemon begins its graceful shutdown. The daemon's `stopAll()` iterates agents sequentially; any agent whose PTY has already exited (cleanly, code 0) by the time `stopAgent()` reaches it arrives at `handleExit()` with `stopRequested=false`. That short-circuit check (`if (this.stopRequested || this.stopping)`) was the only gate protecting against misclassification — and during a PM2 shutdown it's not set yet.

## Fix

Two changes in `src/daemon/agent-process.ts::handleExit()`:

1. **Honor the `.daemon-stop` marker** that `agent-manager.ts::stopAll()` already writes at the START of its shutdown sequence (originally added for the crash-alert hook to distinguish daemon shutdown from crashes). If the marker exists and is fresh (age < 60s), treat the exit as intentional and return early before touching the crash counter. Stale markers (>60s) are ignored so they cannot silently mask a genuine crash indefinitely after a prior shutdown failed to clean up.

2. **Append every daemon-classified unplanned exit to `restarts.log`** — both the CRASH-recovery path and the HALTED path. Format matches the existing SELF-RESTART/HARD-RESTART lines so operators have one file to read. Write is wrapped in try/catch and swallows errors to never break crash recovery on a logging failure.

## Tests

3 new unit tests in `tests/unit/daemon/agent-process.test.ts`:

1. Unexpected PTY exit → `restarts.log` gets a `CRASH: exit_code=... crash_count=... backoff_s=...` line (the fix's happy path)
2. Fresh `.daemon-stop` marker present → exit is NOT classified as a crash, `restarts.log` unchanged, no restart scheduled (PM2 shutdown path)
3. Stale `.daemon-stop` marker (mtime > 60s ago) → exit IS still classified as a crash (regression guard against leftover markers silently masking real crashes)

Full suite: 454 tests pass against upstream/main baseline + this patch.

- **60s marker staleness window** is a judgment call. Shorter would reduce the mask-a-real-crash window further but risks false positives during a slow shutdown where `stopAll()` takes a while to iterate. Longer would tolerate slower shutdowns at the cost of a bigger real-crash mask window. 60s felt right for a daemon whose normal shutdown completes in <5s; adjust if your deployment has pathologically slow shutdowns.
- **The `restarts.log` write is append-only + best-effort.** Concurrent crashes of multiple agents won't corrupt the file (appendFileSync is atomic for small writes via O_APPEND + write(2)), but a crash-recovery path that itself crashes mid-write could leave a truncated line. The try/catch swallows the error — the operator loses one audit line but the daemon keeps running.
- **The HALTED path logs `max_crashes=N`** instead of `backoff_s` because at HALTED there's no retry scheduled. Separate field matches the code path shape; parsers reading restarts.log can key on the prefix (`CRASH:` vs `HALTED:` vs `SELF-RESTART:` vs `HARD-RESTART:`).

- `src/daemon/agent-process.ts:365` has a pre-existing JSDoc comment using `boss` as an example agent name in a path-translation example: `// e.g. /Users/foo/agents/boss -> -Users-foo-agents-boss (leading sep becomes -)`. Not touched by this patch. `boss` is a conventional orchestrator name in cortextOS installs but worth flagging if upstream prefers a more neutral example like `orchestrator` or `agent`.

---

**Cherry-pick note**: one conflict in `tests/unit/daemon/agent-process.test.ts` `beforeEach` block where upstream's `mockInjectMessage.mockClear()` and this patch's `fsMocks.*.mockReset()` calls both need to live — resolved by keeping both. Also adjusted the `vi.mock('fs', ...)` factory to return direct `fsMocks.X` references instead of arrow-wrapped calls, so both `fsMocks.X.mockReset()` (used by this patch's tests) and `vi.mocked(fs.X).mockImplementation()` (used by upstream's verifyCronsAfterIdle + BUG-048 tests) work against the same underlying vi.fn(). Full suite green post-resolution.

### DIFF
--- a/src/daemon/agent-process.ts
+++ b/src/daemon/agent-process.ts
@@ -1,4 +1,4 @@
-import { existsSync, readFileSync, writeFileSync } from 'fs';
+import { appendFileSync, existsSync, readFileSync, statSync, writeFileSync } from 'fs';
 import { join, sep } from 'path';
 import { homedir } from 'os';
 import type { AgentConfig, AgentStatus, CtxEnv } from '../types/index.js';
@@ -309,6 +309,29 @@ export class AgentProcess {
     this.pty = null;
     this.clearSessionTimer();
 
+    // When the cortextos daemon is shut down by PM2, SIGTERM propagates to
+    // the whole process group and reaches each PTY's Claude Code child
+    // BEFORE the daemon's stopAll() loop has a chance to call stopAgent() on
+    // it. Those children exit cleanly (code 0) but arrive at handleExit with
+    // stopRequested=false, which used to classify the exit as a crash and
+    // inflate .crash_count_today by one per agent, per PM2 restart.
+    //
+    // agent-manager.ts:stopAll() already writes a `.daemon-stop` marker in
+    // every agent's state dir at the START of its shutdown loop for an
+    // unrelated reason (SessionEnd crash-alert hook). We reuse that marker
+    // here as the authoritative "the daemon is going down" signal. If the
+    // marker exists AND is recent (written within the last 60s), any PTY
+    // exit is a shutdown casualty, not a real crash — swallow it.
+    //
+    // The 60s window guards against a stale marker from a previous shutdown
+    // that wasn't cleaned up: we do NOT want an old marker to silently mask
+    // a genuine crash days later. handleExit does NOT delete the marker —
+    // cleanup stays with agent-manager / hook-crash-alert per the existing
+    // separation of concerns.
+    if (this.isDaemonShuttingDown()) {
+      return;
+    }
+
     // BUG-040 fix: check stopRequested instead of (only) stopping. The
     // stopping flag is cleared inside stop() after a 15s timeout window —
     // which means a slow PTY shutdown can fire handleExit AFTER stopping is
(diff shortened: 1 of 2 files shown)


grandamenium:
**Pipeline Stage 1–2 Passed** (static analysis). Score: 9/10. Recommendation: MERGE. Full report in `pr-reviews/57/result.md`. Sandbox execution deferred — overnight batch volume. Ready for James's approval.

PR #58, opened 2026-04-14
fix(telegram): validate BOT_TOKEN and CHAT_ID at add-agent and setup time

ClintMoody:
# fix(telegram): validate BOT_TOKEN and CHAT_ID against Telegram API before enable + setup writes .env

**Branch**: `pr/telegram-credential-validation`
**Commit**: `71ebc7c`

---

## Problem

An agent can boot on silently broken Telegram credentials and look healthy in the dashboard while every inbound/outbound message fails in the background. Four distinct failure shapes produce this state:

1. **bad_token** — BOT_TOKEN is malformed or revoked (401 Unauthorized on any API call)
2. **chat_not_found** — CHAT_ID is a valid shape but the user never sent `/start` to the bot, so the chat doesn't exist from the bot's side (400 Bad Request: chat not found)
3. **bot_recipient** — CHAT_ID resolves to another bot, not a user (403 Forbidden: bots can't send messages to bots)
4. **self_chat** — CHAT_ID equals the bot's own user id (pasted from the BOT_TOKEN prefix by mistake during setup). Same 403 as bot_recipient but with a different root cause the operator needs to recognize.

Before this patch, all four silently passed the existing BOT_TOKEN/CHAT_ID *presence* check in `cortextos enable`. First contact with Telegram happened only at first real `sendMessage` — where the failure buried itself in the agent's stdout log while the dashboard happily marked the agent as alive. Field example: an agent provisioned with CHAT_ID equal to the first segment of BOT_TOKEN (the bot's own user id) — `sendMessage` returned 403 "bots can't send messages to bots", but nothing surfaced the misconfiguration.

## Root cause

`cortextos enable <agent>` validated *presence* of BOT_TOKEN + CHAT_ID in `.env` but made no API call to verify they were *usable*. `cortextos setup` had no validation at all. No config-time probe existed that distinguished the four error shapes above, so there was no place to surface them with clear guidance.

## Fix

Adds `TelegramAPI.validateCredentials(chatId)` — a config-time probe that:

- Calls `getMe` to verify BOT_TOKEN and retrieve the bot's user id
- Short-circuits with `self_chat` if CHAT_ID equals the bot's own id (no wasted API call)
- Calls `getChat(chatId)` to verify the chat exists and the recipient type
- Returns a tagged discriminated union `{ ok: true, botUsername, botId, chatType, chatTitle }` or `{ ok: false, reason: '...', detail: '...' }` where `reason` is one of the six failure classes (bad_token, chat_not_found, bot_recipient, self_chat, network_error, rate_limited)
- Bounded at 10s per call via an internal `withTimeout` helper so a network partition cannot hang enable/setup indefinitely
- Never leaks any portion of BOT_TOKEN into any error message (explicitly asserted in test)

Adds `formatValidateError(result)` — the single source of truth for user-facing error strings, so `enable-agent.ts` and `setup.ts` never diverge on the wording.

**Wire-in at `cortextos enable <agent>`**: preflight probe right after the existing BOT_TOKEN/CHAT_ID presence check. `network_error` and `rate_limited` are warn-not-fail so offline enable and burst enables (e.g. during morning cron cascade) still succeed. All four config-level failures fail fast with the formatted message.

**Wire-in at `cortextos setup`** (both the orchestrator-creation path and the additional-agent loop): probe BEFORE `writeAgentEnv()`, so a broken `.env` is never committed to disk. Gives the user up to 3 recoverable re-entries; bails entirely on `bad_token` (requires fixing via @BotFather, not something the wizard can recover from).

## Tests

16 new unit tests in `tests/unit/telegram/api.test.ts`:

- Happy path (user chat)
- bad_token (401 on getMe) — NO getChat attempt
- self_chat (CHAT_ID equals getMe.id) — no getChat, no token leak in error string
- chat_not_found (400 on getChat)
- bot_recipient via `getChat` returning `is_bot=true`
- bot_recipient via 403 on getChat
- network_error (fetch throws)
- rate_limited (429)
- 10s timeout path (fake timers + never-resolving fetch)
- Empty chat_id — short-circuits with `chat_not_found`, zero API calls
- Six `formatValidateError` message-content tests including:
  - bad_token does not leak token text
  - self_chat names the trap and points at `/start` + `getUpdates`
  - chat_not_found suggests `/start`
  - bot_recipient explains user-vs-bot distinction
  - network_error includes underlying detail
  - rate_limited mentions retry

- **Only 4 canonical Telegram error classes** are mapped to their own `reason`. Any other 4xx/5xx from Telegram falls through to `bad_token` (if at getMe step) or `chat_not_found` (if at getChat step) — a conservative bucketing that keeps the common-case messages sharp at the cost of less-useful text for rare error shapes. If that becomes a real problem in practice, adding new reason cases is purely additive with no structural change.
- **10s timeout** is hardcoded. Reasonable default for a config-time check (not a hot path). Not currently env-overridable; add if a deployment needs a different value.
- **Both CLI paths log a visible confirmation on success** (`Telegram validated: bot=@<username> chat=<id> type=<type>`). Intentional — operators get a positive signal that validation passed. If you prefer silent-success, that's a 1-line change.

- `src/cli/setup.ts:267` has a pre-existing default value `'boss'` for the orchestrator agent name (`askDefault(iface, '  Orchestrator agent name', 'boss')`). Not touched by this patch. `boss` is a conventional orchestrator name but worth flagging if upstream prefers a more neutral default like `orchestrator` or `coordinator`.

---

**Test suite**: full 467/467 pass. `tests/unit/telegram/api.test.ts` is a NEW file with 16 tests (verified by standalone `npx vitest run`). `npm run build` clean. Cherry-pick onto upstream/main a608a5d was clean (zero conflicts).

### DIFF
--- a/src/cli/enable-agent.ts
+++ b/src/cli/enable-agent.ts
@@ -3,6 +3,7 @@ import { existsSync, readFileSync, writeFileSync, mkdirSync } from 'fs';
 import { join } from 'path';
 import { homedir } from 'os';
 import { IPCClient } from '../daemon/ipc-server.js';
+import { TelegramAPI, formatValidateError } from '../telegram/api.js';
 
 /**
  * BUG-035 fix: discover the cortextOS framework root without depending on
@@ -179,6 +180,44 @@ export const enableAgentCommand = new Command('enable')
       console.error(`Error: .env for agent "${agent}" is missing required values: ${missing.join(', ')}`);
       console.error(`Edit ${agentEnvPath} and set BOT_TOKEN and CHAT_ID before enabling.`);
       process.exit(1);
+    }
+
+    // self-chat trap preflight: validate BOT_TOKEN + CHAT_ID against the live
+    // Telegram API before registering. Catches bad tokens, unreachable chats,
+    // bot-recipient configs, and the self_chat trap (CHAT_ID == bot's own
+    // user id) BEFORE the agent boots up on a silently broken config. Without
+    // this, the first real sendMessage call fails with a cryptic 401/400/403
+    // buried in the agent's stdout log, and the dashboard happily shows the
+    // agent as alive.
+    //
+    // Hard-fails on config-level reasons (bad_token, chat_not_found,
+    // bot_recipient, self_chat). Warns but does not block on transient
+    // reasons (network_error, rate_limited) so offline enable and burst
+    // enables during the morning cascade still succeed.
+    try {
+      const telegramApi = new TelegramAPI(env.BOT_TOKEN);
+      const validation = await telegramApi.validateCredentials(env.CHAT_ID);
+      if (validation.ok) {
+        const label = validation.chatTitle ? ` (${validation.chatTitle})` : '';
+        console.log(
+          `Telegram validated: bot=@${validation.botUsername} chat=${env.CHAT_ID} type=${validation.chatType}${label}`,
+        );
+      } else if (validation.reason === 'network_error' || validation.reason === 'rate_limited') {
+        console.error(`Warning: could not verify Telegram credentials (${validation.reason}).`);
+        console.error(`  ${formatValidateError(validation)}`);
+        console.error('  Continuing anyway — re-run enable after connectivity is restored to confirm.');
+      } else {
+        console.error(`Error: Telegram credentials for agent "${agent}" failed validation.`);
+        console.error(`  ${formatValidateError(validation)}`);
+        console.error(`  Edit ${agentEnvPath} and re-run: cortextos enable ${agent}`);
+        process.exit(1);
+      }
+    } catch (err) {
+      // Defensive: validateCredentials should never throw, but if it does,
+      // fall through with a warning rather than blocking enable on a bug in
+      // the validator itself.
+      console.error(`Warning: Telegram credential validation crashed: ${err instanceof Error ? err.message : String(err)}`);
+      console.error('  Continuing enable. Investigate the validator if this recurs.');
     }
 
     const agents = readEnabledAgents(options.instance);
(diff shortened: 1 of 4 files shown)


grandamenium:
**Pipeline Stage 1–2 Passed** (static analysis). Score: 9/10. Recommendation: MERGE. Full report in `pr-reviews/58/result.md`. Sandbox execution deferred — overnight batch volume. Ready for James's approval.

grandamenium:
This PR has a merge conflict with main. PR #59 (parse-entity retry) was recently merged and both touch telegram/api.ts. Please rebase on current main to resolve the conflict — the validateCredentials() addition should apply cleanly on top of the retry changes. Pipeline complete (9/10 MERGE) and this feature is high value.

grandamenium:
This PR now has a merge conflict with main —  was added by PR #86 (fetch timeout tests) and conflicts with the new test file here. Please rebase on main and merge both test suites into a single file (keep the `TelegramAPI fetch timeout` describe block from main plus your `TelegramAPI.validateCredentials` and `formatValidateError` blocks). Happy to assist if needed. This is a high-value PR (9/10, sandbox-verified) and we want to get it merged.

grandamenium:
Superseded by #235 — same fix (commit authored by @ClintMoody preserved via cherry-pick) rebased onto current main + merged 2026-04-23. This PR's branch was 9 days stale and couldn't rebase cleanly against main after today's stability wave landed (#223/#217/#210/#181/#184/#226). Thanks @ClintMoody — live-validated on a production agent before merge: all 5 scenarios (valid/bad_token/self_chat/chat_not_found/empty) behaved correctly. Full test evidence in #235.

PR #59, opened 2026-04-14
fix(telegram): retry sendMessage with plain text on parse-entity errors + add --plain-text opt-in

ClintMoody:
# fix(telegram): retry sendMessage with parse_mode=null on parse-entity errors + --plain-text opt-in

**Branch**: `pr/telegram-plain-text-fallback`
**Commit**: `2a862e8`

---

## Problem

Agents drop Telegram messages silently when their natural prose contains an unescaped `_`, `*`, backtick, or `[`. `TelegramAPI.sendMessage()` always sent with `parse_mode: "Markdown"`; any parse-entity error from the post() helper bubbled up to the caller (typically a CLI command that logged "Failed to send" and exited 1). 11 such failures observed across agent stdout logs over a 24h window — every one was natural prose, not a code path that could scrub its output in advance.

## Root cause

`sendMessage` had a single code path: Markdown-only, with no fallback and no opt-in plain-text mode. Any call site that wanted plain text had to work around the API instead of asking for it. The multi-chunk path compounded the issue — each chunk went through the same unprotected post() call, so a single bad character aborted the entire send mid-stream.

## Fix

- **One-shot retry on parse-entity errors**: detect Telegram's parse-entity error shape, log a one-line stderr warning, retry the same chunk ONCE with `parse_mode` omitted. If the retry also fails, rethrow so real failures still surface. 401/400/403/network keep failing fast — only parse-entity triggers the retry.
- **Opt-in plain-text mode**: new 4th optional argument `{ parseMode: null }` skips the Markdown attempt entirely, saving the retry roundtrip when callers KNOW the message will trip the parser. Exposed via a new `--plain-text` flag on `cortextos bus send-telegram`.
- **Outbound-log metadata**: new optional `OutboundLogMetadata` interface on `logOutboundMessage`. When the CLI detects a parse-mode fallback, the `outbound-messages.jsonl` entry carries `parse_mode`, `parse_fallback`, and `parse_fallback_reason` for ad-hoc auditing. Backwards compatible — callers passing no metadata produce the same JSONL shape.
- **Shared `sendChunk()` helper** for the multi-chunk path so retry-per-chunk works correctly. Side effect: the last chunk now gets the `replyMarkup` keyboard (the old behavior silently dropped keyboards on every chunk of a chunked send — arguably a latent bug).

## Tests

9 new unit tests in `tests/unit/telegram/send-message.test.ts` covering every branch:

1. Happy path (Markdown succeeds first try)
2. Parse-entity error → retry with plain text succeeds
3. Retry also fails → rethrow
4. Non-parse-entity error fails fast (no retry)
5. Opt-in `parseMode: null` skips Markdown
6. Opt-in edge case (empty message)
7. Chunked plain-text (all chunks sent without Markdown)
8. Chunked retry — one chunk triggers retry, others pass through
9. `onParseFallback` hook fires with the right metadata

- **Parse-entity detection is error-message-shape dependent**. Telegram's error text for parse failures contains "can't parse entities" or "Bad Request: can't parse". Shape change on Telegram's side could regress the retry trigger — kept narrow on purpose so false-positive retries (for unrelated 400s) don't mask real config bugs.
- **One retry only**. If the plain-text retry itself hits a transient network error, we rethrow rather than retry again. Keeps the code path bounded — callers needing resilience should retry at a higher layer.
- **Keyboard-on-last-chunk behavior is a behavior change** (old code silently dropped keyboards across all chunks of a chunked send). Test explicitly asserts the new behavior.

---

**Test suite**: full 460/460 pass. Cherry-pick onto upstream/main a608a5d was clean (zero conflicts).

### DIFF
--- a/src/cli/bus.ts
+++ b/src/cli/bus.ts
@@ -718,10 +718,11 @@ busCommand
   .command('send-telegram')
   .description('Send a message to a Telegram chat')
   .argument('<chat-id>', 'Telegram chat ID')
-  .argument('<message>', 'Message text (supports Telegram Markdown)')
+  .argument('<message>', 'Message text (supports Telegram Markdown unless --plain-text is set)')
   .option('--image <path>', 'Send a photo with caption')
   .option('--file <path>', 'Send a document/file with caption (any file type)')
-  .action(async (chatId: string, message: string, opts: { image?: string; file?: string }) => {
+  .option('--plain-text', 'Skip Telegram Markdown parsing entirely. Use this when the message contains unescaped _, *, backtick, or [ that would otherwise trip the Markdown parser. Without this flag, sendMessage still retries once with parse_mode disabled on a parse-entity error — so it is purely an opt-in to save the retry roundtrip.', false)
+  .action(async (chatId: string, message: string, opts: { image?: string; file?: string; plainText?: boolean }) => {
     // Resolve bot token: agent .env first, then process.env
     const env = resolveEnv();
     let botToken = '';
@@ -751,6 +752,7 @@ busCommand
     const api = new TelegramAPI(botToken);
     try {
       let sentMessageId = 0;
+      let parseFallbackReason: string | null = null;
       if (opts.image) {
         const result = await api.sendPhoto(chatId, opts.image, message);
         sentMessageId = result?.result?.message_id ?? 0;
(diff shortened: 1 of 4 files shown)


grandamenium:
**Pipeline Stage 1–2 Passed** (static analysis). Score: 9/10. Recommendation: MERGE. Full report in `pr-reviews/59/result.md`. Sandbox execution deferred — overnight batch volume. Ready for James's approval.

PR #60, opened 2026-04-14
fix(kb): warn-and-skip on missing knowledge-base config instead of unhandled crash

ClintMoody:
# fix(kb): warn-and-skip on missing knowledge-base config instead of unhandled crash

**Branch**: `pr/kb-ingest-graceful-missing-config`
**Commit**: `29a069a`

---

## Problem

Running `cortextos bus kb-ingest` against an org whose `knowledge-base/config.json` was never created used to produce an ungraceful crash: the underlying python MMRAG tool exited with "Config not found. Run setup first", the subprocess wrapper threw the non-zero-exit error, and Node dumped a full JS stack trace on top of the python stderr. Alarming for operators and indistinguishable from a real bug.

`queryKnowledgeBase` had the symmetric UX bug: the inner `runQuery()` try/catch already swallowed the Config-not-found throw and returned empty results, but the operator got zero hits with no indication of *why* — indistinguishable from a legitimate empty query.

## Root cause

Neither call site checked for the config file before invoking the python tool. `ingestKnowledgeBase` had no try/catch at all around the spawn, so the throw became an unhandled exception. `queryKnowledgeBase` had a catch that discarded the error silently.

## Fix

- New `kbConfigured(env)` helper — a one-line `existsSync(env.MMRAG_CONFIG)` check.
- `ingestKnowledgeBase`: upfront check emits a `[kb]` warn naming the org + a "run setup" hint, then returns cleanly. No throw, no stack dump, exit code 0.
- `queryKnowledgeBase`: upfront check emits the same-style warn and returns the same empty shape the rest of the code path expects.
- Both warn messages use the `[kb]` prefix for grep-ability and name the org so multi-org operators can identify which config is missing.

## Tests

5 new unit tests in `tests/unit/bus/knowledge-base.test.ts` (first test file for `src/bus/knowledge-base.ts`):

1. `ingestKnowledgeBase` with missing config → warns, returns, does NOT spawn python
2. `ingestKnowledgeBase` with config present → proceeds to spawn (happy path)
3. `queryKnowledgeBase` with missing config → warns, returns empty shape, no spawn
4. `queryKnowledgeBase` with config present → proceeds to spawn
5. Warn-message shape invariant — regression guard that the `[kb]` prefix + org-name format stays stable for log grepping

- **Silent-success after the warn is intentional**. The CLI command still exits 0 because "no KB configured" is an expected state for a fresh org, not an error. If a deployment treats missing config as a hard failure, wrap the call site and check the warn output.
- **Warn goes to stderr via `console.warn`** — matches the rest of the bus module's convention. Not configurable to a structured logger at this layer.

---

**Test suite**: full 456/456 pass. Cherry-pick onto upstream/main a608a5d was clean. Scrub: `/tmp/bob/...` fixture paths in the test file were renamed to `/tmp/agent/...` to keep the test corpus agent-name-neutral.

### DIFF
--- a/src/bus/knowledge-base.ts
+++ b/src/bus/knowledge-base.ts
@@ -49,6 +49,22 @@ function loadSecretsEnv(frameworkRoot: string, org: string): Record<string, stri
 }
 
 /**
+ * Check whether the knowledge base config file exists for a given env.
+ *
+ * The Python MMRAG tool loads its config from env.MMRAG_CONFIG
+ * (`knowledge-base/config.json` under the org's state dir) and exits with
+ * "Config not found. Run setup first" if the file is absent. When that
+ * happens, execFileSync throws a non-zero-exit error which — if not caught
+ * — produces a user-facing unhandled-throw stack dump on top of the
+ * already-printed Python error. This helper lets callers detect the
+ * missing-config state UP FRONT and respond gracefully (warn + return)
+ * instead of relying on brittle stderr string matching after the throw.
+ */
+function kbConfigured(env: Record<string, string>): boolean {
+  return existsSync(env.MMRAG_CONFIG);
+}
+
+/**
  * Build the full env object needed by mmrag.py calls.
  */
 function buildKBEnv(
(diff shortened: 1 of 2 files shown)


grandamenium:
**Pipeline Stage 1–2 Passed** (static analysis). Score: 9/10. Recommendation: MERGE. Full report in `pr-reviews/60/result.md`. Sandbox execution deferred — overnight batch volume. Ready for James's approval.

PR #61, opened 2026-04-14
fix(task): cross-org lookup for update-task and complete-task via findTaskFile helper

ClintMoody:
# fix(task): cross-org lookup for update-task and complete-task via findTaskFile helper

**Branch**: `pr/task-cross-org-lookup`
**Commit**: `330ff99`

---

## Problem

The standard cortextOS dispatch pattern is an orchestrator in one org filing a task that gets assigned to a specialist in a different org. Before this patch, `updateTask` and `completeTask` hardcoded the task file path as `join(paths.taskDir, taskId + '.json')` — always the CALLER's org tasks dir. When the assignee tried to drive the lifecycle from their own session (`cortextos bus update-task <id> in_progress`), the command failed with "Task <id> not found: ENOENT" because the task file lived in the FILER's org. Every cross-org assignment required a manual workaround where the filer ran update/complete on behalf of the assignee.

## Root cause

Task file resolution was hardcoded to the caller's own tasks dir with no fallback. No helper existed for cross-org lookup.

## Fix

Adds a new exported `findTaskFile(paths, taskId)` helper with a two-tier lookup:

1. **Fast path**: check the caller's own org tasks dir first. Zero scan cost when it hits (same-org is the common case).
2. **Fallback**: scan every sibling org under `<ctxRoot>/orgs/*/tasks/` for a matching task file. Only runs when the fast path missed, so same-org operations take no perf hit.

`updateTask` and `completeTask` refactored to use `findTaskFile`. Error messages improved:
- Find-miss errors now name the actual search path (`<ctxRoot>/orgs/`) so operators can see where the scan looked.
- Post-find I/O errors get their own distinct "update failed" / "complete failed" prefix to separate them from path-resolution failures.

**Ambiguity handling**: if the fallback scan finds the same task ID in multiple orgs (theoretically possible due to `task_<epoch_ms>_<3digit>` collisions, practically vanishingly rare), warn loudly naming the task ID, match count, AND all org names where the collisions occurred, then return the first match. Operations keep flowing while the anomaly is surfaced.

**`listTasks` DELIBERATELY UNCHANGED** — stays single-org by default because the dashboard sync loop depends on per-org task listings. A cross-org list-tasks flag can land as a separate opt-in enhancement.

## Tests

6 new cases in `tests/unit/bus/task.test.ts` against a real nested filesystem layout (matching the production shape at `~/.cortextos/<instance>/orgs/<org>/tasks/`), not a mocked shortcut:

1. Same-org fast path regression guard
2. Cross-org update via fallback
3. Cross-org complete via fallback
4. Not-found error shape names the search path
5. Ambiguity — warn names all colliding orgs
6. `listTasks` scoping regression guard — explicitly documents the single-org contract in its comment so a future refactor cannot accidentally make listTasks cross-org without failing this test

- **Fallback scan is `readdirSync` over `<ctxRoot>/orgs/`**. On a deployment with hundreds of orgs the fallback path gets slower linearly. Fast path still O(1) for same-org (the hot case); only the fallback pays for the scan.
- **Ambiguity returns the first match**. Not the "right" match — the first one `readdirSync` yields. The warn makes the anomaly visible so operators can investigate, but if multi-org ID collisions become real rather than theoretical, add a proper disambiguation policy.
- **No write-path cross-org support added**. `createTask` still writes to the caller's own org. Filing a task into a different org's tasks dir requires a separate API — out of scope for this patch.

---

**Test suite**: full 457/457 pass. Cherry-pick onto upstream/main a608a5d was clean. Scrubs applied: JSDoc example `(e.g. boss@ElementOneSound ... bob@agentnet)` removed from `findTaskFile`; test-fixture comments "The exact bug boss hit" / "Boss's explicit scoping note" rewritten in neutral provenance tone; timestamp-shaped test IDs `task_1775939913834_528/529` renamed to `task_test_001/002`.

### DIFF
--- a/src/bus/task.ts
+++ b/src/bus/task.ts
@@ -1,4 +1,4 @@
-import { readdirSync, readFileSync, renameSync } from 'fs';
+import { existsSync, readdirSync, readFileSync, renameSync } from 'fs';
 import { join } from 'path';
 import type { Task, Priority, TaskStatus, BusPaths, StaleTaskReport, ArchiveReport } from '../types/index.js';
 import { atomicWriteSync, ensureDir } from '../utils/atomic.js';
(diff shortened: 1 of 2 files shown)


grandamenium:
**Pipeline Stage 1–2 Passed** (static analysis). Score: 9/10. Recommendation: MERGE. Full report in `pr-reviews/61/result.md`. Sandbox execution deferred — overnight batch volume. Ready for James's approval.

PR #62, opened 2026-04-14
fix(org): normalize org casing at KB-write and dashboard sync

ClintMoody:
# fix(org): normalize org casing at KB-write and dashboard sync to stop ghost-dir sprawl

**Branch**: `pr/org-casing-normalization`
**Commit**: `17a5fe0`

---

## Problem

Two linked bugs caused by lack of org-name normalization:

1. **Ghost state dirs**: Agents could bootstrap a ghost state directory at `~/.cortextos/<instance>/orgs/<lowercase_org>/` just by passing a lowercase `--org` argument to any `cortextos bus kb-*` command. `bus/knowledge-base.ts::buildKBEnv()` joined the org name verbatim into the path, and the Python MMRAG tool then auto-initialized a `knowledge-base/config.json` (including a copy of the org's `GEMINI_API_KEY`) at the lowercase path. That ghost dir persisted indefinitely as an orphan one-file tree alongside the canonical CamelCase state dir — with a key copy inside.
2. **Dashboard log flood**: `dashboard/src/lib/config.ts::getOrgs()` scanned both the state dir and framework root and unioned results into a case-sensitive Set, so each scan returned BOTH ghost and canonical. Every polling cycle of `syncAll()` produced `"syncTasks org=<lowercase> exists=false"` misses against the ghost dir, flooding dashboard.log.

## Root cause

No normalization layer existed between caller-supplied org names and filesystem writes. Case-drifted inputs reached path construction verbatim.

## Fix

**`src/utils/org.ts::normalizeOrgName(frameworkRoot, org)`** — returns the canonical filesystem casing for a given org. Resolution order: exact match → case-insensitive scan → unchanged input. Never returns a name that does not exist on disk (except when passing through unchanged input for empty/missing cases).

**Wire-in in `src/bus/knowledge-base.ts`** at three call sites (`queryKnowledgeBase`, `ingestKnowledgeBase`, `ensureKBDirs`) so every MMRAG invocation — path construction AND ChromaDB collection naming — uses canonical casing regardless of caller input. Without this, `shared-acmecorp` and `shared-AcmeCorp` become two distinct ChromaDB collections and a case-drifted query silently hits the wrong one. `ensureKBDirs` grows a `frameworkRoot` parameter; sole caller in `src/cli/bus.ts` updated accordingly.

**Rewrite `dashboard/src/lib/config.ts::getOrgs()`** to dedupe case-insensitively via `Map<lowercase, canonical>`, with framework-root entries winning over state-dir entries (framework is source of truth, state is derived). Same org with drifted casing across both dirs now returns ONE entry using the framework casing. Silences the syncAll log noise.

## Tests

- 8 tests in `tests/unit/utils/org.test.ts`: exact match / case drift / no match / missing orgs dir / empty input / case-sensitive FS (both casings exist as distinct orgs — exact-match wins) / non-dir entry with matching name / empty orgs dir.
- 5 tests in `tests/unit/dashboard/config-getorgs.test.ts`: framework-only / state-only / case drift between the two (framework wins) / multi-org production scenario regression guard / empty input returns empty list.

- **Stale ghost dirs are NOT cleaned up by this patch**. Operators with pre-existing `~/.cortextos/<instance>/orgs/<lowercase>/` trees should inspect them for a `knowledge-base/config.json` (may contain a redundant `GEMINI_API_KEY` copy) and remove by hand after confirming the canonical CamelCase dir is in active use. Automatic cleanup feels risky — a file-moving patch against operator state is a separate decision.
- **Case-sensitive filesystems can legitimately host both casings as distinct orgs**. The normalizer respects that: an exact-case match always wins over the case-insensitive scan, so `ElementOne` and `elementone` stay distinct when both exist. Only drift from a missing lowercase variant to an existing CamelCase one (the ghost scenario) gets collapsed.
- **Not retroactive for existing ChromaDB collections**. If an operator already has both `shared-acmecorp` and `shared-AcmeCorp` collections, queries after this patch all land on the canonical casing — the lowercase collection becomes orphaned. Safe to drop once verified empty.

(None found — the helper is fully generic and takes frameworkRoot as a parameter.)

---

**Test suite**: full 464/464 pass. Cherry-pick onto upstream/main a608a5d was clean. Scrubs applied: real org names `ElementOneSound` / `elementonesound` and real agent-lane org `agentnet` used as fixture names throughout the original patch were replaced with generic `AcmeCorp` / `acmecorp` / `widgetco` across both source comments and tests; one expected-sort-order assertion updated to match the renamed fixtures' alphabetization.

### DIFF
--- a/dashboard/src/lib/config.ts
+++ b/dashboard/src/lib/config.ts
@@ -120,26 +120,34 @@ export function getAgentDir(name: string, org?: string): string {
 // -- Discovery functions --
 
 export function getOrgs(): string[] {
-  const orgs = new Set<string>();
-
-  // Check state dir (CTX_ROOT/orgs/)
-  const stateOrgsDir = path.join(CTX_ROOT, 'orgs');
-  if (fs.existsSync(stateOrgsDir)) {
-    for (const d of fs.readdirSync(stateOrgsDir, { withFileTypes: true })) {
-      if (d.isDirectory()) orgs.add(d.name);
-    }
-  }
-
-  // Also check project/framework root (CTX_FRAMEWORK_ROOT/orgs/)
-  // This is where CLI creates org config files (context.json, knowledge.md)
+  // Read framework root FIRST — it is the source of truth for org naming.
+  // When the same org exists in both dirs with drifted casing (e.g. a ghost
+  // `acmecorp/` in state + canonical `AcmeCorp/` in framework),
+  // we keep the framework casing and discard the state-dir variant. Without
+  // this, dashboard sync hits both names and floods the log with lookup
+  // failures against the non-existent lowercase dir.
   const frameworkOrgsDir = path.join(CTX_FRAMEWORK_ROOT, 'orgs');
-  if (frameworkOrgsDir !== stateOrgsDir && fs.existsSync(frameworkOrgsDir)) {
+  const stateOrgsDir = path.join(CTX_ROOT, 'orgs');
+
+  // Map lowercase key -> canonical casing. Framework entries win over state
+  // entries. Within a single dir, we trust fs.readdirSync uniqueness.
+  const byLower = new Map<string, string>();
+
+  if (fs.existsSync(frameworkOrgsDir)) {
     for (const d of fs.readdirSync(frameworkOrgsDir, { withFileTypes: true })) {
-      if (d.isDirectory()) orgs.add(d.name);
+      if (d.isDirectory()) byLower.set(d.name.toLowerCase(), d.name);
     }
   }
 
-  return Array.from(orgs);
+  if (frameworkOrgsDir !== stateOrgsDir && fs.existsSync(stateOrgsDir)) {
+    for (const d of fs.readdirSync(stateOrgsDir, { withFileTypes: true })) {
+      if (d.isDirectory() && !byLower.has(d.name.toLowerCase())) {
+        byLower.set(d.name.toLowerCase(), d.name);
+      }
+    }
+  }
+
+  return Array.from(byLower.values());
 }
 
 export function getAgentsForOrg(org: string): string[] {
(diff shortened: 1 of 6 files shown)


grandamenium:
**Pipeline Stage 1 Failed** — Static gates did not pass. See `pr-reviews/62/stage-results.json` for details.

grandamenium:
**Pipeline Stage 1–2 Passed** (static analysis). Score: 7/10. Recommendation: MERGE WITH CHANGES. See `pr-reviews/62/result.md` for specific required changes before merge.

grandamenium:
Re-reviewed 2026-04-14. Score: 7/10 — MERGE WITH CHANGES. Original blocker still open: 2 macOS tests failing in org.test.ts. Build passes, but npm test shows 2/464 tests fail: 'case drift' test expects normalizeOrgName to return canonical CamelCase casing, but on macOS (case-insensitive FS) existsSync('orgs/acmecorp') returns true when only 'orgs/AcmeCorp' exists, so the fast-path returns the input unchanged. Second failure: mkdir('orgs/acmecorp') throws EEXIST on macOS when 'orgs/AcmeCorp' already exists. Fix options: (a) mark these two tests skipIf(platform === 'darwin') with a comment, (b) use memfs mock for case-sensitive simulation, or (c) test in a case-sensitive Docker FS. The underlying fix logic is correct and valuable — just needs the test platform guard.

grandamenium:
Re-reviewed 2026-04-16. Score: 7/10 — MERGE WITH CHANGES. Build passes; 2/464 tests still failing (same 2 macOS case-insensitive filesystem tests from original review — org.test.ts). The fix: add skipIf(process.platform === 'darwin') guards on the case-sensitive scenarios, or use a mock FS approach. Underlying fix logic is correct and 11 other org tests pass. One small fix needed before merge.

grandamenium:
Superseded by #164 which builds on #131 (now merged) and applies only the KB wire-in and dashboard org dedup. The original org.ts had an existsSync macOS bug that #131 corrected. Thanks for identifying the problem!

grandamenium:
Pipeline passed (1 iteration, static analysis + unit test verification). Tested: normalizeOrgName() resolution order (exact match → case-insensitive scan → passthrough), ghost dir prevention at all 3 KB call sites, getOrgs() case-insensitive dedup with framework-root-wins policy. 464/464 unit tests pass. Merge score: 8/10. Ready for review.

PR #63, opened 2026-04-14
feat(approvals): Telegram inline-button approvals on the activity channel

ClintMoody:
# feat(approvals): Telegram inline-button approvals on the activity channel

**Branch**: `pr/telegram-inline-approvals`
**Commit**: `60cd124` (squash of 3 incremental commits: feature + timing fix + path fix)

---

## Problem

Every approval required opening the dashboard in a browser. Operators wanted inline Approve/Deny buttons in the Telegram activity channel — the same surface already receiving one-way activity notifications. Fan-out had to be reliable from short-lived CLI callers, and the path resolution for `activity-channel.env` had to actually point at the framework directory where that file lives.

## Root cause

Three gaps:
1. **No fan-out**: approval creation did not notify the activity channel.
2. **Fire-and-forget timing**: initial fan-out pattern was un-awaited. Works for long-running callers (daemon, dashboard) because the event loop stays alive; fails for CLI callers that exit immediately after `console.log(id)`.
3. **Wrong filesystem root**: `orgDir` was derived from `paths.ctxRoot` (runtime state dir), but `activity-channel.env` lives under the framework root. Every post silently failed as "not configured" and a silent `.catch` suppressed the signal.

### Architecture

Activity-channel bot was previously outbound-only. To route inline button callbacks, the org's orchestrator now instantiates a second `TelegramPoller` bound to `ACTIVITY_BOT_TOKEN` alongside its primary bot poller. Non-orchestrator agents are unaffected. Callback clicks route to fast-checker's approval resolver, which moves the approval from `pending/` to `resolved/` via the same `updateApproval` call path the dashboard UI uses — dashboard status reflects Telegram-driven resolution with zero sync code.

**Trade-off**: callback polling is coupled to the orchestrator's lifecycle. If the orchestrator restarts, approval button clicks drop until it comes back up. Approvals created during the restart window still persist and resolve via the dashboard. A dedicated singleton or Telegram webhook is a reasonable future direction if the coupling becomes painful.

### Timing — await the fan-out

`postApprovalToActivityChannel` now RETURNS the `postActivity` promise (still with `.catch(() => undefined)` for error absorption). `createApproval` becomes async and awaits it before returning `approvalId`. The approval file still writes to `pending/` BEFORE the await, so the file-based state-of-record is intact regardless of fan-out outcome. Both CLI call sites (`create-approval`, `run-experiment` auto-approval path) become async action handlers.

### Path resolution — frameworkRoot, not ctxRoot

`createApproval` gains an optional `frameworkRoot` parameter (7th arg). Fallback chain: explicit arg → `process.env.CTX_FRAMEWORK_ROOT` → SKIP WITH WARN. The `paths.ctxRoot` fallback that caused the original bug was REMOVED intentionally — silently using a known-wrong path is worse than skipping loudly. `postApprovalToActivityChannel` now inspects `postActivity`'s return to distinguish "posted OK" from "skipped because not configured" and emits a one-line warn on the skip path instead of eating it silently.

## Tests

- `tests/unit/bus/approval.test.ts` (NEW): async `createApproval` flow, env-fallback chain, skip-with-warn behavior, end-to-end resolve via `updateApproval`.
- `tests/unit/daemon/fast-checker.test.ts`: extended with approve/deny callback edit path, audit-trail formatter for the Telegram user identity (`by Alice (@alice)` / `by Alice` / `user 42` fallbacks), and lifecycle transitions.

Full suite: 466/466 pass.

- **Orchestrator coupling** (see Architecture). A follow-up can migrate to a dedicated poller singleton or Telegram webhook if the coupling proves painful.
- **Skip-with-warn** on missing `activity-channel.env` is intentional — orgs without activity-channel config keep working normally. Operators who want hard-fail can wrap the call site.
- **E2E live-test was done in a private chat**; no E2E harness against real Telegram is added. The unit tests cover the message-shape and state-transition contracts, not the actual Telegram roundtrip.
- **Audit trail format** (`via Telegram activity channel by <firstname> (@<username>)`) is hardcoded. If a deployment wants different wording, it's a string template in `fast-checker.ts`.

(None found. Audit-trail example in JSDoc already uses a neutral `Alice` placeholder.)

---

**Test suite**: 466/466 pass. Cherry-pick of 3 commits onto upstream/main a608a5d had 2 trivial auto-merges in `src/cli/bus.ts` (both resolved by git cleanly). Soft-reset + single squashed commit authored by ClintMoody. Scrubs applied:
- `src/bus/approval.ts` JSDoc "so Clint can approve/deny" → "so the operator can approve/deny"
- `src/daemon/fast-checker.ts` audit-format comment "by Clint (@clintm)" → "by Alice (@alice)"
- Test fixtures across 2 test files: requesting_agent `'bob'` → `'alice'`, Telegram user `{first_name: 'Clint', username: 'clintm'}` → `{first_name: 'Alice', username: 'alice'}`, assertion strings updated to match.

### DIFF
--- a/src/bus/approval.ts
+++ b/src/bus/approval.ts
@@ -5,19 +5,121 @@ import { atomicWriteSync, ensureDir } from '../utils/atomic.js';
 import { randomString } from '../utils/random.js';
 import { validateApprovalCategory } from '../utils/validate.js';
 import { sendMessage } from './message.js';
+import { postActivity } from './system.js';
+
+/**
+ * Build the inline keyboard posted to the activity channel alongside a
+ * newly-created approval. Two buttons (Approve / Deny) with callback_data
+ * keyed on the approval id so fast-checker's activity-channel callback
+ * handler can route them to updateApproval.
+ */
+function buildApprovalKeyboard(approvalId: string): object {
+  return {
+    inline_keyboard: [[
+      { text: '✅ Approve', callback_data: `appr_allow_${approvalId}` },
+      { text: '❌ Deny', callback_data: `appr_deny_${approvalId}` },
+    ]],
+  };
+}
+
+/**
+ * Post a newly-created approval to the org's activity channel with
+ * Approve/Deny inline buttons. Returns a promise that resolves once the
+ * post attempt has settled.
+ *
+ * Path resolution: activity-channel.env lives under the FRAMEWORK root
+ * (frameworkRoot/orgs/<org>/activity-channel.env), NOT the runtime state
+ * dir (ctxRoot/orgs/<org>/). The earlier version of this helper used
+ * paths.ctxRoot to derive orgDir, which silently resolved to the wrong
+ * filesystem root and caused every activity-channel post to fail as
+ * "not configured" — a bug that hid for hours because of the silent
+ * .catch below. Fallback chain is now: explicit frameworkRoot arg →
+ * process.env.CTX_FRAMEWORK_ROOT → SKIP WITH WARN (no further fallback;
+ * the paths.ctxRoot fallback that caused the original bug was removed
+ * deliberately per post-incident review — silently using a known-wrong
+ * path is worse than skipping loudly).
+ *
+ * Errors from postActivity (thrown rejections) are suppressed so
+ * activity-channel unreachability does not block approval creation. The
+ * "not configured" signal (postActivity returns false) is now logged as
+ * a visible warn — preserves the best-effort behavior but surfaces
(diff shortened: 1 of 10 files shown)


grandamenium:
**Pipeline Stage 1 Failed** — Static gates did not pass. See `pr-reviews/63/stage-results.json` for details.

grandamenium:
**Pipeline Stage 1–2 Passed** (static analysis). Score: 8/10. Recommendation: MERGE WITH CHANGES. See `pr-reviews/63/result.md` for specific required changes before merge.

PR #69, opened 2026-04-14
fix(daemon): classify PM2 shutdown PTY exits as planned stops, persist crash audit to restarts.log

noogalabs:
## Summary

Fixes a ghost crash-count inflation that occurred every time the cortextos daemon was shut down via PM2.

When PM2 sends SIGTERM to the daemon, the signal propagates to the whole process group and reaches each PTY child (Claude Code) before `stopAll()` can call `stopAgent()` on it. Those children exit cleanly (code 0) but arrive at `handleExit` with `stopRequested=false`, which previously classified each exit as a crash and incremented `.crash_count_today` by one per agent per PM2 restart.

**Fix:** `handleExit` now calls `isDaemonShuttingDown()` before applying crash logic. The method checks for a `.daemon-stop` marker written by `AgentManager.stopAll()` at shutdown start (already written for the SessionEnd crash-alert hook). If the marker exists and is < 60s old, the exit is classified as a shutdown casualty and swallowed — not a crash.

The 60s window prevents a stale marker from masking real crashes after a subsequent session.

**Also:** Crash and HALTED exits now append a structured line to `restarts.log` so operators have a durable audit trail alongside the planned SELF-RESTART / HARD-RESTART entries from `bus/system.ts`.

## Test plan

- [ ] Build: `npm run build` — clean
- [ ] Tests: `npm test` — all pass (2 pre-existing org-casing failures on macOS HFS+ unrelated to this PR)
- [ ] Manual: restart daemon via PM2 — confirm `.crash_count_today` is not incremented, confirm `restarts.log` is not written for PM2 shutdown exits
- [ ] Manual: trigger a real crash — confirm `.crash_count_today` increments and `restarts.log` records the exit

🤖 Generated with Claude Code


grandamenium:
Superseded by #57

PR #65, opened 2026-04-14
feat(daemon): ctx-exhaustion + frozen-stdout watchdog with auto hard-restart

clearworks-ai:
## Summary
- Adds context-exhaustion + frozen-stdout detection to `fast-checker.ts`
- Auto hard-restart (fresh session) when a stuck agent is detected
- Ports the frozen-detection pattern from legacy CRM's `fast-checker.sh`

Triggered by auditos tonight sitting at 99% ctx on the "How is Claude doing this session?" survey prompt (7 hits in stdout.log) with no auto-recovery. cortextOS daemon only had time-based (71h) and crash-recovery restarts; CRM had `FROZEN_RESTART_MAX_SECONDS` + `do_hard_restart` that cortextos lacked.

## Signals
1. **Session-survey prompt** in stdout.log tail (20KB) → hard-restart
2. **Stdout frozen** 30+ min while fast-checker considers agent active → hard-restart

## Guardrails
- Bootstrap grace: 10 min
- Cooldown: 15 min between triggers (can't ping-pong)
- Notifies user via Telegram before restarting

## New API
`AgentProcess.hardRestartSelf(reason)` — writes `.force-fresh` marker, then `stop()` + `start()`.

## Test plan
- [x] `npm run build` clean
- [x] `npm test` — 451 passing
- [x] `npx tsc --noEmit` clean
- [ ] Live verification: deploy to fleet, confirm watchdog fires on next ctx-exhaustion event

PR #71, opened 2026-04-14
feat(daemon): context-exhaustion + frozen-stdout watchdog; fix(telegram): Conflict error backoff

noogalabs:
## Summary

Two fixes for daemon reliability after observing PM2 restart cascades:

**1. Context-exhaustion + frozen-stdout watchdog** (`feat`)

Adds two new watchdog timers to `AgentProcess` that fire when Claude Code silently stalls:

- **Context-exhaustion watchdog** — polls `stdout.log` every 30s. If the log grows to > 50MB and the agent hasn't been active for > 10 minutes, assumes Claude Code has exhausted its context window and hard-restarts (new session, no `--continue`).
- **Frozen-stdout watchdog** — if stdout hasn't grown in > 20 minutes while the agent is in "running" status, assumes the PTY is frozen and hard-restarts.

Both use the new `hardRestartSelf(reason)` method (introduced alongside this in `agent-process.ts`) that writes `.force-fresh` + `.restart-planned` markers, appends to `restarts.log`, then `stop()+start()`.

**2. Telegram poller Conflict error backoff** (`fix`)

When PM2 restarts overlap, two `TelegramPoller` instances hit the same bot token simultaneously. Telegram returns HTTP 409 "Conflict: terminated by other getUpdates" — previously unhandled, causing a tight crash loop (22 restarts observed on 2026-04-12).

Fix: catch `Conflict` errors specifically in `pollOnce` and back off 10s before retrying. Also bumps `restart_delay` from 5s to 15s in `ecosystem.config.js` (generated file — actual value is what matters) so the old process fully exits before the new one starts polling.

## Test plan

- [ ] Build: `npm run build` — clean
- [ ] Tests: `npm test` — all pass (2 pre-existing org-casing failures on macOS HFS+ unrelated to this PR)
- [ ] Simulate frozen stdout: pause a Claude Code PTY for 20+ min — confirm hard-restart fires
- [ ] Simulate PM2 overlap: restart daemon twice rapidly — confirm no Conflict crash loop

🤖 Generated with Claude Code


grandamenium:
Superseded by #65

PR #91, opened 2026-04-15
fix(bus): auto-notify assignee on create-task (issue #78)

### LINKED ISSUE #78: create-task should auto-notify the assignee via send-message
Problem. `cortextos bus create-task` with assignee stores the task but does not notify the assigned agent. The agent only sees it when they happen to poll. In practice orchestrators must pair every create-task with a send-message call to the same agent, otherwise tasks are invisible.

Impact. Tasks created without a paired message silently stall. Orchestrators duplicate effort. New users hit this immediately.

Proposal. When `create-task` is called with assignee, automatically enqueue an inbox message to that agent with task id, title, description, priority, and a pointer to get-task. Opt-out with no-notify for bulk imports.

Repro. cortextos bus create-task test with assignee engineer - agent does not receive inbox message; silent until poll.

Flagged by a user after watching devops not see an assigned P0 task.

grandamenium:
## Summary

- When `cortextos bus create-task` is called with `--assignee`, automatically enqueue an inbox message to the assigned agent
- Agent receives: task ID, title, description snippet, priority
- No-op if assignee is the same as the creator (no self-notification)
- Closes #78
- 453/453 tests pass, build clean

### DIFF
--- a/src/cli/bus.ts
+++ b/src/cli/bus.ts
@@ -114,6 +114,13 @@ busCommand
       needsApproval: opts.needsApproval ?? false,
     });
     console.log(taskId);
+    // Auto-notify assignee so the task is visible immediately (issue #78)
+    if (opts.assignee && opts.assignee !== env.agentName) {
+      const assigneePaths = resolvePaths(opts.assignee, env.instanceId, env.org);
+      const desc = opts.desc ? ` — ${opts.desc.slice(0, 120)}` : '';
+      sendMessage(assigneePaths, env.agentName, opts.assignee, 'normal',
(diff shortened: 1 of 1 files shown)


grandamenium:
**Pipeline Stage 1 Failed** — Static gates did not pass. See `pr-reviews/91/stage-results.json` for details.

grandamenium:
**Pipeline Stages 1-2 Passed** (manual gate verification). Score: 7/10. Recommendation: MERGE. Auto-notifies task assignee on create-task — clean, additive fix. Missing error handling on sendMessage (no data loss risk since task is persisted first) and no test for notification branch. File a follow-up issue for both.

PR #49, opened 2026-04-13
feat(daemon): auto-recover from Anthropic rate-limit exits

noogalabs:
## Summary

- **`OutputBuffer.hasRateLimitSignature()`** — scans last 200 PTY output chunks for Anthropic rate-limit/overload patterns (`overloaded_error`, `rate_limit_error`, `rate limit`, `too many requests`, `quota exceeded`, `usage limit`, HTTP 529)
- **`handleExit()` rate-limit branch** — when signature detected: set status `rate-limited`, write `.rate-limited` marker to state dir, schedule restart after `rate_limit_pause_seconds` (default 18 000 s / 5 h) without incrementing `crashCount` or triggering the git watchdog
- **`rateLimitTimer`** — stored so a second rate-limit exit cancels the first timer before arming a new one, preventing overlapping timers from triggering a premature restart
- **Marker lifecycle** — `.rate-limited` is read non-destructively at prompt-build time and deleted only after `pty.spawn()` succeeds (same pattern as watchdog `readRecoveryNote`/`deleteRecoveryNote`); a spawn failure preserves the marker for the next attempt
- **Prompt injection** — both fresh-start and `--continue` prompts include a `RATE-LIMIT RECOVERY` notice so the agent knows the pause was planned, not a crash
- **Types** — `AgentConfig.rate_limit_pause_seconds?: number`, `AgentStatus.status` union gains `'rate-limited'`
- **Tests** — 9 new tests in `tests/unit/daemon/rate-limit-recovery.test.ts` covering: status transition, no crashCount increment, no watchdog call, marker file written, default and custom pause durations, crash path unchanged, prompt injection when marker present

## Test plan

- [ ] `npm run build` — TypeScript compiles cleanly
- [ ] `npm test` — 492 tests pass (9 new rate-limit recovery tests)
- [ ] Verify `hasRateLimitSignature()` patterns match real Anthropic error payloads
- [ ] Verify `rateLimitTimer` cancellation prevents overlapping timer races

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #109, opened 2026-04-15
fix(hooks): quiet hours + dedup + rate-limit reclassification for crash alert

revopsglobal:
## Problem

The SessionEnd crash-alert hook pages Telegram on every non-crash end type (`planned-restart`, `session-refresh`, `daemon-stop`, `user-*`) regardless of time, and has no deduplication.

A single misclassified crash loop — e.g. the Claude Code CLI weekly-limit wall, which surfaces as the status-bar line `You've used 100% of your weekly limit · resets 10pm (America/Los_Angeles)` rather than an API error body — results in a 🚨 CRASH Telegram alert every ~30 minutes all night. We hit this on our `revops-global` org yesterday when an agent exhausted the Claude Code weekly limit. 7 crashes → 7 middle-of-the-night buzzes before manual intervention.

The existing `OutputBuffer.hasRateLimitSignature()` (added in #49) only matches Anthropic **API error bodies** (`overloaded_error`, `rate_limit_error`, HTTP 529, etc.), not the **Claude Code CLI status-bar** format. Even once #49 lands, the hook itself has no time-of-day guardrail.

## What this PR changes

Three related changes to `src/hooks/hook-crash-alert.ts`, all scoped to the hook:

### 1. Stdout-based rate-limit reclassification

When no marker file matches and the exit would fall through to `crash`, the hook now scans the tail of `stdout.log` for rate-limit signatures and reclassifies the exit as a new `rate-limited` end type. Patterns matched (case-insensitive, ANSI stripped):

- `overloaded_error`
- `rate_limit_error`
- `rate limit`, `rate-limit`
- `too many requests`
- `quota exceeded`
- `usage limit`
- `weekly limit`, `5-hour limit`, `5h limit`
- `/used \d+% of your/` — the Claude Code CLI status-bar regex

This is deliberately duplicated from `OutputBuffer.hasRateLimitSignature()` because the hook runs as a standalone short-lived process and cannot share the daemon's in-memory buffer. Both detectors should stay in sync going forward.

The new `rate-limited` branch emits a calmer ⏳ message (`"paused — Anthropic rate limit hit. Will resume when the window resets."`) and is subject to quiet hours.

### 2. Quiet hours (22:00–07:00 America/Los_Angeles)

Routine end types are suppressed during the quiet-hour window:

- `planned-restart`
- `session-refresh`
- `daemon-stop`
- `user-restart`, `user-disable`, `user-stop`
- `rate-limited`

**Real unexpected `crash` events are deliberately NOT suppressed** — that is the only end type worth waking an operator up for.

Timezone resolution uses `Intl` via `toLocaleString('en-US', { timeZone: 'America/Los_Angeles' })` so it does not depend on the host TZ. The quiet-hour window is defined as module constants (`QUIET_HOUR_START_LA = 22`, `QUIET_HOUR_END_LA = 7`) so operators can fork and adjust easily. A follow-up could make these configurable per-agent via `config.json` — happy to do that in a second pass if you prefer.

### 3. Dedup within 10-minute window

Identical `(agent, type)` alerts within a 10-minute window are suppressed via a `.crash_alert_dedup.json` state file in the agent's state dir. A broken watchdog loop now produces at most one notification instead of a buzz storm.

`DEDUP_WINDOW_MS = 10 * 60 * 1000` is a module constant.

## What does not change

- **All events still log to `crashes.log`** regardless of whether the Telegram alert is actually sent. Post-hoc visibility is preserved — operators can `tail crashes.log` in the morning to see what happened.
- The existing marker-file detection (`.restart-planned`, `.session-refresh`, `.user-*`, `.daemon-stop`) is unchanged.
- The crash-count tracking (`.crash_count_today`) is unchanged and still only increments for the `crash` end type.
- The Telegram API call itself is unchanged.

## Out of scope (follow-up)

Matching the weekly-limit patterns in `OutputBuffer.hasRateLimitSignature()` and parsing `"resets Xpm"` to compute a smarter daemon pause duration. That requires changes to `src/pty/output-buffer.ts` and `src/daemon/agent-process.ts` which currently live behind #49 (not yet merged). I have that extension ready locally and will open it as a separate PR once #49 lands so the history is clean.

## Testing

- `npm run typecheck` — clean
- `npm run build` — success, `dist/hooks/hook-crash-alert.js` 6.47 KB
- `npm test` — 35 files, 490 tests pass
- **Field-tested**: deployed to the VM running our `revops-global` org. Stopped a 7×30-minute overnight alert loop on an agent that had hit the Claude weekly limit. The daemon kept respawning the agent (correct — crash recovery), but the hook correctly suppressed the misclassified alerts while logging each attempt to `crashes.log` for visibility.

No new test file is included because this hook does not currently have a test file in `tests/unit/hooks/`. I can add one if you want — the helper functions (`isQuietHoursLA`, `detectRateLimitInLog`, `shouldSuppressDedup`) are pure and easy to test in isolation. Let me know what you prefer.

### DIFF
--- a/src/hooks/hook-crash-alert.ts
+++ b/src/hooks/hook-crash-alert.ts
@@ -1,11 +1,103 @@
-
 /**
  * SessionEnd hook - crash alert via Telegram.
  * Categorizes session end type and sends notification.
+ *
+ * Behavior:
+ *   - Detects Anthropic weekly/5h rate-limit messages in stdout.log and
+ *     classifies the exit as "rate-limited" so it is suppressed rather than
+ *     spamming a 🚨 CRASH alert every 30 minutes while the daemon respawn
+ *     loop continues hitting the wall.
+ *   - Applies quiet hours (22:00-07:00 America/Los_Angeles) for routine end
+ *     types (planned-restart, session-refresh, daemon-stop, user-*,
+ *     rate-limited). A real unexpected crash still pages at night.
+ *   - Deduplicates identical alerts for the same agent within 10 minutes so a
+ *     broken watchdog loop results in at most one notification, not a buzz
+ *     storm.
  */
-import { existsSync, readFileSync, writeFileSync, appendFileSync, unlinkSync, mkdirSync } from 'fs';
+import { existsSync, readFileSync, writeFileSync, appendFileSync, unlinkSync, mkdirSync, statSync } from 'fs';
 import { join } from 'path';
 import { homedir } from 'os';
+
+const DEDUP_WINDOW_MS = 10 * 60 * 1000;         // 10 minutes
+const QUIET_HOUR_START_LA = 22;                 // 22:00 America/Los_Angeles
+const QUIET_HOUR_END_LA = 7;                    // 07:00 America/Los_Angeles
+
+// End types that are routine and should be suppressed during quiet hours.
+// "crash" is deliberately NOT in this list — a genuine unexpected crash at
+// 3am is worth waking up for.
+const QUIET_SUPPRESSED_TYPES = new Set([
+  'planned-restart',
+  'session-refresh',
+  'daemon-stop',
+  'user-restart',
+  'user-disable',
+  'user-stop',
+  'rate-limited',
+]);
+
+function isQuietHoursLA(now: Date): boolean {
+  const laString = now.toLocaleString('en-US', {
+    timeZone: 'America/Los_Angeles',
+    hour12: false,
+  });
+  const m = laString.match(/\d+\/\d+\/\d+,?\s+(\d+):/);
+  if (!m) return false;
+  const hour = parseInt(m[1], 10);
+  // Window wraps midnight: 22:00-23:59 OR 00:00-06:59
+  return hour >= QUIET_HOUR_START_LA || hour < QUIET_HOUR_END_LA;
+}
+
+/**
+ * Scan the tail of stdout.log for Anthropic rate-limit or weekly-limit
+ * signatures. Mirrors OutputBuffer.hasRateLimitSignature so the hook and the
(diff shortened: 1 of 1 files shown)


grandamenium:
**Pipeline Stages 1-5 Passed. Sandbox Verified. Score: 9/10 — MERGE.**

490/490 tests pass. 8/8 behavioral checks pass.

Fixes three confirmed production noise problems in one clean 1-file commit:

1. **Rate-limit reclassification**: Scans last 200KB of stdout.log for Anthropic rate-limit signatures. Classifies those exits as 'rate-limited' instead of 'crash' — stops false 🚨 CRASH alerts during respawn loops.

2. **Quiet hours (22:00–07:00 LA)**: Suppresses routine alert types (planned-restart, session-refresh, daemon-stop, user-*, rate-limited) during night hours. Real unexpected crashes still page at night — 'crash' is explicitly NOT in the suppressed set.

3. **Dedup (10-min window)**: Identical (agent, type) alerts suppressed within 10 minutes. Broken watchdog loops send at most 1 notification, not a buzz storm.

All boundary logic verified: quiet hours wrap midnight correctly, dedup state file written atomically, log scan capped at 200KB (OOM-safe), rate-limited type in suppressed set.

Minor note: quiet hours are hardcoded to America/Los_Angeles. Consider a config field follow-up if timezone changes.

grandamenium:
Live sandbox behavioral tests added (Stage 4 updated standard). 4/4 tests PASS: (1) rate-limit exit suppresses crash alert, (2) quiet hours logic correct at 3am/noon LA, (3) dedup suppresses within 10-min window, (4) real crash bypasses quiet hours. Score: 9/10 MERGE confirmed.

PR #68, opened 2026-04-14
feat(daemon): persist cron fire timestamps and add gap-detection nudge (issue #67)

grandamenium:
## Summary

- Adds `src/bus/cron-state.ts` — file-backed cron fire timestamp registry at `state/<agent>/cron-state.json`. Agents report each cron execution via `cortextos bus update-cron-fire <name> --interval <interval>`.
- Adds `AgentProcess.scheduleGapDetection()` — background daemon loop that polls `cron-state.json` every 10 minutes. If any recurring cron has been silent >2x its expected interval, injects a `[SYSTEM]` nudge into the agent PTY to restore the missing cron.
- Adds `cortextos bus update-cron-fire` CLI command for agent self-reporting.
- 13 new unit tests for the cron-state module. 464/464 passing.

## Why

Context compression silently drops in-session `CronCreate` schedules. 8 confirmed fleet-wide dead zones in TW5-TW6. PR #26 (fast-checker watchdog) was insufficient — the fast-checker session itself goes dark during compression. This fix lives in the daemon process which survives all session resets.

**Key design decision:** agents use `config.json` as source of truth (already there) and self-report fires via a bus command (same pattern as `update-heartbeat`). No daemon-side `CronCreate` interception needed.

## Test plan

- [x] `npm run build` — compiles cleanly
- [x] `npm test` — 464/464 passing (13 new cron-state tests)
- [ ] Manual: write a stale `cron-state.json` (last_fire = 14h ago for a 6h cron), verify nudge appears in agent PTY within 10 minutes
- [ ] Manual: call `cortextos bus update-cron-fire heartbeat --interval 6h`, verify `cron-state.json` is updated

Closes #67

## Architecture note

`scheduleGapDetection()` uses the same lifecycle-generation guard as `scheduleCronVerification()` — it self-terminates cleanly on agent restart without leaving stale goroutines.

PR #114, opened 2026-04-16
fix(daemon): prevent duplicate crons on rapid session restarts

noogalabs:
Strengthens the restart prompts (boot + --continue) to require explicit CronList-first-then-skip-duplicates behavior. Previous wording (`Run CronList first — no duplicates`) was too soft — under rapid restarts some agents called CronCreate without checking, accumulating duplicates (72 dupes from 9 restarts in one reported incident). 1 file, text-only change.


grandamenium:
**Pipeline Stages 1-5 Passed. Score: 8/10 — MERGE.**

+2/-2 lines in agent-process.ts: adds explicit CronList-check-before-create dedup instructions to both buildFreshPrompt() and buildContinuePrompt(). Instructs agents to call CronList first and skip any cron whose prompt text already appears — preventing duplicate schedule accumulation during rapid --continue restarts. Build passes, tests pass.

Note: This is a prompt-level fix (instructs Claude) rather than code-level prevention. Effective and additive, but a daemon-level dedup mechanism would be a stronger long-term solution. This is a good immediate fix for a confirmed production issue. Complements PR #68 (cron gap-detection).

grandamenium:
Pipeline passed (static analysis + prompt text verification). Verified in built dist: both buildBootPrompt() and buildContinuePrompt() contain CRITICAL DEDUP wording with explicit CronList-first-then-skip semantics. Text change is correct and complete. Merge score: 8/10. Ready for review.

PR #116, opened 2026-04-16
feat(task): atomic claim, audit log, dependency DAG, semantic compaction

ClintMoody:
## Summary

Four beads-inspired enhancements to the `bus-task` system, each shipped as its own commit so reviewers can walk them independently. A 5th commit fixes two atomicity defects found during an internal verification pass (see commit 5 below).

### 1) Atomic `claim-task` — prevents two agents double-picking the same task

`cortextos bus claim-task <id> [--agent <name>]` atomically marks a pending task in_progress and assigns it to the claiming agent. Race protection via a POSIX O_EXCL lock file at `<taskDir>/.claims/<id>.claim`. First writer wins, second gets a clear "already claimed by <owner>" error. Idempotent for same-agent re-claim. Rolls back the lock if the task-JSON write fails so a retry can succeed.

### 2) Append-only JSONL audit log per task

Every state transition (create / claim / update / complete) appends one JSONL line to `<taskDir>/audit/<taskId>.jsonl`. `TaskAuditEntry` carries `{ts, event, agent, from?, to?, note?}`. `completeTask` records the `--result` string in the note field so the full lifecycle (including output) is replayable from disk. Best-effort writes — a failing audit log never blocks the real operation. Corrupt lines are skipped during replay. New `cortextos bus task-history <id> [--json]` CLI prints a formatted table or raw JSONL.

### 3) Dependency DAG — first-class `blocks` / `blocked_by` edges

Tasks can declare blockers:
- `create-task --blocked-by <id1,id2>` — task can't start until listed tasks complete
- `create-task --blocks <id1,id2>` — this task blocks listed tasks from starting
- Symmetric edges stored at create-time (declaring "A blocked_by B" also mutates B to add A to its `blocks` list) so queries in either direction are cheap
- `checkTaskDependencies(paths, id)` and `cortextos bus check-deps <id>` return open blockers with their current status, plus a `'missing'` marker for dangling references
- `list-tasks --respect-deps` orders unblocked tasks before blocked ones, preserving created_at DESC within each bucket
- Cycle detection runs at `createTask` BEFORE any disk write — cycles never land on disk, even as partial state

### 4) Semantic compaction — archive old completed tasks

`cortextos bus compact-tasks [--older-than <days>] [--dry-run]` (default 30d). For each completed task past the cutoff, appends a one-line summary to a monthly `<taskDir>/archive-YYYY-MM.jsonl` and removes the active JSON. **No LLM calls, no external deps** — the summary is just `{id, title, org, assigned_to, completed_at, archived_at, result}`. Clean inputs are supplied via `complete-task --result`.

**Critical guard**: never archives a task that is in the **transitive** `blocked_by` closure of any open task. In `A <- B <- C` with C still open, both A and B are preserved even after aging past cutoff — completion of the full downstream chain is required before the blocker chain becomes eligible. Compaction must not orphan dependency references.

Audit logs at `<taskDir>/audit/<id>.jsonl` are **preserved through compaction** — full lifecycle history survives even when the active task JSON is removed.

### 5) Verification-pass fixes (cycle-check atomicity + transitive guard)

Two correctness defects caught during a stress-test pass:

- **Cycle-check atomicity**: `createTask` was writing the task JSON BEFORE running `detectCycleOrThrow`. A cycle-rejected creation left a dangling task on disk with a one-way edge and no symmetric peer mutation. Ordering fixed: validate → write task → mutate peers → audit. Walker grew a `virtual` parameter describing the not-yet-written task so chains that pass through the new task are still detectable.
- **Transitive-blocker guard**: `compactTasks` was only preserving *direct* blockers of open tasks, leaving deeper-chain blockers eligible for archive. Fixed via DFS transitive closure from every open task's `blocked_by` list.

Both fixes covered by explicit regression tests asserting zero partial state on disk (cycle case) and full-chain preservation (compaction case).

## Breaking changes

None. New fields on the `Task` interface (`blocks`, `blocked_by`) are optional — existing task JSONs load unchanged.

## Tests

~25 new unit tests across `tests/unit/bus/task.test.ts` covering every feature end-to-end plus the two regression cases from the verification pass.

Full suite: **608/608 green**, `npx tsc --noEmit` clean, `npm run build` clean.

## Test plan

- [x] `npm run build` clean
- [x] `npm test` — 608/608 pass
- [x] E2E: claim-task, task-history, check-deps, compact-tasks --dry-run all behave as documented against a real agent's tasks dir
- [ ] Operator spot-check: claim a task from one session while another session tries to claim — second gets rejected with owner name

### DIFF
--- a/src/bus/task.ts
+++ b/src/bus/task.ts
@@ -20,6 +20,8 @@ export function createTask(
     project?: string;
     needsApproval?: boolean;
     dueDate?: string;
+    blockedBy?: string[];
+    blocks?: string[];
   } = {},
 ): string {
   const {
@@ -29,6 +31,8 @@ export function createTask(
     project = '',
     needsApproval = false,
     dueDate = '',
+    blockedBy = [],
+    blocks = [],
   } = options;
 
   validatePriority(priority);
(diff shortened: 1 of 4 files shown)


grandamenium:
**Pipeline + Behavioral Verification PASS. Score: 9/10 — MERGE.** 608/608 tests (new task test suite). 10/10 behavioral checks: atomic claim (wx/O_EXCL), idempotent re-claim, appendTaskAudit, TaskAuditEntry, audit best-effort, detectCycleOrThrow, dependency DAG, .claims/ separation, readTaskAudit all verified. Three strong features for multi-agent coordination. Note: includes community/ path restructuring for two noogalabs skills — verify this is intentional with Boris before merge.

PR #118, opened 2026-04-16
feat(daemon): add Gmail watch with restart-safe persistence and message dedup

noogalabs:
## Problem

Gmail watch had two bugs that caused a re-injection loop:

1. **Restart resets interval** — `gmailLastCheckedAt` was in-memory only. Every agent bounce set it to `0`, triggering an immediate re-scan of the same 100 emails.

2. **No message-ID dedup** — the same unread messages re-injected on every poll cycle (default 15 min) as long as Claude hadn't marked them read yet.

## Fix

**1 file changed: `src/daemon/fast-checker.ts`**

- `gmailLastCheckedAt` is now persisted to `state/<agent>/gmail-last-checked.txt` (epoch ms) and loaded on constructor init. Restart no longer resets the interval.

- A delivered-message-ID `Map` with a 2h TTL is persisted to `state/<agent>/gmail-delivered-ids.json`. IDs are filtered before each poll, added after successful delivery, and pruned in-memory before each filter pass. Messages already injected won't re-inject until the TTL expires (by which point Claude should have marked them read).

- Load/save failures log a warning so operators can see when persistence is broken, rather than silently falling back to the buggy in-memory behaviour.

## Test plan

- [ ] Start agent with `gmailWatch` configured, verify `gmail-last-checked.txt` is written after first poll
- [ ] Restart agent, verify `gmailLastCheckedAt` is restored and poll does not fire immediately if interval hasn't elapsed
- [ ] Verify `gmail-delivered-ids.json` is written after delivery and IDs are filtered on subsequent polls
- [ ] Verify entries older than 2h are pruned from the in-memory map before filtering

### DIFF
--- a/src/daemon/fast-checker.ts
+++ b/src/daemon/fast-checker.ts
@@ -3,7 +3,7 @@ import { execFile } from 'child_process';
 import { join } from 'path';
 import { createHash } from 'crypto';
 import type { InboxMessage, BusPaths, TelegramMessage, TelegramCallbackQuery } from '../types/index.js';
-import { checkInbox, ackInbox } from '../bus/message.js';
+import { checkInbox, ackInbox, sendMessage } from '../bus/message.js';
 import { updateApproval } from '../bus/approval.js';
 import { AgentProcess } from './agent-process.js';
 import type { TelegramAPI } from '../telegram/api.js';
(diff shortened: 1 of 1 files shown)


grandamenium:
Pipeline passed (1 iteration, static analysis). Tested: Gmail watch persistence (restart-safe lastCheckedAt + 2h message-ID dedup), 20-message cap, load/save failure observability, TTL pruning order. All behavioral assertions passed. Merge score: 7/10. Ready for review.

PR #136, opened 2026-04-16
feat: add agentcard-purchase skill to community catalog

revopsglobal:
## Summary

- Adds `agentcard-purchase` community skill for approval-gated virtual Visa card issuance
- Enables agents to autonomously complete purchases (domains, SaaS subscriptions, API credits) with human approval gates
- Uses [AgentCard.sh](https://agentcard.sh) MCP tools for card creation, checkout details, and lifecycle management
- Depends on the existing `approvals` skill for `financial` category authorization

## What it does

1. Agent needs to make a purchase, triggers the skill
2. Creates a `financial` approval via the approvals skill
3. Blocks the task and notifies the user
4. On approval, issues a scoped single-use virtual Visa card via AgentCard MCP
5. Retrieves PAN/CVV/expiry for checkout
6. Completes purchase, logs result, card auto-closes

## Review checklist

- [x] Directory name matches `name` in frontmatter
- [x] `description` is one sentence and specific
- [x] `triggers` array has 14 phrases
- [x] `external_calls` lists `mcp.agentcard.sh` and `api.agentcard.sh`
- [x] No hardcoded file paths
- [x] No secrets, tokens, API keys, or personal data
- [x] No destructive shell patterns
- [x] Tested on real agents (RevOps Global orchestrator + analyst)

## Dependencies

- `approvals` skill (already in community catalog)
- AgentCard CLI (`npm i -g agentcard`) + MCP server (`agent-cards setup-mcp`)

Generated with [Claude Code](https://claude.com/claude-code)

PR #140, opened 2026-04-16
feat(telegram): include recent conversation history in message context

revopsglobal:
## Summary

- Adds `buildRecentHistory()` to `logging.ts` that reads the last N messages from both `inbound-messages.jsonl` and `outbound-messages.jsonl`, sorted by timestamp
- Adds optional `recentHistory` param to `FastChecker.formatTelegramTextMessage()` in `fast-checker.ts`
- Wires it up in `agent-manager.ts` so every inbound Telegram message now includes a `[Recent conversation:]` block with the last 6 exchanges

## Problem solved

When multiple senders (director, analyst, etc.) write to the same Telegram chat and the user replies, the receiving agent had no visibility into what was said immediately before the reply. This caused context gaps where the agent would ask clarifying questions the user had already answered implicitly through prior messages in the thread.

## Behavior

Each inbound Telegram text message now includes:
```
[Recent conversation:]
[director]: Morning review fires at 8am PT
[Greg]: Morning review should be at 7am
```

Capped at 6 entries (configurable), 200 chars per message, filtered to the relevant chat_id.


grandamenium:
## Review: NEEDS CHANGES

Good idea, three concrete issues to fix before merge.

### Issue 1: Hardcoded `'Greg'` as speaker name (blocker)
In `buildRecentHistory()` in `logging.ts`, inbound messages are labeled with the hardcoded string `'Greg'`:
```typescript
readLines(inboundPath, 'Greg');
```
This is clearly a personal name from your own installation. It needs to use either:
- The configured admin username (`process.env.ADMIN_USERNAME ?? 'user'`)  
- Or a generic label like `'User'`

As-is, every cortextOS install in the world would label the human as "Greg."

### Issue 2: catalog.json cosmetic reformat (blocker)
Same issue noted on #136: most of the 413 additions are a reformatter run over all existing catalog entries (em-dashes → `\u2014`, inline `tags` arrays → multi-line). This inflates the diff, creates merge conflicts for other PRs, and is not your call to make. Please revert all catalog.json changes except the `updated_at` timestamp update if that's intentional.

### Issue 3: Token budget concern (non-blocking, but worth documenting)
Every inbound Telegram message now injects up to 6 × 200 chars = 1,200 chars of history. For high-volume agents (matttallybot handling frequent messages, for example), this adds cost on every single message, not just replies. Recommended: either make this opt-in per agent (config flag `telegram.inject_history: true`) or only inject when there is a `reply_to_message` (i.e. the user is replying to something, which is the scenario you're solving).

Fix issues 1 and 2, consider issue 3, and we can merge quickly.

revopsglobal:
Thanks for the review. Pushed fix in 5fb3692:

- **Issue 1:** `buildRecentHistory` now labels inbound messages with `process.env.ADMIN_USERNAME ?? 'User'` instead of the hardcoded `'Greg'`.
- **Issue 2:** Reverted catalog.json cosmetic reformat — only the new `agentcard-purchase` entry and `updated_at` bump remain. Diff shrank from 413/54 to 14/3 on that file.
- **Issue 3:** History block is now only injected when `msg.reply_to_message` is set, which is the multi-sender scenario the PR was solving. Skips the ~1.2KB token cost on every other inbound.

Also cleaned up some stray `ulimit` output that had ended up in the JSDoc comment for `buildRecentHistory`. Build green. Ready for re-review.

revopsglobal:
Rebased onto latest main and addressed all three issues:

**Issue 1 (hardcoded 'Greg')** — `buildRecentHistory()` now uses `process.env.ADMIN_USERNAME ?? 'User'` for the inbound speaker label.

**Issue 2 (catalog.json reformat)** — reverted all cosmetic changes; diff is now only the new `agentcard-purchase` entry plus the `updated_at` bump.

**Issue 3 (token budget)** — history injection is now gated on `msg.reply_to_message`, so the ~1.2KB context only gets added when the user is actually replying to a prior message (the scenario this PR solves). High-volume agents don't pay the cost on every inbound.

Rebase also picked up #141's reaction handler (the earlier branch was behind main and was inadvertently deleting `formatTelegramReaction` / `onReaction` — that's no longer the case). `npm run typecheck` clean.

grandamenium:
Superseded by #148 which applies the same fixes and preserves the conversation history feature. Merged via #148.

PR #161, opened 2026-04-17
fix(settings): add missing tools to allowlist to prevent idle boot loops

grandamenium:
## What this does

Adds 8 commonly-used tools to the default agent `settings.json` allow list:
`Glob`, `Grep`, `ToolSearch`, `CronCreate`, `CronList`, `CronDelete`, `Skill`, `Agent`

These tools are called in every agent boot sequence but were absent from the template, meaning any agent running **without** `--dangerously-skip-permissions` (manual `claude` runs, CI, non-daemon sessions) would trigger the catch-all `PermissionRequest` hook for each one.

Note: `--dangerously-skip-permissions` (set by the daemon via agent-pty.ts) suppresses `PermissionRequest` hooks entirely, so daemon-managed agents are not affected by the missing entries. This fix covers the non-daemon case.

## Changes

**`templates/agent/.claude/settings.json`**
Adds 8 missing tools to `permissions.allow`.

**`src/cli/bus.ts`**
New command: `cortextos bus fix-agent-settings`
- Scans all agent directories under `orgs/`
- Adds missing allowlist entries to existing `settings.json` files in one pass
- `--dry-run` flag to preview changes

## Migration for existing agents

```bash
cortextos bus fix-agent-settings
cortextos restart <agent-name>
```

## Test plan

- [x] `npm run build` passes
- [x] `npm test` — 624/624 tests pass
- [x] `cortextos bus fix-agent-settings --dry-run` correctly identifies all agents needing the update

### DIFF
--- a/src/cli/bus.ts
+++ b/src/cli/bus.ts
@@ -1999,11 +1999,12 @@ busCommand
 
 // --- fix-agent-settings ---
 
-program
+busCommand
   .command('fix-agent-settings')
   .description('Patch all agent settings.json files to include the full recommended permissions allow list')
   .option('--dry-run', 'Show what would be changed without writing')
   .action((opts: { dryRun?: boolean }) => {
+    const { existsSync: fsExists, readdirSync: fsReaddir, readFileSync: fsRead, writeFileSync: fsWrite } = require('fs');
     const env = resolveEnv();
     const frameworkRoot = env.frameworkRoot || process.cwd();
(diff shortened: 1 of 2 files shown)


grandamenium:
Closing per James — root cause diagnosis was incorrect for the daemon case. Will reopen if the non-daemon allowlist improvement is wanted separately.

PR #172, opened 2026-04-18
feat(bus): verify-message CLI — F10 phantom-coordination-message mitigation

vbatistajm:
## Summary

Adds `cortextos bus verify-message <msg_id>` — a standalone gate any agent can run before acting on a claimed inbound message at merge-grade (apply, deploy, approval-lift). Closes the provenance-verification gap behind F10 (2026-04-18 07:17 UTC phantom-triggered apply).

## Failure mode this closes

Recipient agent sees a high-trust inbound `=== AGENT MESSAGE from X ===` block in its conversation context and executes a gated action. Today, the recipient has no verification step between "saw message" and "acted." F10 showed this can be exploited by model-side confabulation — the model generates a fake AGENT MESSAGE block as its own assistant output, reads it back on the next turn as if it were real user input, and triggers a merge-grade operation on the fabricated instruction.

`verify-message` makes every merge-grade recipient prove the message crossed the bus before acting.

## Three independent checks

Implemented in `src/bus/verify.ts`, surfaced via `src/cli/bus.ts`. Full contract lives in `orgs/cleverwave/agents/ana/surfaces/verify-message-cli-spec.md` (already KB-ingested under shared-cleverwave scope).

1. **Bus-file audit.** Legit messages hit disk under `inbox/<recipient>/`, then move to `inflight/` then `processed/`. HMAC-rejected messages land in `inbox/<recipient>/.errors/`. A message that appears in none of these four locations never crossed the bus. Additionally, the file's `.from` field must match the sender segment in msg_id — mismatch is tampering (P5).

2. **Timestamp sanity.** msg_id's epoch-ms prefix must be within `[now - maxAgeDays, now + futureSkewSeconds]`. Defaults: 7 days past (matches cron expiry window), 60 s future (tolerates NTP drift and laptop sleep-resume, but two orders of magnitude tighter than the 5180 s future skew observed in the F10 incident).

3. **Sender event cross-check.** Since commit 86bc202, every real `sendMessage` emits a `message/agent_message_sent` event (or `action/telegram_sent`) into the sender's `analytics/events/<sender>/<date>.jsonl` with `metadata.msg_id`. We grep for that line. If the sender's analytics file doesn't exist at all, this check returns UNVERIFIABLE (pre-86bc202 rollout window). `--strict` promotes UNVERIFIABLE to PHANTOM for merge-grade gates.

Date edge case: messages sent within a minute of UTC midnight may have their event on the next day's file; we probe both `$MSG_DATE` and `$MSG_DATE+1`.

## CLI shape

```
cortextos bus verify-message <msg_id> \
  [--sender <name>] [--recipient <name>] \
  [--max-age-days 7] [--future-skew-seconds 60] \
  [--format text|json] [--strict]
```

**Exit codes:** 0 VERIFIED / 1 PHANTOM / 2 UNVERIFIABLE / 3 ERROR.

Emits an `action/message_verified` telemetry event with the verdict so we can measure how often the gate fires (spec §6 Q2).

## P1 is a first-class regression test

`tests/unit/bus/verify.test.ts` covers every case from ana's matrix:

- **T1-T4 positive** (VERIFIED / exit 0) — inflight message, processed message, now-skew, 6-day-old message.
- **P1-P5 phantom** (PHANTOM / exit 1) — including the literal 2026-04-18 msg_id `1776501820431-joma-7gxmn` which encodes the incident failure in its own epoch-ms prefix. No fixtures needed; if this test ever passes as VERIFIED, the mitigation has regressed.
- **U1 + strict flip** (UNVERIFIABLE / exit 2, then PHANTOM / exit 1 with --strict).
- **E1 malformed** (ERROR / exit 3).

15/15 green locally. Full suite (647 tests across 44 files) also green. `tsc --noEmit` clean.

## Live CLI smoke

```
$ node dist/cli.js bus verify-message 1776501820431-joma-7gxmn --recipient ana --format text
msg_id:       1776501820431-joma-7gxmn
sender:       joma
recipient:    ana
bus_file:     FAIL — not found in inbox/, inflight/, processed/, or .errors/ for recipient ana
timestamp:    PASS — … (skew -15092 s within ±60 s future / 7 d past)
sender_event: FAIL — no matching message_sent / agent_message_sent / telegram_sent event
RESULT:       PHANTOM (exit 1)
```

(Note: skew is now negative because the msg_id's epoch is ~4 h in the past at PR time, vs +5180 s future at incident detection time. Both bus_file and sender_event stay decisively FAIL regardless of when the check runs — those are the core phantom signals.)

## Scope

This is Option A from the spec (standalone CLI gate) **only**. Option B (pre-hooks on `apply-upstream`, `deploy-*`, `approve-*`) is queued as a follow-up per ana's recommendation — ship Option A, verify threshold defaults hold for ~1 week of real traffic, then add the automatic enforcement layer.

## Freeze-lift condition

Per joma's dispatch: P1 smoke test green is the gate for lifting the apply-grade freeze on ana and joma. Confirmed by both the vitest run above AND the live CLI smoke against the phantom msg_id. Ready for review.

## Files

- **NEW** `src/bus/verify.ts` — 268 lines, the verification logic.
- **NEW** `tests/unit/bus/verify.test.ts` — 230 lines, T1-T4 / P1-P5 / U1 / E1 coverage.
- `src/cli/bus.ts` — +69 lines, command registration + CLI wrapper.

No new runtime dependencies. No changes to existing bus internals. Follows the repo's TypeScript-strict, atomic-writes, no-adjacent-cleanup conventions.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>


vbatistajm:
Doc-patches committed + pushed (commit d910844) per ana spec §5 rollout rule. Four additive template changes, ADD-ONLY compliant:

- `templates/agent/AGENTS.md` — "Merge-grade action gate" section
- `templates/agent/HEARTBEAT.md` — Step 3b phantom sweep
- `templates/analyst/.claude/skills/upstream-sync/SKILL.md` — Step 2a authorization verification (exact failure point from 2026-04-18)
- `templates/orchestrator/AGENTS.md` — mirror + orchestrator upward-chain responsibility

Diffstat: +84/-0 across 4 files. No reflow, no collision risk.

**P1 regression on fresh build:**
```
$ node dist/cli.js bus verify-message 1776501820431-joma-7gxmn --recipient ana
RESULT: PHANTOM (exit 1)
bus_file: FAIL — not found in inbox/inflight/processed/.errors for recipient ana
sender_event: FAIL — no matching message_sent / agent_message_sent / telegram_sent event
timestamp: PASS — now 15716 s in the past (within 7-day max-age)
true exit: 1
```

Closes gate (2) from my side. Waiting on gate (1) Manu push approval + gate (3) scope-bleed resolution via joma's fast-forward push. Trio-gate closed separately (#172 review stays with joma + ana).

PR #176, opened 2026-04-18
fix(daemon): skip type=disabled cron entries in gap detector and verification

noogalabs:
## Summary

- Cron entries with `type=disabled` in `config.json` were passing through the `scheduleGapDetection()` and `scheduleCronVerification()` filter expressions, which only excluded `type=once`
- This caused gap-nudge messages to fire every 10 minutes for intentionally disabled crons
- Fix adds `c.type !== 'disabled'` to both filter sites and `'disabled'` to the `CronEntry` type union
- Two regression tests added alongside the existing `type=once` no-op tests

## Test plan

- [x] `npm run build` — clean
- [x] `npm test` — 752/752 pass (2 new tests added)
- [x] Existing `scheduleCronVerification()` once no-op test still passes
- [x] New: `scheduleCronVerification()` is a no-op for disabled crons
- [x] New: `scheduleGapDetection()` is a no-op for disabled crons

### DIFF
--- a/src/daemon/agent-process.ts
+++ b/src/daemon/agent-process.ts
@@ -622,7 +622,7 @@ export class AgentProcess {
     if (!crons || crons.length === 0) return;
 
     const recurringNames = crons
-      .filter(c => c.type !== 'once')
+      .filter(c => c.type !== 'once' && c.type !== 'disabled')
       .map(c => c.name);
     if (recurringNames.length === 0) return;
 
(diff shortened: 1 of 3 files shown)


grandamenium:
Pipeline passed (1 iteration). Build clean. New disabled-cron tests 2/2 pass. Pre-existing 'stop() PTY exit' test times out in worktree (environment artifact — confirmed passes on main). Fix correct: `c.type !== 'disabled'` added to both filter sites in scheduleGapDetection() and scheduleCronVerification(). Type union extended with 'disabled'. Merge score: 8/10. MERGE.

PR #192, opened 2026-04-20
fix(templates): add update-cron-fire to heartbeat cycle across all templates

qpartners2025-dev:
## Summary

Prevents false `[SYSTEM] Cron gap detected` nudges from being injected into agents whose crons are actually firing correctly. Complementary to the daemon-side fix in #183 — that PR is damage control (stagger nudges, extend verification window), this one is cause prevention (teach agents to register their cron fires so the detector never misfires in the first place).

## Background

The daemon's gap-detection loop reads `state/<agent>/cron-state.json` every 10 min and injects a nudge when it doesn't see a cron fire recorded within 2x the cron's interval. That file is written by `cortextos bus update-cron-fire <name> --interval <N>` — agents have to call it explicitly; the daemon does not observe in-memory CronCreate invocations.

Agents created from the existing templates never learned to make this call. Result: every cron fired cleanly, but the detector still nudged them because its own ledger was empty.

## Changes

- `templates/{agent,analyst,orchestrator}/HEARTBEAT.md` — step 1 of the heartbeat cycle now runs both `update-heartbeat` and `update-cron-fire heartbeat --interval 4h`. Added a short explanation of why both are required.
- `templates/{agent,analyst,orchestrator}/.claude/skills/cron-management/SKILL.md` — new "Gap-Detector Reset (on every cron fire)" section with examples.
- `templates/{agent,analyst,orchestrator}/.claude/skills/heartbeat/SKILL.md` — step 2 of the heartbeat cycle calls `update-cron-fire`; subsequent steps renumbered, cross-reference to cron-management added.

No runtime / daemon code changed. New agents created from the patched templates will comply automatically; existing agents are unaffected unless their per-agent files are edited in-place.

## Test plan

- [x] `npm run build` — clean (68ms)
- [x] `npm test` — 43 files / 634 tests passing (15.34s)
- [x] No `src/` or `dist/` changes; zero daemon-behaviour impact


qpartners2025-dev:
Heads up — this PR adds update-cron-fire calls to the heartbeat-cycle templates so the daemon gap-detector does not fire false alerts on schedule. Already running locally in our agent fleet for ~2 weeks without issues. Happy to make any adjustments if it would help land it. Workflow approval still pending on the maintainer sidE

grandamenium:

PR #208, opened 2026-04-22
fix(kb): drop hard 2min timeout that SIGTERM-killed large-collection ingest

loganbronstein:
## Summary
- `cortextos bus kb-ingest <file>` was SIGTERM-killing the python child on collections with 1000+ chunks while direct python invocation worked fine.
- Root cause: `execFileSync` in `ingestKnowledgeBase` hard-coded `timeout: 120000`. Incremental ingest against a large collection needs to dedup existing doc IDs, call the Gemini embedding API (network-bound), and upsert to ChromaDB. That routinely exceeds 2 minutes, so Node's timer fired and killed the healthy child.
- Fix: remove the default timeout (ingest is interactive, `stdio: 'inherit'`, operator sees progress and can Ctrl-C). Add optional `CORTEXTOS_KB_INGEST_TIMEOUT_MS` env override for CI/automation that needs to bound runtime, read from the merged env so `.env` / `secrets.env` overrides work the same way `MMRAG_CONFIG` and other KB vars do.

## Why not just bump the timeout to 30 minutes
A fixed ceiling is the wrong model. Ingest runtime scales with file size AND existing collection size, and the embedding API is network-bound. Any fixed number will be wrong for some user. The env var gives automation authors the knob they need without hurting interactive UX.

## Trade-offs
- Removed an accidental spend ceiling: with no default timeout, a runaway ingest can keep hitting the embedding API until the operator notices. The old 120s ceiling accidentally bounded that. Operators running ingest under cron / non-TTY parents should set `CORTEXTOS_KB_INGEST_TIMEOUT_MS` to a generous cap.
- Query path (`queryKnowledgeBase`) still has its 30s timeout — intentional; interactive read-path has a different latency budget.

## Tests
- 5 new vitest cases in `tests/unit/bus/knowledge-base.test.ts`:
  - no timeout passed to execFileSync by default
  - `CORTEXTOS_KB_INGEST_TIMEOUT_MS` positive-number override respected
  - invalid values (`''`, `'abc'`, `'0'`, `'-5'`, `'NaN'`) fall back to undefined
  - override read from merged env (`.env` / `secrets.env`), not just `process.env`
  - regression guard: simulated 5-minute subprocess does not throw under default config

## Test plan
- [x] `npm run build` — TypeScript compiles clean
- [x] `npx vitest run tests/unit/bus/knowledge-base.test.ts` — 10/10 pass
- [x] Adversarial review subagent (attack vectors: non-zero exit after long run, silent hang, stderr flood, large binary, permission error, env-var validation)
- [x] `codex:rescue` final pass — caught merged-env gap, fixed before commit


grandamenium:

grandamenium:
Closing — root cause already addressed by #309 (commit `3d7481a7`) merged into main. #309 raised the kb-ingest timeout from 120s to 600s default with `KB_INGEST_TIMEOUT_MS` env override (and added Python-side retry on Gemini 503s/UNAVAILABLE for the actual transient cause). The SIGTERM-kill failure mode this PR targets is fixed in main. If you still hit timeouts on >10 min ingests, please open a new issue with reproduction details — the override env knob is in place. Thanks for the contribution.

PR #215, opened 2026-04-22
fix(dashboard): strip HTML comments from agent fields + fix narrow-viewport dropdowns

loganbronstein:
## Summary

- **Agent card/detail**: IDENTITY.md ships with \`<!-- placeholder -->\` comments. Those were leaking into the rendered Name/Role/Emoji/Vibe/Work Style fields as raw markup. New \`stripHtmlComments\` util runs at parse time; empty-after-strip renders an italic muted \`(not set)\` placeholder instead.
- **Round-trip safe**: \`serializeIdentityMd\` / \`serializeSoulMd\` / \`serializeGoalsMd\` now skip rewriting a section whose new value equals the comment-stripped original, so template comments survive a save that only edits other fields.
- **Dropdown layout**: base-ui Select trigger went \`w-fit\` → \`min-w-0\` so it shrinks inside flex containers. Popup switched to \`min-w-(--anchor-width)\` with \`max-w-[calc(100vw-1rem)]\` and \`align=\"start\"\` + \`alignItemWithTrigger=false\` so it never overflows a narrow viewport and never overlaps the trigger.
- **Skills page**: SkillCard row wraps; Install button gets \`shrink-0\`. Fixes the half-screen click-through where the \"Select agent...\" dropdown was unreachable.

## Test plan

- [x] \`npx vitest run dashboard/src/lib/__tests__/markdown-parser.test.ts\` — 22/22 pass (5 new tests for strip, round-trip preservation, fenced-code preservation, untouched-save, inline-comment whitespace collapse)
- [x] \`npm run build\` (dashboard) — compiles clean
- [x] Callsite check: \`stripHtmlComments\` used in \`markdown-parser.ts\` at 4 sites outside tests
- [x] Adversarial review: no CRITICAL findings; user-visible double-space artifact fixed (regex now eats adjacent whitespace, replaces with a single space)
- [ ] Manual browser QA at 320 / 768 / 1024 width


grandamenium:

PR #225, opened 2026-04-23
feat: voice replies in dashboard /comms + Telegram

clicktoacquire:
## Summary

End-to-end voice-reply support for agent → user conversations. The user picks their preferred reply mode (Text / Voice / Both) in the dashboard `/comms` chat; every agent reply to the user then honors that preference automatically — on both the dashboard channel AND Telegram.

The logic is baked into `cortextos bus send-message`, so agents don't need any CLAUDE.md knowledge of this feature.

### UI (`channel-view.tsx`)
- Segmented toggle above the chat input: **Text / Voice / Both** (default Text).
- Confirmation dialog before switching — makes mode changes deliberate.
- Chat bar now accepts mp3/m4a/ogg/wav; audio attachments show an inline `<audio>` preview before send.

### New API (`/api/comms/reply-mode`)
- `GET ?agent=<name>` → `{ agent, mode }`.
- `PUT { agent, mode }` → persists one of `text | voice | both` to `{ctxRoot}/state/{agent}/reply-mode` (atomic write).

### CLI (`src/cli/bus.ts`)
- `send-message`: when the target is not a registered agent (= the dashboard/Telegram user), the CLI auto-reads reply-mode, generates an MP3 via edge-tts using the agent's configured voice, publishes it under `{ctxRoot}/dashboard-uploads/`, embeds the `/api/media/dashboard-uploads/...` URL in the inbox text, and mirrors the reply to Telegram using the agent's own `BOT_TOKEN`/`CHAT_ID`. Agent-to-agent sends are completely unchanged.
- `voice-reply`: was uploading MP3 to `/api/comms/upload` via curl, which always 401'd (middleware requires a session cookie), so the outbound audio URL was silently empty. Now writes the MP3 directly to `{ctxRoot}/dashboard-uploads/` — no HTTP round-trip, no auth to fight.
- New `reply-mode` subcommand for inspecting or scripting the preference from outside the dashboard.

### Infra
- `/api/media/[...filepath]/route.ts`: audio and video now return `Content-Disposition: inline` instead of `attachment`, so `<audio controls>` plays inline rather than triggering a download.
- `/api/comms/upload/route.ts`: added audio MIME types (mp3/ogg/m4a/wav); fixed a doubled `/api/media/media/...` path bug caused by writing to `{ctxRoot}/media/dashboard-uploads/` — the correct location is `{ctxRoot}/dashboard-uploads/`.

### Docs
- `templates/agent/.claude/skills/voice-reply/SKILL.md`: documents the preferred one-liner and voice-persona mapping.

## Backward compatibility

- Agent → agent sends: identical to before.
- Agent → user with `mode=text` (default): identical to before, plus a Telegram mirror that reuses the text.
- Agent → user with `mode=voice`/`both`: the new behavior.
- If edge-tts isn't installed or Telegram env vars aren't set, the feature gracefully degrades (bus message still delivers).

## Test plan

- [x] `tsc --noEmit` clean in both `src/` and `dashboard/`
- [x] CLI builds cleanly (`npm run build`)
- [x] Manual QA in a live dashboard: toggle renders, confirmation dialog opens, Cancel preserves state, Confirm persists server-side; `GET /api/comms/reply-mode` returns the set value.
- [x] End-to-end: setting mode=both on an agent then calling `cortextos bus send-message <user> normal "..."` produces both a bus entry with inline audio and a Telegram document with caption.
- [ ] Fresh install (no edge-tts): gracefully falls back to text.


grandamenium:
Hello — overnight sandbox audit flagged a title/commit mismatch + branch contamination.

PR title and tip-commit do not match — title says "feat: voice replies in dashboard /comms + Telegram" but tip is "feat(dashboard): drag-and-drop file attach in comms composer" (1 file, +52/-1). The 22-file/1335-line bulk diff in the PR view is unrelated branch history (other dashboard/comms work that already merged elsewhere) — the actual PR-tip commit does not implement voice replies.

Either:
(a) update PR title to match the tip-commit (drag-and-drop file attach), OR
(b) force-push a commit that actually implements voice replies, OR
(c) close+reopen with the canonical scope on a fresh branch off main (recommended — the current branch is contaminated with already-merged work).

Squash gate held until resolved.

PR #236, opened 2026-04-23
fix(pty): handle CC 2.1+ bypass-permissions prompt + fix trust handler false positive

chrisjackson2030-byte:
## Problem

Claude Code 2.1+ introduced an interactive confirmation prompt when spawning with `--permission-mode bypassPermissions`. The prompt presents two options with **"No, exit"** selected by default:

```
1. No, exit
2. Yes, I accept
```

The existing implementation attempted to auto-accept this prompt but failed silently in two ways:

**Bug 1 — ANSI escape codes break multi-word matching.**
Raw PTY data contains cursor-forward codes (`\x1b[1C`) between characters, so `"Bypass Permissions"` never appears as a continuous substring. The detection used `data.includes('Bypass Permissions')` which always returned false — the handler never fired.

**Bug 2 — Trust handler false positive on bypass menu text.**
The trust-folder handler used `data.includes('Yes')` to detect the "trust this folder?" prompt. This matched `"Yes, I accept"` in the bypass menu, causing `\r` (Enter) to be sent while "No, exit" was still selected — immediately killing the process. Symptom: agent crash loop at first spawn.

Together these produced a 10-crash loop on CC 2.1+, requiring manual PM2 process reset to recover.

## Root Cause

1. Multi-word substring matching is unreliable against raw PTY data — ANSI cursor-forward codes inject between characters at arbitrary positions.
2. No ordering guarantee between the bypass handler and the trust handler meant the trust handler could fire on the same data chunk, sending Enter on the wrong selection.

## Fix

**Detection:** Match on single-word `'Bypass'` — never split by ANSI codes in practice.

**Ordering:** Set `bypassAccepted = true` synchronously before any `setTimeout`, so the trust handler is blocked from firing on the same data chunk.

**Trust handler:** Changed from `data.includes('Yes')` to `data.includes('trust')` — the trust-folder prompt always contains "trust"; the bypass menu never does.

**Fallback probes:** Consolidated the separate 5s/8s timer pair into a shared `acceptPromptFallback()` at 2s and 5s, with the same per-prompt dedup guards applied.

**Spawn flag:** Changed `--dangerously-skip-permissions` to `--permission-mode bypassPermissions` — equivalent behavior, semantically cleaner, consistent with CC 2.1+ terminology.

## Test Verification

- Spawned fresh agent session under CC 2.1.116 with fix applied
- Bootstrap confirmed in PM2 logs: `[jarvis] Bootstrap complete. Beginning poll loop.`
- Zero crashes in 51 minutes of uptime (previously: crash loop within seconds of first spawn)
- Fallback probes added as safety net; tested via code review and ordering analysis

## Rollback Plan

Revert this commit. The previous `--dangerously-skip-permissions` + timer-based detection is in git history and restores cleanly. Note: reverting re-introduces the crash loop on CC 2.1+.

### DIFF
--- a/src/pty/agent-pty.ts
+++ b/src/pty/agent-pty.ts
@@ -153,9 +153,40 @@ export class AgentPTY {
 
     this._alive = true;
 
-    // Set up output capture
+    // Track which interactive prompts we've already responded to.
+    let bypassAccepted = false;
+    let trustAccepted = false;
+
+    // Set up output capture + inline prompt detection.
+    //
+    // IMPORTANT: Raw PTY data contains ANSI escape codes between words, so multi-word
+    // substring matching ("Bypass Permissions", "No, exit") fails silently.
+    // Use single-word substrings that are never split by ANSI codes.
+    //
+    // Prompts handled:
+    //   1. Bypass Permissions warning (CC 2.1+) — identified by raw "Bypass" in data.
+    //      Default selection is "No, exit". Send down-arrow (move to "Yes, I accept")
+    //      then Enter. bypassAccepted is set SYNCHRONOUSLY before any setTimeout so
+    //      the trust handler below cannot fire on the same data chunk.
+    //   2. "trust this folder?" prompt — identified by "trust" (only fires if no bypass
+    //      prompt was detected in this data chunk). Enter accepts (default is Yes).
     this.pty.onData((data: string) => {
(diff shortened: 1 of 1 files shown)


chrisjackson2030-byte:
Closing — will revisit upstream contribution timing later. Fix remains applied locally.

PR #240, opened 2026-04-24
feat(telegram): ack inbound user messages with 👍 + typing indicator

michaelzizz:
## Summary
- Daemon reacts with 👍 and fires `sendChatAction('typing')` the moment an inbound user Telegram message passes the ALLOWED_USER gate — before dedup, media download, or PTY injection, so users see an ack within ~1s regardless of downstream latency.
- New `TelegramAPI.setMessageReaction` helper (empty emoji clears); new `src/telegram/receipt-ack.ts` holds the fire-and-forget logic plus a per-bot `Set<number>` of already-ack'd message_ids, bounded at 1000 entries, so Telegram redeliveries after a handler exception don't double-react.
- Wired into `AgentManager.onMessage` after the ALLOWED_USER gate so bot messages and unauthorized senders are skipped. Both API calls are fire-and-forget — failures are logged but never block message processing.

## Why
When an agent is busy (processing media, generating a response, or just slow on a cold cache), users today get no signal that their message arrived at all. A 👍 + typing indicator within ~1s is a cheap, ubiquitous UX fix.

## Test plan
- [x] `npm run build` — clean
- [x] `npx tsc --noEmit` — clean
- [x] `npm test` — 691/691 passing, including 3 new unit tests covering default emoji payload shape, empty-emoji clears reaction, and API error propagation
- [ ] Manual: send a message to a running agent's bot; 👍 + typing indicator should appear within ~1s
- [ ] Manual: verify re-sending the same `message_id` (simulated redelivery) does not trigger a second reaction API call
- [ ] Reactions may need to be enabled per-bot via @BotFather before they appear in clients — flagging for ops

### DIFF
--- a/src/daemon/agent-manager.ts
+++ b/src/daemon/agent-manager.ts
@@ -12,6 +12,7 @@ import { logInboundMessage, cacheLastSent, logOutboundMessage, buildRecentHistor
 import { collectTelegramCommands, registerTelegramCommands } from '../bus/metrics.js';
 import { stripControlChars } from '../utils/validate.js';
 import { processMediaMessage } from '../telegram/media.js';
+import { ackTelegramReceipt } from '../telegram/receipt-ack.js';
 
 type LogFn = (msg: string) => void;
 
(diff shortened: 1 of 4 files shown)


michaelzizz:
Wrong target — routing to fork's main per orchestrator. Will reopen against michaelzizz/cortextos:main.

PR #242, opened 2026-04-24
fix(bus): reduce observability noise across heartbeats, events, and experiments

ErosContrerasss:
## Summary
Three framework fixes that all surfaced from the same fleet telemetry dump (`orgs/eros-os/agents/finance/memory/2026-04-24_heartbeat_divergence.csv`): activity was happening while agents looked stale to the monitor, and experiments were losing their measurement description on every iteration.

## 1. `logEvent` refreshes the agent's heartbeat timestamp
**Problem:** 76.4% of fleet activity events landed while the heartbeat was >5min stale. Activity is liveness — if an agent is writing events, the stale-heartbeat monitor should not page on it. We were generating noise instead of measuring real downtime.

**Fix:** New helper `refreshHeartbeatTimestamp` in `src/bus/event.ts` bumps `last_heartbeat` on the existing `heartbeat.json` and preserves every other field (status, current_task, mode, loop_interval). No-op when no heartbeat file exists yet; fully best-effort so a failing heartbeat write never blocks the event write itself. Zero changes required on the agent side.

## 2. `completeTask` emits a `task/task_completed` activity event
**Problem:** Previous pattern required agents to follow every `bus complete-task` with a separate `bus log-event task task_completed`. Half the time the log-event call was forgotten and the dashboard feed missed the completion.

**Fix:** The event is now emitted from inside `completeTask` (src/bus/task.ts), so any caller (CLI today, programmatic tomorrow) gets the behavior. Attribution uses the task's `assigned_to` + `org`.

## 3. `createExperiment` inherits from the matching cycle
**Problem:** When the autoresearch skill ran repeat experiments on the same metric, `measurement` kept coming through blank because the agent forgot `--measurement` on the re-run. The cycle config in `experiments/config.json` already holds the canonical measurement method.

**Fix:** New helper `findCycleDefaults` in `src/bus/experiment.ts` looks up the cycle for the same metric + agent and fills in `measurement`, `direction`, `window`, and `surface` when they aren't passed explicitly. Explicit options still win so ad-hoc overrides continue to work.

## Test plan
- [x] `npm test` — 678 tests pass across 47 files
- [x] `npm run build` clean
- [x] 3 new test suites cover each fix:
  - `tests/unit/bus/event.test.ts` — heartbeat refresh side-effect + corrupt-heartbeat resilience + no-op when no heartbeat file
  - `tests/unit/bus/task.test.ts` — task_completed event emission with correct agent/org/metadata
  - `tests/sprint3-experiments.test.ts` — cycle inheritance + explicit-option override + fallthrough to static defaults

### DIFF
--- a/src/bus/event.ts
+++ b/src/bus/event.ts
@@ -1,7 +1,7 @@
-import { appendFileSync } from 'fs';
+import { appendFileSync, existsSync, readFileSync } from 'fs';
 import { join } from 'path';
-import type { EventCategory, EventSeverity, BusPaths } from '../types/index.js';
-import { ensureDir } from '../utils/atomic.js';
+import type { EventCategory, EventSeverity, BusPaths, Heartbeat } from '../types/index.js';
+import { atomicWriteSync, ensureDir } from '../utils/atomic.js';
 import { randomString } from '../utils/random.js';
 import { validateEventCategory, validateEventSeverity, isValidJson } from '../utils/validate.js';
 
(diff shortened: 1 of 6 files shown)


grandamenium:

Cherry-pick of this branch onto current soak (`nightly-soak-2026-05-07` HEAD: `85ffb7b5`) hits conflicts. Files like `src/bus/experiment.ts`, `src/bus/task.ts`, `tests/sprint3-experiments.test.ts`, and `tests/unit/bus/event.test.ts` have moved on `main` since this branch was opened (2026-04-24).

Note: this PR overlaps with the observability-noise area touched by other in-flight work — when rebasing, please re-confirm the scope matches what's now in `main` (some heartbeat/event-noise reductions may already have been addressed by other merged PRs since 2026-04-24).

Squash gate held until rebase lands.

PR #244, opened 2026-04-25
fix(watchdog): complete resetWatchdogState — guard, circuit-breaker fields, and field test

noogalabs:
## Summary

Completes the watchdog state reset introduced in the prior PR. Three issues found via nightly code review:

- **Non-Telegram agents never get watchdog reset**: `onStatusChanged` was registered inside `if (telegramApi && chatId)`, so agents without Telegram credentials silently skipped the reset on every crash recovery. Moved the callback registration unconditionally outside the guard; Telegram notifications remain nested inside the credential check.
- **Circuit-breaker triple not cleared**: `resetWatchdogState()` left `watchdogCircuitBroken`, `watchdogRestarts`, and `watchdogCircuitBrokenAt` dirty. A recovered agent with a previously-tripped circuit breaker would remain blind to stalls for up to 30 minutes post-recovery. All three fields added to the reset.
- **Reset fired on every `running` emission**: The transition guard had no `prevStatus !== 'running'` check, so any re-broadcast of an already-running status would zero a queued context handoff deadline mid-session.

Also adds `@internal` JSDoc to `resetWatchdogState()`, improves `execFile` timeout error logging (distinguishes timeout from other failures), and fixes a stale `3-digit` → `6-digit` JSDoc comment in `task.ts`.

## Test plan

- [ ] `npm run build` — passes clean
- [ ] `npm test` — 47 files / 709 tests pass
- [ ] New `resetWatchdogState — field coverage` test asserts all 12 watchdog fields including circuit-breaker triple are zeroed
- [ ] Non-Telegram agent logs `Watchdog state reset (agent transitioned to running)` after crash recovery

### DIFF
--- a/src/bus/task.ts
+++ b/src/bus/task.ts
@@ -203,7 +203,7 @@ export function checkTaskDependencies(
  *      matching task file. Only runs when the fast path missed, so
  *      same-org operations take no perf hit.
  *
- * Task IDs are generated as `task_<epoch_ms>_<3digit_random>` so real
+ * Task IDs are generated as `task_<epoch_ms>_<6digit_random>` so real
  * collisions are effectively impossible — but if the scan ever finds the
  * same ID in multiple orgs (e.g. due to a bug in ID generation or a manual
  * file copy), we warn loudly naming the task ID, the match count, AND the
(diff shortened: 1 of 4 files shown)


noogalabs:
Build verified locally — 828 tests green, 56 test files pass. Ready for upstream merge when you get a chance, James. No urgency.

PR #248, opened 2026-04-25
fix(cli): close silent-Telegram-disable trap (task_637 misdiagnosed)

loganbronstein:
## Summary

Logan filed `task_1776901473789_637` titled **"Fast-checker not spawned for new agent"** after marketing got 5 Telegram messages between 18:23-21:00 UTC, none reached the session.

**The bug title is misdiagnosed.** Fast-checker DOES spawn. The actual silent failure is the **TelegramPoller**. Daemon log proves it:

```
[ipc] start-agent marketing from cortextos start
[marketing] SECURITY: BOT_TOKEN is set but ALLOWED_USER is missing.
            Refusing to enable Telegram. ...
[marketing] Bootstrap complete. Beginning poll loop.   ← fast-checker IS running
```

`agent-manager.ts` only constructs the `TelegramPoller` `if (telegramApi && chatId)`. The security gate (lines 224-231) nulls `telegramApi` when `ALLOWED_USER` is missing, leaving the poller unconstructed. Fast-checker polls the agent's local inbox just fine — but nothing pulls messages from the Telegram API into that inbox. From the user's terminal `cortextos start <agent>` returns success and the agent appears running. The warning lands in `~/.pm2/logs/cortextos-daemon-out.log` where users never look.

Why hittable: `cortextos add-agent` writes a `.env` template containing only `BOT_TOKEN=` and `CHAT_ID=` — no `ALLOWED_USER=` field, no comment explaining it's required. Same gap in `cortextos setup`'s `writeAgentEnv()`. The path of least resistance produces a Telegram-deaf agent.

### 1. `src/cli/add-agent.ts`
Extend the `.env` template to include `ALLOWED_USER=` placeholder + a comment block explaining what it is, why the daemon requires it (security: anyone who finds the bot @handle could otherwise control the agent), and how to find your numeric Telegram user ID (`@userinfobot` or the `from.id` field in `getUpdates`).

### 2. `src/cli/setup.ts`
`writeAgentEnv()` now accepts an optional `allowedUser` param. The wizard prompts for it after the chat ID, defaulting to chat ID (correct for personal bots; the numeric user ID and personal chat ID are usually the same).

### 3. `src/cli/start.ts`
[...]

PR #245, opened 2026-04-25
feat(hooks): post-merge PR notification (Tier 1 Telegram + Tier 2 digest)

loganbronstein:
## Summary

- Closes the "merged but Logan never knew" gap. New `scripts/post-merge-notify.sh` classifies the just-merged PR against Logan's notify rules (set 2026-04-23) and either Tier 1 pings him immediately or Tier 2 queues into a daily digest.
- **Tier 1** (score >= 3): new skill / agent / bus command / integration / dashboard surface / material quality upgrade. Sends `"New thing you can use: <title>. <coaching>. PR: <url>"` via `cortextos bus send-telegram`.
- **Tier 2** (score 1-2): borderline. Appends a JSON line to `~/.cortextos/$CTX_INSTANCE_ID/state/pr-digest.jsonl`.
- **Skip** (score <= 0): bug fixes, docs, infra tweaks, refactors, tests.
- Default-to-quiet on ambiguity (matches Logan's no-noise preference).
- `scripts/post-merge-digest-send.sh` (also new) reads the accumulator, sends one consolidated message, atomically clears the file. Designed for daily cron at 22:00 America/Chicago. Cron wiring tracked as follow-up — runnable script ships now.
- `scripts/hooks/post-merge` (modified) adds a second background fork alongside the existing supabase-deploy fork. Non-blocking; every failure path inside the notifier exits 0 so a slow Telegram or missing `cortextos` binary cannot wedge a `git pull`.

## Behavior knobs (env vars)

| Var | Default | Effect |
|---|---|---|
| `POST_MERGE_NOTIFY_SKIP=1` | unset | Kill switch — exit 0 immediately |
| `POST_MERGE_NOTIFY_DRY_RUN=1` | unset | Print verdict + would-send, skip Telegram + digest write |
| `POST_MERGE_NOTIFY_BOSS_ENV=<path>` | `orgs/cortex/agents/boss/.env` | Source of `BOT_TOKEN` + `CHAT_ID` |
| `POST_MERGE_NOTIFY_DIGEST_FILE=<p>` | `$CTX_ROOT/state/pr-digest.jsonl` | Accumulator path |
| `POST_MERGE_NOTIFY_BASE_BRANCH=<b>` | `main` | Hook is no-op on other branches |
| `POST_MERGE_NOTIFY_REPO_URL=<url>` | derived from `git remote get-url origin` | PR link prefix |

## Test plan

[...]

PR #243, opened 2026-04-25
feat(hooks): pre-push Plan-First Review gate (W9b follow-up)

loganbronstein:
## Summary

- Closes the W9b follow-up from PR #216: Plan-First Review rule was enforced by convention only. This PR adds the tool-level `pre-push` gate so the 10-persona panel can't be skipped by accident.
- New `scripts/plan-review-gate.sh` (162 lines, macOS bash 3.2 compatible) invoked by `scripts/hooks/pre-push` BEFORE build/tests, so a missing plan fails fast instead of after a 2-min test run.
- Two acceptable plan shapes: full panel with `QUORUM: PASS (>= 8/10)` and mandatory Security + DataIntegrity PASS, OR a single-line `TRIVIAL-TASK-EXEMPT: <reason>` for tiny work. `PLAN_REVIEW_BYPASS=true` is the documented escape, telemetry-logged via `cortextos bus log-event`.
- `.gitignore` flipped from blanket `.claude/` ignore to `.claude/*` + explicit re-includes so plan files in `.claude/plan-reviews/` are tracked across clones and visible in PR diffs. Existing re-includes (commands, orchestration) preserved.
- Carries its own plan-review file at `.claude/plan-reviews/feat-pre-pr-plan-review-gate-plan.md` (10/10 panel) so the gate passes against itself.

## Test plan

- [x] Missing plan file → BLOCK with coaching message (verified)
- [x] `TRIVIAL-TASK-EXEMPT: <reason>` → PASS (verified)
- [x] `TRIVIAL-TASK-EXEMPT:` with empty reason → BLOCK (verified)
- [x] Full panel `QUORUM: PASS (8/10)` with Security + DataIntegrity PASS → PASS (verified)
- [x] Full panel `QUORUM: PASS (7/10)` → BLOCK (verified)
- [x] Security row marked FAIL → BLOCK even at 9/10 (verified)
- [x] Security persona missing entirely → BLOCK (verified)
- [x] `PLAN_REVIEW_BYPASS=true` → PASS, telemetry logged (verified)
- [x] Protected branch (main) → skipped (verified)
- [x] `npm run build` clean (verified)
- [x] Real `git push` exercises the hook against this branch's own plan file (verified — this PR was pushed through the gate)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #249, opened 2026-04-25
fix(hooks): plan-review-gate word-boundary verdict match (closes own #415, extends #243)

loganbronstein:
## Summary

Closes `task_1777093291910_415`, which I filed against the gate I shipped in PR #243 while writing the plan-review file for PR #248. The persona-row verdict check used a substring grep:

```sh
if echo "$PERSONA_LINE" | grep -qiE "FAIL|REJECT|BLOCK"; then
```

This false-positived on prose words like **"failure", "blocking", "non-blocking", "unblocking", "rejection"** that authors legitimately use in plan-review notes. I hit it twice this session — PR #245 plan ("hook failure cannot corrupt a merge") and PR #248 plan ("non-blocking so a slow Telegram cannot wedge"). Both forced me to rewrite legitimate prose just to satisfy the regex.

## Relationship to PR #243

This branch is built on top of `feat/pre-pr-plan-review-gate` (PR #243, still open) so it can compile against the gate script that PR introduces. The diff in this PR therefore contains BOTH:

1. The PR #243 contents (original gate + integration into pre-push hook + .gitignore flip + README + TEMPLATE + plan-review file).
2. **The new fix on top**: `scripts/plan-review-gate.sh` switches from substring to word-boundary matching, plus 13 integration tests, plus this PR's own plan-review file.

Reviewer guidance: if PR #243 is reviewed/merged first, this PR's diff against main collapses to just commit `96550a7` (the actual fix). I considered cherry-picking just the fix into a separate branch off main but the gate doesn't exist on main yet, so the test file would have nothing to test.

## The fix

Switch from substring `grep -iE` to word-boundary `grep -wiE` against an explicit list of verdict words + their common inflections:

```sh
PERSONA_VERDICT_PATTERN='FAIL|FAILED|FAILS|REJECT|REJECTED|REJECTS|BLOCK|BLOCKED|BLOCKS'
if echo "$PERSONA_LINE" | grep -wqiE "$PERSONA_VERDICT_PATTERN"; then
```

`-w` requires the match to be preceded and followed by non-word characters (alphanumeric + underscore boundary). Result:

| Input | Old behavior | New behavior |
|---|---|---|
| `Security \| FAIL \|` | BLOCK ✓ | BLOCK ✓ |
| `Security \| BLOCKED \|` | BLOCK ✓ | BLOCK ✓ (in expanded list) |
| `Security \| FAILS \|` | BLOCK ✓ | BLOCK ✓ (in expanded list) |
| `Security \| PASS \| no failure mode` | **BLOCK (false positive)** | PASS ✓ |
| `Security \| PASS \| non-blocking hook` | **BLOCK (false positive)** | PASS ✓ |
| `Security \| PASS \| adds rejection rules` | **BLOCK (false positive)** | PASS ✓ |

## Test plan

- [x] 13 new integration tests in `tests/unit/scripts/plan-review-gate.test.ts` via real `bash` invocation against a scratch git repo + plans dir. Mirrors the existing `auto-deploy-supabase.test.ts` pattern.
  - True-positives that must still BLOCK: FAIL, REJECTED, BLOCK, BLOCKED, FAILS, lowercase `fail`
  - False-positives that must now PASS: "failure", "non-blocking", "blocking", "unblocking", "rejection" prose
  - Compositions: trivial-exempt path still passes, missing-QUORUM-line still blocks
- [x] Self-validation: this PR's own plan-review file deliberately includes "rejection", "non-blocking", and "BLOCK" prose in the mandatory persona rows. With the fixed pattern the gate passes against it — proof the fix works against the exact phrasings that bit me.
- [x] Full suite: 769/770 (the 1 fail is the pre-existing dashboard channels flake)
- [x] `npm run build` clean

## Adversarial pass

- Considered adding WARN as a third non-PASS verdict — out of scope. If reviewers want a soft warning, they can use prose in the row (which now actually works post-fix).
- Considered migrating the gate from bash to node — bash 3.2 is the constraint and `-w` handles macOS just fine. Migration would unlock proper word-boundary regex (`\b`) but the cost/value isn't there.
- Considered the reviewer-ambiguity edge case — what if someone writes "Security | FAIL — wait actually PASS"? Gate still blocks (FAIL is its own word). That's correct: the reviewer should pick one verdict.

## Plan-review artifact

`.claude/plan-reviews/fix-plan-review-gate-word-boundary-plan.md` (10/10 panel). Note that this very file would have tripped the OLD pattern — its inclusion in this PR is itself the proof that the fix works.


grandamenium:
Closing per nightly-soak walk-through review (2026-05-07).

This PR fixes a verdict-regex bug in PR #243's pre-push gate. PR #243 closed today as too heavyweight a default for cortextOS, so #249's paired fix is moot.

Thanks.

PR #255, opened 2026-04-25
fix(dashboard): atomic CSRF refetch at submit time to defeat StrictMode mount-race

dexdan193-create:
## Summary

The login page intermittently fails sign-in with `MissingCSRF` in dev mode. Root cause: React StrictMode double-invokes the mount-time `useEffect` that fetches `/api/auth/csrf`, and when the two in-flight responses resolve out of order, the browser's `authjs.csrf-token` cookie ends up pinned to one token while `csrfTokenRef.current` holds the other. NextAuth's CSRF validator compares the body token against the cookie token and rejects the mismatch.

The fix: re-fetch the CSRF token *atomically at submit time* with `cache: 'no-store'`, immediately before constructing the request body. The cookie that the server reads on POST is necessarily the one set by this most-recent `/api/auth/csrf` response, so cookie and body are forced into sync regardless of any earlier mount-time race.

## Reproduction

1. `cd dashboard && npm run dev` (StrictMode is enabled in dev)
2. Open `/login` in a fresh incognito window
3. Submit valid credentials
4. ~30-50% of attempts return `MissingCSRF` redirect; the rest succeed.
5. Reload + retry shows the same intermittent pattern.

In production builds (`npm run build && npm start`) StrictMode is off so the double-mount does not occur and the bug is masked.

## Diagnosis

Sequence of events in the failing case:

1. `useEffect` mount fires. StrictMode immediately fires it a second time.
2. Both invocations call `fetch('/api/auth/csrf')`. Each request causes the server to set the `authjs.csrf-token` cookie.
3. The two responses can resolve in either order. Whichever resolves *first* writes its token to `csrfTokenRef.current`. Whichever resolves *second* writes its token to the cookie (because the cookie is set by the response, not by the assignment in JS).
4. If the second-to-resolve response carries a different token, then `csrfTokenRef.current` and the cookie are out of sync at submit time.
5. The form POSTs with the body token from the ref + the cookie token from the second response → mismatch → `MissingCSRF`.

## Fix

`handleSubmit` now performs an additional `/api/auth/csrf` fetch immediately before constructing the urlencoded body, with `cache: 'no-store'` to force a network round-trip. The response sets the cookie atomically and provides the matching token for the body. Cookie and body are guaranteed in sync at the moment of POST.

Falls back to the mount-time `csrfTokenRef.current` if the refetch throws (network blip), so the UX degrades to the existing race condition rather than to a hard failure.

## Test plan

- [x] Local dev mode: 20 sign-in attempts, 20 successes, 0 `MissingCSRF` (was ~30-50% failure rate pre-patch)
- [x] Local prod build: 5 sign-in attempts, 5 successes (no regression — was already passing)
- [ ] Upstream CI: hopefully has Playwright e2e against `/login` that catches the regression directly.

## Why not disable StrictMode

StrictMode catches a real class of bugs (effects that cannot tolerate double-mount). The login page has the bug, not StrictMode. Disabling it would mask the bug locally but leave it latent for any production deploy that triggers a comparable race.

## Notes

- One file touched: `dashboard/src/app/(auth)/login/page.tsx`
- No new dependencies, no type changes
- The fallback in the `catch {}` is intentional — degrading to the pre-patch behaviour rather than blocking the user is the right call when the network blips
- Running this patch locally since 2026-04-22 with zero observed `MissingCSRF` regressions

### DIFF
--- a/dashboard/src/app/(auth)/login/page.tsx
+++ b/dashboard/src/app/(auth)/login/page.tsx
@@ -80,8 +80,24 @@ export default function LoginPage() {
     const usernameInput = form.querySelector('input[name="username"]') as HTMLInputElement | null;
     const passwordInput = form.querySelector('input[name="password"]') as HTMLInputElement | null;
 
+    // Re-fetch CSRF token at submit time so body token matches the current
+    // cookie. React StrictMode double-invokes the mount-time CSRF useEffect
+    // in dev; if the two in-flight /api/auth/csrf responses resolve out of
+    // order, the browser's authjs.csrf-token cookie can end up pinned to a
+    // different token than csrfTokenRef.current — which the server rejects
+    // as MissingCSRF. An atomic fetch-then-submit here forces cookie and
+    // body into sync regardless of earlier mount-time races.
+    let submitToken = csrfTokenRef.current;
+    try {
+      const freshCsrf = await fetch('/api/auth/csrf', { credentials: 'same-origin', cache: 'no-store' });
+      const freshData = await freshCsrf.json();
+      if (freshData?.csrfToken) submitToken = freshData.csrfToken;
+    } catch {
+      // Fall back to the mount-time token if the refetch fails.
+    }
+
     const body = new URLSearchParams();
-    body.set('csrfToken', csrfTokenRef.current || '');
+    body.set('csrfToken', submitToken || '');
     body.set('username', usernameInput?.value || '');
     body.set('password', passwordInput?.value || '');
 


grandamenium:

Cherry-pick of this branch onto current soak (`nightly-soak-2026-05-07` HEAD: `85ffb7b5`) hits a conflict on `dashboard/src/app/(auth)/login/page.tsx`, which has moved on `main` since this branch was opened (2026-04-25).

The fix (atomic CSRF refetch at submit time to defeat StrictMode mount-race, +17/-1, single file) is well-scoped with concrete dev-mode evidence (~30-50% intermittent failure pre-patch → 0/20 post-patch). Once rebased, this should land quickly.

Squash gate held until rebase lands.

PR #263, opened 2026-04-27
fix(bus): populate measurement_start/end on runExperiment + soft-warn evaluate

ErosContrerasss:
## Summary

- runExperiment now populates `measurement_start` (= started_at) and `measurement_end` (= start + parsed window). Both were previously left null after run-experiment, forcing downstream tooling and operators to use `started_at` as a proxy and guess the window. Rubi has flagged this 3 times.
- Window parser handles `s` / `m` / `h` / `d` / `w` units plus full-word aliases (`hours`, `days`, etc.). Returns null on anything unparseable so we don't fabricate an end-time.
- `evaluateExperiment` adds a soft warning (no throw) when called before `measurement_end` — agents may want to short-circuit on dramatic results, and we'd rather warn than block. Skips the check when `measurement_end` is missing so experiment files written before this commit stay evaluatable.

## Test plan

- [x] `npm test` — 719 tests pass across 47 files (35 of those in `tests/sprint3-experiments.test.ts`, including 6 new in a `measurement window` describe block)
- [x] `npm run build` clean
- [x] Tests cover: start = started_at + window offset, every supported unit (s/m/h/d/w + word forms), unparseable-window warning + null end, evaluate soft-warn before end, no warn after end, no warn on legacy rows missing the field


grandamenium:
Sandbox audit (cortext-designer) at head 2beb2af PASSed 9/10 against pre-soak base. When attempting to squash into nightly-soak-2026-05-07 tonight, hit a merge conflict in tests/sprint3-experiments.test.ts against the current soak base. Could you rebase against grandamenium/main and push? I will re-test once you ping. (Not a defect in the PR — just a base-divergence cleanup.)

PR #264, opened 2026-04-27
fix(enable-agent): add missing writeFileSync import (BUG-036 regression)

revopsglobal:
## Summary

- `writeDisableMarker` in `src/cli/enable-agent.ts` was importing `{ existsSync, readFileSync, mkdirSync }` from `'fs'` — `writeFileSync` was missing
- The `writeFileSync(...)` call inside the function threw a `ReferenceError` that the surrounding `try/catch` silently swallowed
- Result: `.user-disable` marker was never written, so `hook-crash-alert` classified every intentional `cortextos disable` as a crash (false 🚨 CRASH alarms)
- `writeStopMarker` in `stop.ts` had the correct import — that's why `.user-stop` worked but `.user-disable` didn't

**Fix:** add `writeFileSync` to the `fs` import (one character change).

**Root cause analysis:** Identified by investigating 2 pre-existing failures in `tests/unit/cli/lifecycle-markers.test.ts` (task_1777313307656_812). The symptom was: directory was created (test 2 passed) but file was never written (test 1 failed) — pointing directly to the `writeFileSync` call failing silently inside catch.

## Test plan

- [x] `tests/unit/cli/lifecycle-markers.test.ts` — was 2 failing, now 7/7 pass
- [x] Full suite — 1101/1101 pass, no regressions


revopsglobal:
Wrong target — opened against upstream by mistake.

PR #265, opened 2026-04-28
fix(daemon): teach gap detector how often DoW/DoM/Month-restricted crons fire

ErosContrerasss:
## Summary

Crons with day-of-week, day-of-month, or month restrictions were all classified as 24h by \`cronExpressionMinIntervalMs()\`, so the gap detector fired false-positive nudges every 48h on:

- Sunday-only — analyst's \`catalog-browse: 4 23 * * 0\`
- Friday-only — finance's \`friday_checkin: 7 17 * * 5\`
- Weekday range — \`0 9 * * 1-5\` (Fri→Mon = 72h gap)
- Monthly DoM — finance's \`monthly_invoice_prep: 17 9 28 * *\`
- Quarterly — finance's \`btw_deadline_tracker: 23 9 20 1,4,7,10 *\`
- Yearly — finance's \`year_end_prep: 43 10 15 11 *\`

Frank had 7 of these patterns, all surfacing the same nudge spam in TW#10.

## Fix

Rewrites the function to estimate the **max gap between consecutive fires** (not the typical fire frequency — that's the wrong number for gap-detection, which fires on dead zones rather than rate):

| Pattern | Returned interval |
|---|---|
| DoW restricted (single, range, list, step) | max gap between active weekdays |
| DoM every-N (\`*/N\`) | N days |
| DoM fixed (single day) | 31d (worst-case month length) |
| Month + DoM | 365d (yearly cadence) |
| Month-only restriction | 365d (conservative) |
| Lists / mixed DoM/DoW/Month combos | 31d fallback |

Over-estimation is the safe direction: it can only delay a legitimate nudge, never trigger a false-positive on a working cron — which was the problem reported.

The existing precedence in \`agent-process.ts\` already prefers an explicit \`interval\` field on a cron entry when present, so authors of unusual cron expressions can still override the derived value by setting both \`cron\` and \`interval\` on the config entry.

## Test plan

- [x] \`npm test\` — 723 tests pass across 47 files (16 new in a \`cronExpressionMinIntervalMs\` describe block)
- [x] \`npm run build\` clean
- [x] Tests cover every pattern from the affected production configs (analyst + finance + max), plus malformed inputs

### DIFF
--- a/src/bus/cron-state.ts
+++ b/src/bus/cron-state.ts
@@ -95,26 +95,129 @@ export function parseDurationMs(interval: string): number {
 }
 
 /**
- * Estimate the minimum expected firing interval for a 5-field cron expression.
- * Handles common patterns (every-N-minutes, every-N-hours, daily) without an
- * external library. Returns a conservative 48h fallback for anything else.
+ * Estimate the maximum expected gap between consecutive fires for a 5-field
+ * cron expression. The gap-detector multiplies this by 2 to derive its nudge
+ * threshold, so the value must reflect the worst-case real-world gap — not
+ * the typical case — or restricted crons get false-positive nudges.
+ *
+ * Coverage (using "[/]" to denote the cron step operator without confusing
+ * this JSDoc block):
+ *   - every-N-minutes:           "[/]N * * * *"          → N min
+ *   - every-N-hours:             "<m> [/]N * * *"        → N h
+ *   - daily fixed-hour:          "<m> <h> * * *"         → 24 h
+ *   - day-of-week restricted:    "<m> <h> * * <dow>"     → max gap between
+ *                                                          active weekdays
+ *                                                          (1-5 → 72 h,
+ *                                                          Sun-only → 168 h)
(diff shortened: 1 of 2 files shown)


grandamenium:

grandamenium:
Closing per nightly-soak walk-through review (2026-05-07).

Verdict: the function this PR rewrites — `cronExpressionMinIntervalMs` — no longer has any runtime caller in the codebase. The gap-detection loop in `agent-process.ts` referenced in the PR description has been removed as part of the Phase 5 external-cron migration. Current consumers (`cron-health.ts`, `cron-scheduler.ts`, `ipc-server.ts`, `bus.ts`) all use `parseDurationMs` only; cron expressions are treated as interval=0/unknown by design.

The fix is technically correct but applies to dead code. Merging would add 213 lines (impl + tests) of unreachable logic.

Thanks for the careful work and the well-documented bug surface — if the gap-detection loop is ever reintroduced, this PR is a useful reference for the right interval semantics.

Following up: open issue to delete `cronExpressionMinIntervalMs` entirely if confirmed unused.

PR #267, opened 2026-04-28
fix(daemon): emit telegram_received bus event on inbound messages

ErosContrerasss:
## Summary

Rubi's v3 experiment window (Apr 25-28) found **0 \`telegram_received\` events fleet-wide** despite Eros replying to multiple agents in that period. The JSONL inbound archive was being written, but the daemon never emitted a bus event — so dashboards and experiment cycles that count inbound traffic saw zero on healthy agents.

The outbound side already emits \`message/telegram_sent\` from \`cortextos bus send-telegram\` (\`src/cli/bus.ts:1007\`). The inbound side had nothing equivalent.

## Fix

Adds \`recordInboundTelegram\` to \`src/telegram/logging.ts\` that does both side effects in one call:

1. **logInboundMessage** to \`{ctxRoot}/logs/{agent}/inbound-messages.jsonl\` — same shape as before, untouched
2. **logEvent('message', 'telegram_received', 'info')** with metadata:
   - \`chat_id\`, \`message_id\`, \`from_id\`, \`from_name\`
   - \`has_media\` — boolean for photo/document/voice/audio/video/video_note
   - \`text_chars\` — text or caption length (the value experiment cycles want for engagement scoring)

The agent-manager poller handler in \`src/daemon/agent-manager.ts\` is updated to call the new helper instead of inlining the JSONL write.

The \`logEvent\` call is wrapped: a \`logEvent\` failure (e.g. unwritable analytics dir) cannot break message processing — the JSONL still goes through and the failure is surfaced through the daemon log callback.

## Test plan

- [x] \`npm test\` — 711 tests pass across 47 files
- [x] \`npm run build\` clean
- [x] 3 new tests in \`tests/unit/telegram/logging.test.ts\`:
  - happy path: text message produces both JSONL row + bus event with the right shape
  - photo path: \`has_media=true\` and \`text_chars\` reflects the caption length
  - resilience: when the bus-event emit throws, the JSONL row is still written and the failure is surfaced through the log callback

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #269, opened 2026-04-28
fix(daemon): emit telegram_disabled warning event when daemon refuses Telegram at startup

ErosContrerasss:
## Summary

Lucy/health was silently inbound-disabled for days because the daemon's fail-closed security check refused Telegram (ALLOWED_USER missing from `.env`) but only logged the refusal to stdout. The operator never saw it on the dashboard, the agent's heartbeat looked healthy, and outbound sends still worked — so the symptom was "agent silent on Telegram with no visible error." Identical pattern flagged twice (Apr 26 + Apr 28).

Confirmed via `pm2 logs cortextos-daemon`:
- `[health] SECURITY: BOT_TOKEN is set but ALLOWED_USER is missing.` ← present every restart
- `[health] Telegram poller started` ← absent (every other agent has this line)

## Fix

Adds an `emitTelegramDisabled` helper in `src/daemon/agent-manager.ts` and wires it into every refusal site in `startAgent`:

| Reason | Trigger |
|---|---|
| `bad_token_format` | `BOT_TOKEN` doesn't match `\d+:[A-Za-z0-9_-]+` |
| `bad_allowed_user_format` | `ALLOWED_USER` not numeric |
| `missing_allowed_user` | `BOT_TOKEN` set, `ALLOWED_USER` missing (Lucy's case) |
| `missing_chat_id` | `BOT_TOKEN` set, `CHAT_ID` missing (was completely silent before) |

Each refusal now emits `error/telegram_disabled` (severity `warning`) with `metadata.reason`. Dashboard can group by reason; Rubi's recurrence detector picks up persistent config issues across daemon restarts.

**Severity choice — `warning`, not `error`:** PR#266's metrics-collector filter excludes `severity=warning` from `errors_today`, so the errors gauge stays clean (these are config issues, not failures). The `category=error` keeps "show me everything broken" filters working.

The emitter is wrapped: if `logEvent` itself throws (e.g. unwritable analytics dir), we surface through the log callback rather than aborting the agent startup path that called us.

## Test plan

- [x] `npm test` — 714 tests pass across 48 files (6 new in `tests/unit/daemon/telegram-disabled-event.test.ts`)
- [x] `npm run build` clean
- [x] Tests cover every reason via `it.each`, plus the resilience case (logEvent throw → log callback fires, no rethrow), plus the symbiosis with PR#266's severity filter


grandamenium:
Sandbox audit (cortext-designer) at head 4518764 PASSed 9/10 against pre-soak base. When attempting to squash into nightly-soak-2026-05-07 tonight, hit a merge conflict in src/daemon/agent-manager.ts after #267 (telegram_received event) was squashed in — both PRs touch the agent-manager event-emission path. Could you rebase against grandamenium/main (which will pick up #267) and push? I will re-test once you ping. (Not a defect in the PR — logical overlap with a sibling PR that landed first.)

PR #299, opened 2026-05-03
fix(daemon): improve gap nudge with CronCreate hint — closes /loop AskUserQuestion silent fail

mccucobi-source:
## Summary

- Daemon's gap-detection nudge now embeds a clean cron expression + the verbatim prompt from `config.json`, so the agent can recreate a missing cron via `CronCreate` directly.
- Closes the silent failure mode where intervals ≥60 min triggered `/loop`'s `AskUserQuestion` confirmation that the agent can't answer.
- Falls back to the old `/loop` instruction (with a note to use the verbatim prompt) when the interval can't be expressed as a clean cron (e.g. `7m`, `5h`, `2w`).

## Why

CronCreate auto-expires entries after 7 days. When an interval-based cron drops off, the previous nudge said "run /loop 4h <cron prompt>" — but `/loop` for ≥60 min intervals pops `AskUserQuestion`, which the agent has no way to answer in a cron-injected nudge context. Result: nudge fires, agent reads it, can't act, cron stays missing.

This change makes the nudge actionable in all cases where the interval is convertible.

## Implementation

- `intervalToCronExpression(interval)` added to `src/bus/cron-state.ts`. Mirrors `/loop`'s rounding rules: only emits expressions for intervals that cleanly divide their unit (e.g. `5m → */5 * * * *`, `4h → 0 */4 * * *`, `1d → 0 0 */1 * *`, `1w → 0 0 * * 0`); returns `null` otherwise.
- `runGapDetectionLoop` signature widened to include `prompt`. `monitorable` filter now requires a valid `prompt` string so the nudge is always actionable.
- New nudge text: when convertible, instructs `CronCreate` directly with `cron`, `recurring=true`, and a JSON-quoted `prompt`. Otherwise falls back to a `/loop` hint that explicitly tells the agent to copy the verbatim prompt from `config.json`.

## Test plan

- [x] 9 new unit tests for `intervalToCronExpression` (clean intervals, non-divisible cases, all units, malformed input)
- [x] `npm run build` clean
- [x] `npm test` — 748/748 pass (was 739, +9 new)
- [ ] Smoke: simulate stale cron-state.json, observe new nudge text in agent PTY

## Scope note

Pragmatic Option A: improves the existing nudge surface without daemon-side cron self-recreate (CronCreate is a Claude Code session-only tool — the daemon can't call it). Hook 1 (PR #298, agent-crash → bus alert) and Hook 3 (approval → Telegram) are separate PRs.

### DIFF
--- a/src/bus/cron-state.ts
+++ b/src/bus/cron-state.ts
@@ -95,6 +95,41 @@ export function parseDurationMs(interval: string): number {
 }
 
 /**
+ * Convert a duration interval ("4h", "30m", "1d", "1w") into a 5-field cron
+ * expression. Returns null when the interval doesn't cleanly fit a recurring
+ * cron slot (e.g. "7m" leaves uneven gaps; "5w" has no clean weekly form).
+ *
+ * Mirrors the rounding rules in the /loop skill so daemon-injected nudges can
+ * give the agent a CronCreate-ready expression instead of a /loop command
+ * (which prompts AskUserQuestion for ≥60-minute intervals).
+ */
+export function intervalToCronExpression(interval: string): string | null {
+  const match = /^(\d+)(m|h|d|w)$/.exec(interval.trim());
+  if (!match) return null;
+  const n = parseInt(match[1], 10);
+  if (n <= 0) return null;
+  const unit = match[2];
+
+  if (unit === 'm') {
+    if (n <= 59) return 60 % n === 0 ? `*/${n} * * * *` : null;
+    if (n % 60 !== 0) return null;
+    const hours = n / 60;
+    return hours <= 23 && 24 % hours === 0 ? `0 */${hours} * * *` : null;
+  }
+  if (unit === 'h') {
+    if (n <= 23) return 24 % n === 0 ? `0 */${n} * * *` : null;
+    return null;
+  }
+  if (unit === 'd') {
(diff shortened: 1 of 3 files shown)


grandamenium:
Closing on supersession by v0.2.0 external persistent crons system (PR #308 / merge `9887c382`).

**Rationale:** This PR fixes the gap-nudge text when an agent's session cron silently drops via CronCreate's 7-day auto-expiry, by hinting at `CronCreate` directly instead of `/loop` (which pops AskUserQuestion on ≥60min intervals).

The 7-day auto-expiry is a CronCreate session-only tool artifact. v0.2.0 introduces daemon-owned external persistent crons via `scheduled_tasks.json` + `src/daemon/cron-scheduler.ts` that:
- Survive agent restarts and context handoffs
- Do not auto-expire
- Are registered at agent boot from `config.json` directly, not via CronCreate

Standard recurring work (heartbeats, monitors, polling) is now external-persistent by default. The dominant /loop + CronCreate failure mode this PR addresses is no longer the primary persistence path.

If session-only CronCreate usage is still surfacing this exact pain post-v0.2.0, file a fresh issue with concrete repro and we'll revisit.

Per James 2026-05-05: closing Tier 1 PRs solved by external crons; surviving Tier 1 (#298 agent_crashed alert, #301 approval ping) batch-test via cortext-designer in single sandbox session.

PR #300, opened 2026-05-03
feat(daemon): cron-list mismatch detection — Hook 2, closes 7-day silent cliff

mccucobi-source:
## Summary

- Daemon now detects crons that have silently dropped out of the agent's live `CronList` and injects a forced-recreate nudge **before** the existing gap detector's 2x-interval threshold elapses (which on a weekly cron is 14 days of silent failure).
- New CLI `cortextos bus update-cron-list` lets the agent mirror its `CronList` output to `state/<agent>/cron-list.json` every heartbeat.
- Step 1b added to all four heartbeat templates so new agents adopt the dump-then-detect protocol automatically.

## Why

Closes the 7-day `CronCreate` auto-expiry cliff at the earliest possible signal: the moment a cron disappears from `CronList`, not 2x the interval later. PR #299 closed the related symptom (gap-nudge text recommending `/loop` on long intervals); this PR closes the larger silent-failure class itself.

## How it works

1. **Agent dump** (heartbeat step 1b): agent calls `CronList`, builds a JSON array of `{name, prompt}` matching live entries against `config.json`, pipes via `cortextos bus update-cron-list`. CLI atomic-writes to `state/<agent>/cron-list.json` with `{updated_at, crons[]}`.
2. **Daemon poll** (`scheduleCronListMismatchDetection`): every 10 min, daemon reads `cron-list.json`, computes which `config.json` crons are missing from the live dump (via `findMissingCrons` — matches by name, falls back to prompt equality so renames don't false-positive).
3. **Forced-recreate nudge**: for each missing cron, daemon injects `[SYSTEM] Cron \"X\" is in config.json but missing from your last CronList dump (likely silently dropped via 7-day auto-expiry). Recreate it NOW with CronCreate (do NOT use /loop): cron=\"…\", recurring=true, prompt=…`. Reuses #299's `intervalToCronExpression` for the cron-expression conversion.
4. **Edge handling**: file missing → warn-once + skip; >1h stale → skip + wait next heartbeat; malformed JSON → skip. No false positives if the agent never dumps.

## Implementation

- `src/bus/cron-list.ts`: `writeCronList` / `readCronList` / `findMissingCrons` (pure function so daemon loop and unit tests share impl).
- `src/cli/bus.ts`: `update-cron-list` command (stdin or `--json`).
- `src/daemon/agent-process.ts`: `scheduleCronListMismatchDetection` + `runCronListMismatchLoop`.
- `src/daemon/agent-manager.ts`: kicks off mismatch detection alongside gap detection on agent start.
- `templates/{agent,analyst,orchestrator,hermes}/HEARTBEAT.md`: Step 1b. Live per-agent `HEARTBEAT.md` files are patched separately by the orchestrator agent per propagation policy.

## Dependency

Builds on **PR #299** (improved gap nudge + `intervalToCronExpression`). Needs a clean rebase if #299 lands second; conflicts are trivial.

## Test plan

- [x] 15 new unit tests (`tests/unit/bus/cron-list.test.ts`): round-trip, missing/malformed/stale handling, mismatch detection by name + by prompt, edge cases (disabled / once / no-prompt configs).
- [x] `npm run build` clean.
- [x] `npm test` — 751/751 pass.
- [ ] Smoke: stage a `cron-list.json` missing one config cron, observe forced-recreate nudge in agent PTY.

### DIFF
--- a/src/bus/cron-list.ts
+++ b/src/bus/cron-list.ts
@@ -1,0 +1,106 @@
+/**
+ * Daemon-side mirror of the agent's live CronList output (cron-list.json).
+ *
+ * The agent dumps its current CronList tool result here on every heartbeat
+ * via `cortextos bus update-cron-list` (stdin JSON). The daemon polls this
+ * file and compares against config.json's `crons[]` to detect crons that
+ * have silently dropped out of the active session — typically because of
+ * the 7-day CronCreate auto-expiry, an unobserved restart, or compaction.
+ *
+ * On mismatch, the daemon injects a forced-recreate nudge so the agent
+ * recreates the cron via CronCreate before the gap-detector's 2x-interval
+ * threshold has had time to elapse (which on a weekly cron is 14 days of
+ * silent failure — unacceptable).
+ *
+ * Storage: state/<agent>/cron-list.json (same dir as cron-state.json).
+ */
+
+import { existsSync, readFileSync } from 'fs';
+import { join } from 'path';
+import { atomicWriteSync, ensureDir } from '../utils/atomic.js';
+
+export interface CronListEntry {
+  /** The cron's logical name as it appears in config.json. */
+  name: string;
+  /** The verbatim prompt the cron fires (matched against config.json on agent dump). */
+  prompt: string;
+}
+
+interface CronListFile {
+  updated_at: string;
+  crons: CronListEntry[];
+}
+
+function cronListPath(stateDir: string): string {
+  return join(stateDir, 'cron-list.json');
+}
+
+/**
+ * Atomically write the agent's current CronList contents.
+ * Called by `cortextos bus update-cron-list` (stdin JSON).
+ */
+export function writeCronList(stateDir: string, crons: CronListEntry[]): void {
+  ensureDir(stateDir);
+  const payload: CronListFile = {
+    updated_at: new Date().toISOString(),
+    crons,
+  };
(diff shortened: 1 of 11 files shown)


grandamenium:
Closing on supersession by v0.2.0 external persistent crons system (PR #308 / merge `9887c382`).

**Rationale:** This PR closes the 7-day silent cliff caused by `CronCreate`'s auto-expiry — agent dumps live `CronList`, daemon detects when `config.json` crons are missing from the dump, injects a forced-recreate nudge.

The 7-day silent cliff is a CronCreate artifact. v0.2.0 daemon-owned external persistent crons (`scheduled_tasks.json` + `src/daemon/cron-scheduler.ts`):
- Are not subject to 7-day auto-expiry
- Are registered + persisted by the daemon at boot from `config.json`, with `lastGoodSchedule` snapshot + ENOSPC fallback
- Do not require agents to mirror their `CronList` to disk for daemon to verify

The primary user pain (recurring crons silently stop firing after 7 days) is structurally eliminated. Hook 2's mismatch-detection mechanism would still have marginal value as a tripwire for any session-only CronCreate edge cases that survive the migration, but it depends on PR #299 (also closed on supersession), and the core problem it addresses is now solved at a different layer.

If post-v0.2.0 monitoring shows persistent crons silently dropping, that's a different bug to file against `cron-scheduler.ts` directly.

Per James 2026-05-05: closing Tier 1 PRs solved by external crons; surviving Tier 1 (#298 agent_crashed alert, #301 approval ping) batch-test via cortext-designer in single sandbox session.

PR #272, opened 2026-04-30
feat(bus): hooks framework — Day-1 stub + Day-2 per-handler wiring + telemetry

noogalabs:
## Summary

Adds an extensible bus-side hooks framework that lets handlers register against hook types and return structured results (`fire` / `block` / `escalate`) for downstream observability + safety. Includes the bus event emit pipeline (`hook_fire` / `hook_block` / `hook_escalate`) so test harnesses and dashboards can count hook outcomes cleanly.

Designed in three phases:
- **Day-1 (stub):** dispatcher exists, treats every attempt as `hook_fire` with `outcome=no_handler_registered`. No handler refusal capability.
- **Day-2 (this PR's added wiring):** `HandlerResult` contract, in-process registry (`registerHandler` / `clearHandlerRegistry` / `_getRegisteredHandler`), dispatcher routes results to the correct event taxonomy. Throws caught as `block(handler_threw)`. Backwards-compat: undefined return = fire, empty registry = Day-1 stub semantic.
- **Day-3 (follow-up PR):** built-in handler implementations (bash spawn / send_message / log_event / webhook fetch) wired against the same `registerHandler` API.

## What changed

- New `src/bus/hooks.ts` — exports:
  - `HandlerResult` type: `{ action: 'fire' | 'block' | 'escalate', reason?: string, meta?: Record<string, unknown> }`
  - `HandlerFn` type: sync-or-async, returns `HandlerResult | void`
  - `registerHandler(type, fn)` / `clearHandlerRegistry()` / `_getRegisteredHandler(type)`
  - `dispatchHook(type, eventMeta)` — routes via registered handler, catches throws as block, emits the right bus event
  - `emitHookBusEvent(eventName, meta)` — fire-and-forget execFile to `cortextos bus log-event`
  - `loadHookRegistry()` / `matchHooks()` — config-driven hook routing
- New `tests/unit/bus/hooks.test.ts` — 18 tests covering: loadHookRegistry (missing file fail-open, malformed JSON fail-open, valid parse), matchHooks (enabled/disabled/agent_filter own+cross, priority sort, metadata deep-match), dispatchHook (no-handler/undefined/fire-with-meta/block/escalate/throw/async/bookkeeping-fields-always-present), registry CRUD.

## Test plan

- [x] 18 new unit tests, all passing
- [x] `npm run build` clean
- [x] TypeScript no errors (`npx tsc --noEmit` clean)
- [x] Backwards-compat verified: empty registry produces same Day-1 stub behavior as before

[...]

PR #305, opened 2026-05-04
Time-window loop-detector history to fix cross-session wedge

loganbronstein:
## Summary

- Loop detector was blocking the first Bash call of every new session because prior-session Skill/Bash alternation history persisted in `loop-detector.json` forever and tripped the 24-alternation ping-pong threshold instantly on boot.
- Fix: filter history to entries within `HISTORY_RETAIN_MS` (60s) inside `decideHookAction`. Real loops fire fast and stay inside the window; any normal idle gap clears stale history.
- Stale entries are also dropped from `nextState.history` on every decision (allow / block / escape) so the on-disk file never resurrects ancient state.

## Test plan

- [x] `npx vitest run tests/unit/hooks/loop-detector.test.ts` — 22/22 pass, including 4 new cross-session-staleness regression tests:
  - boot scenario: stale alternation does not block first new call
  - drops stale entries from nextState.history on allow
  - drops stale entries from nextState.history on block
  - still blocks when alternation is fresh and within retain window
- [x] `npm run build` — clean
- [x] `npm test` — 773 pass; 1 file failure (`dashboard/src/lib/__tests__/sync.test.ts`) is pre-existing `better_sqlite3` ABI mismatch unrelated to this change (verified by stashing this diff and re-running)
- [x] Live verification: hook is rebuilt in `dist/hooks/hook-loop-detector.js`; the wedge that blocked this session's boot self-cleared once the time-windowed build was active


grandamenium:
Hello — overnight sandbox audit flagged a title/commit mismatch.

PR title and tip-commit do not match — title says "Time-window loop-detector history to fix cross-session wedge" but tip is "Fix scribe cron restore and vault memory pipeline" (8 files, +182/-25).

Either:
(a) update PR title to match the tip-commit, OR
(b) force-push a commit that matches the title, OR
(c) close+reopen with the canonical scope.

Squash gate held until resolved.

loganbronstein:
Closing per 2026-05-11 decision. This PR was structurally wrong as targeted:

1. `hook-loop-detector.ts` exists only on `fix/loop-detector-regression` (fork branch). Not on upstream main and not intended to be — it was wired into a fork-specific settings.json pattern.
2. Upstream has its own hook framework at `src/bus/hooks.ts` (landed via PR #272 da437c3). Bringing fork-only files into upstream is the wrong layer.
3. Branch accumulated 22 unrelated commits on top, making the PR diff incoherent.

Path B chosen: loop-detection logic will be re-implemented as a proper hook **inside** upstream's `src/bus/hooks.ts` and shipped via a fresh, clean PR. Plan-first, then implementation, then real-world end-to-end verification on an actual stuck-agent scenario.

This branch's existing work stays available on `fix/loop-detector-regression` as historical reference for the new implementation.

PR #319, opened 2026-05-06
fix(bus): send-telegram hangs indefinitely when rgos-mirror drain wedges

revopsglobal:
## Summary

- `send-telegram` was hanging for up to 83 minutes (observed as 12h zombies) after a successful send when the Supabase retry queue had hundreds of entries
- Root cause: `logEvent` fires `mirrorEventToRgos` via `setImmediate`, which schedules `drainRetryQueue`; each drain entry holds a 10s fetch open, keeping the event loop alive
- Fix: add explicit `process.exit(0)` in the send-telegram action handler immediately after all local writes complete (`logOutboundMessage`, `cacheLastSent`, `logEvent` JSONL) — fires before any `setImmediate` callbacks so the drain never runs in this short-lived CLI process
- Also adds `TELEGRAM_API_BASE_URL` env-var override to `TelegramAPI` for clean e2e test mocking

## Test plan

- [x] 3 new e2e subprocess tests verify exit within 5s: mirror disabled, Supabase unreachable + 50-entry queue, Supabase reachable
- [x] Full suite: 1189/1189 passing
- [x] Verified retry queue already persisted to disk — fire-and-forget contract intact


revopsglobal:
opened in error after --continue restart; correct PR is RevOps-Global-GIT/cortextos#56

PR #325, opened 2026-05-06
fix(cli): hard-restart now kills PID via daemon IPC

loganbronstein:
## Bug

\`cortextos bus hard-restart\` issued from inside an agent's own CLI wrote the \`.force-fresh\` + \`.restart-planned\` markers and printed \`Hard restart planned\`, but did NOT signal the daemon to actually kill the running process. The same PID kept executing with stale settings until some other path (operator, OS, scheduled restart) triggered a restart.

Discovered 2026-05-06 while wiring the coder agent's Telegram guardrail hook. The agent reported \"hard restart planned\" successfully, but the same PID kept running with stale settings cache and the new project hook never loaded. We had to externally \`cortextos stop coder && cortextos start coder\` to actually take effect — exactly what \`hard-restart\` was supposed to do.

## Fix

After writing the markers, send the same daemon IPC \`restart-agent\` signal that \`self-restart\` (\`src/cli/bus.ts:707\`) and \`auto-compact-agent\` (\`src/cli/bus.ts:766\`) already use. The daemon kills the process and respawns it; the \`.force-fresh\` marker (consumed in \`src/daemon/agent-process.ts:421\`) tells the respawn to skip \`--continue\` and start a fresh session.

Mirrors the existing pattern exactly. No new IPC contract.

## Compat

Adds \`--no-ipc\` flag to preserve the legacy marker-only behavior. No in-tree callers need this today, but it leaves the door open for ops scripts that intentionally arm markers and let some other path trigger the restart. The default flips to \"actually restart\" — which is what every documented use of \`hard-restart\` already assumed.

## Tests

- Full unit suite passes (\`npm test\` — 759/760, the one failure is unrelated dashboard comms route, not touched by this PR).
- Bus unit tests specifically: \`npx vitest run tests/unit/bus\` → 197/197 pass.
- Build: \`npm run build\` clean.

## Manual verification path (post-merge)

1. From inside an agent CLI: \`cortextos bus hard-restart --reason \"manual test\"\`
2. Daemon kills PID, respawns without \`--continue\`
3. Agent boots fresh, reads AGENTS.md, restores crons
4. \`.force-fresh\` and \`.restart-planned\` markers are consumed (deleted)

## Related

- task_1778042740480_306 — doc fix #323 (also flags the bug in agent-management skill)

Closes task_1778042747359_985.


loganbronstein:
Closing as duplicate. Upstream main already shipped the same fix in PR #217 (commit a38ef7a, 'fix(bus): hard-restart now sends IPC restart-agent to terminate the session'). I built this on a fork that was 66 commits behind upstream and didn't have that fix yet. The loganbronstein fork's origin/main needs to pull from upstream/main to pick up #217 and the other 65 missing commits — that is a separate ops task, not a code change. Apologies for the noise.

Issue #330, opened 2026-05-06
codex bot: Telegram typing indicators do not show (sendChatAction not wired)


grandamenium:
## Discovery

Found by James during PR #322 soak observation 2026-05-06 ~14:13 UTC. Codex Telegram bot does not show typing indicators while a turn is in flight, where claude-code Telegram bots correctly emit them.

## Originating PR

PR #322 — `feat(pty): add CodexPTY adapter — exec-mode Codex CLI runtime`. Layered fix request per James 2026-05-06 14:00 UTC real-time iteration policy: file issue + route to cortext-designer + commit on top of `feat/codex-pty-v2` + re-cherry-pick into integration branch as single squash.

## Gap

The CodexPTY adapter is missing the periodic `sendChatAction` Telegram bot API call (action `typing`) that the claude-code adapter emits while Claude is producing tokens. This is what drives the "agent is typing..." indicator in the Telegram client.

## Expected Behavior

While the codex agent is processing a turn (between user message inbound and turn.completed JSONL), the Telegram bot should emit `sendChatAction` with `action=typing` periodically (every ~4s — Telegram caches the indicator for 5s).

## Suggested Fix Direction

Hook the typing-action emitter into the codex JSONL stream parsing so it fires on `thread.message.received` (or equivalent inbound signal) and stops on `turn.completed`. Reuse the claude-code adapter's typing-indicator timer pattern if one exists.

## Soak Cycle Context

This is layered fix #3 on PR #322 (after #328 cortextos restart and #slash-commands-issue). Bundle with the slash-command issue — both go into one cortext-designer commit + one re-squash on integration branch `nightly-soak-2026-05-06`.

Issue #329, opened 2026-05-06
codex bot: Telegram slash command menu does not populate (setMyCommands not registered)


grandamenium:
## Discovery

Found by James during PR #322 soak observation 2026-05-06 ~14:13 UTC. When typing `/` in the Telegram codex bot conversation, no slash command menu pops up. Compared against Sentinel (Claude agent) which correctly shows the menu (`/agent_management`, `/catalog_browse`, `/comms`, `/community_publish`, `/cron_management`, etc.).

## Originating PR

PR #322 — `feat(pty): add CodexPTY adapter — exec-mode Codex CLI runtime`. Layered fix request per James 2026-05-06 14:00 UTC real-time iteration policy: file issue + route to cortext-designer + commit on top of `feat/codex-pty-v2` + re-cherry-pick into integration branch as single squash.

## Gap

The CodexPTY adapter is missing the bootstrap step that runs `cortextos bus register-telegram-commands <bot-token> <scan-dirs...>` (the Telegram bot API `setMyCommands` call) which the claude-code adapter performs during agent startup. Without this, no slash commands are registered with Telegram for the codex bot, so the menu stays empty.

## Expected Behavior

Typing `/` in the codex Telegram bot should show the same slash menu populated from the agent's `.claude/skills/` directory, identical to claude-code agents.

## Suggested Fix Direction

Wire the equivalent of the claude-code bootstrap path into `src/pty/codex-pty.ts` (or wherever the CodexPTY adapter does its post-spawn setup). Reuse `cortextos bus register-telegram-commands` rather than duplicating logic.

## Soak Cycle Context

This is layered fix #2 on PR #322 (after #328 cortextos restart). Bundle with #typing-indicator-issue (to be filed) — both go into one cortext-designer commit + one re-squash on integration branch `nightly-soak-2026-05-06`.

PR #334, opened 2026-05-06
feat(pty): redact HuggingFace hf_* tokens at PTY layer

ClintMoody:
Extends the PTY output redactor with a second pattern for HuggingFace-shaped access tokens — `hf_` prefix + 32+ alphanumerics — alongside the existing JWT redaction. Matches the canonical gitleaks rule `hf_[A-Za-z0-9]{32,}` so the redactor and the scan rule agree on what counts as a token.

## Origin
2026-05-06 baseline gitleaks drift caught 27 hf_ token instances persisted to one fleet's boss-agent logs (1 inbound-messages.jsonl + 26 stdout.log). Logs were scrubbed in-place via `sed`; this change closes the hole at the source so future leaks redact at the PTY layer instead of needing another scrub pass.

## Approach
Same defense-in-depth pattern as the existing JWT redactor (fix lives at the PTY layer, not at individual code paths) — any future tool that happens to print a token (`curl -v`, env dump, debug log of a request, etc.) gets redacted the same way.

## Changes
- `src/pty/redact.ts`: add `HF_TOKEN_PATTERN` constant + chain it after JWT replace in `redactSecrets()`. Replacement marker: `[REDACTED_HF_TOKEN]`.
- `tests/unit/pty/output-buffer.test.ts`: 3 new cases — single-token redaction in disk log + ring buffer, length-floor guard at 32 body chars, JWT + HF co-occurrence in same chunk.

## Verification
- `npm run build` clean
- `npx vitest run tests/unit/pty/output-buffer.test.ts` → 9/9 pass

## Known limitations
Inherits the existing chunk-boundary limitation documented in `redact.ts`: if an HF token straddles a 4KB OS-buffer boundary across two `push()` calls, neither chunk matches the regex on its own. Same limitation already documented for JWT — the existing chunk-boundary regression test still covers the contract.


grandamenium:
Sandbox audit (cortext-designer) — flagging a scope mismatch before any squash.

**Stated scope:** redact HuggingFace `hf_*` tokens at the PTY layer (one-feature `feat(pty)` PR).

**Actual diff vs current grandamenium/main:** 91 files changed, +6838 insertions, -1324 deletions. The diff carries far more than the stated scope, including a wholesale removal of `tests/unit/pty/codex-pty.test.ts` (a test file added in #322 / extended for the #329+#330 typing-indicator fix). Squashing this PR as-is would clobber an unrelated PTY adapter test.

**What I think is happening:** the branch base is stale (`009191bd`, well behind main `68055a36`), and a large pre-staged tooling/test branch is being dragged into the diff alongside the actual hf_* redaction change.

**Ask:** could you rebase against current grandamenium/main and force-push so the PR contains only the hf_* redaction hunks? I'll re-test once you ping me. (Same goes for #336/#339/#340 from the same base — separate comments coming on each.)

PR #338, opened 2026-05-06
feat(hooks): PreToolUse loop detector — repetition/ping-pong blocking + essential ops whitelist

revopsglobal:
## Summary

Adds `src/hooks/hook-loop-detector.ts`, a PreToolUse hook that detects and interrupts tool-use loops before they burn through context or spin indefinitely.

### Detection strategies

**Repetition** — the same (tool, args-hash) pair appears ≥ 3 times in the last 20 calls. The hook returns a `block` decision with a diagnostic message asking the agent to reconsider.

**Ping-pong** — two tools alternate and together dominate ≥ 80% of the last 20 calls (e.g. Edit → Bash → Edit → Bash…). Same block behaviour.

### Essential ops whitelist

Operations that agents must repeat legitimately (heartbeat, inbox check, messaging, task polling, reading bootstrap files) are whitelisted and never contribute to block thresholds. They are still recorded in history so that non-essential repetition is still detected even when essential calls are interspersed.

**Whitelisted operations:**
- `cortextos bus update-heartbeat`, `check-inbox`, `send-message`, `ack-inbox`, `send-telegram`, `log-event`, `update-task`
- Reads of `HEARTBEAT.md`, `GOALS.md`

### Also included

Fixes 3 cron-scheduler tests that assumed UTC but `cron-parser` uses system timezone, causing spurious failures on non-UTC CI runners.

## Test plan

- [ ] `npm run build` — TypeScript compiles clean
- [ ] `vitest run tests/unit/hooks/loop-detector.test.ts` — 28/28 pass
- [ ] `vitest run tests/unit/daemon/cron-scheduler.test.ts` — 28/28 pass
- [ ] Register hook in `.cortextos/settings.json` `hooks.PreToolUse` array; trigger a known loop pattern and confirm block fires

## Traceability

Source: RevOps-Global-GIT/cortextos
- `a6897dc` feat(hooks): PreToolUse loop detection — repetition and ping-pong blocking
- `51d5430` fix(loop-detector): exempt cron scripts from repetition blocking
- `7c61e95` fix(hooks): add essential ops whitelist to loop detector + fix timezone tests
- `d6c2ef5` fix(loop-detector): exempt analyst activity-sync script from ping-pong blocking

### DIFF
--- a/src/hooks/hook-loop-detector.ts
+++ b/src/hooks/hook-loop-detector.ts
@@ -1,0 +1,327 @@
+/**
+ * hook-loop-detector.ts — PreToolUse hook: detects and blocks repeated tool loops.
+ *
+ * Ports the two-strategy detection algorithm from desplega-ai/agent-swarm into
+ * a cortextOS PreToolUse hook.
+ *
+ * Detection strategies:
+ *   1. Repetition — the same (tool, args-hash) pair appears ≥ REPETITION_BLOCK
+ *      times in the last HISTORY_SIZE calls.
+ *   2. Ping-pong — two tools alternate and together dominate ≥ 80% of the last
+ *      PINGPONG_WINDOW calls.
+ *
+ * On detection the hook writes a block decision to stdout and exits 0.
+ * All other paths exit 0 silently so the tool call proceeds.
+ *
+ * State: {ctxRoot}/state/{agentName}/loop-detector.json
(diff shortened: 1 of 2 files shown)


revopsglobal:
Closing — not pushing to grandamenium per RevOps-Global policy. Will keep these as internal-only on the fork.

PR #339, opened 2026-05-06
feat(daemon): write exit-reason.json so dashboard can distinguish planned restarts from crashes

ClintMoody:
The SessionEnd crash-alert hook (`src/hooks/hook-crash-alert.ts`) already classifies every session end — `planned-restart`, `session-refresh`, `daemon-stop`, `user-stop`, `rate-limited`, `daemon-crashed`, `crash` — and appends one line to `crashes.log`. The dashboard, however, has no way to tell at-a-glance whether an agent that is currently down is "intentionally stopped" or "actually crashed" without tailing-and-parsing the audit log on every refresh.

This change writes a small structured `exit-reason.json` into the agent's state dir alongside the existing `crashes.log` append. The file is **overwritten** on each session end (point-in-time snapshot, not an append log) so the dashboard reads the most recent exit reason as JSON without any log-parsing.

## Schema

`state/<agent>/exit-reason.json`:

```json
{
  "type": "planned-restart" | "crash" | "rate-limited" | "user-stop" | "daemon-stop" | "user-restart" | "user-disable" | "session-refresh" | "daemon-crashed",
  "timestamp": "<ISO 8601 UTC>",
  "reason": "<string>",
  "last_task": "<string>"
}
```

The four fields mirror `crashes.log` exactly so the two sources stay consistent.

## Approach

Piggybacks on the existing `crashes.log` append site (`hook-crash-alert.ts:241`) instead of consolidating from the multiple stop-marker writer call-sites (`cli/stop.ts`, `daemon/agent-manager.ts`, `daemon/index.ts`). The hook already does the full classification work — re-using its endType/reason/lastTask values keeps the logic in one place.

The write is **best-effort** and never throws — a write failure cannot block the alert path. `crashes.log` remains the audit log; `exit-reason.json` is the dashboard-friendly snapshot.

## Tests (4 new, all in `tests/unit/hooks/hook-crash-alert.test.ts`)

- `writes a JSON file with all four fields populated` — happy path
- `overwrites the previous exit-reason atomically (latest wins)` — snapshot semantics, not append
- `preserves empty-string fields rather than dropping them` — dashboard parsing relies on shape
- `does not throw when the state dir is not writable (best-effort write)` — IO-failure safety

786/786 vitest pass; build clean.

## Dashboard integration (out of scope here)

The dashboard reader is a separate change and depends on this file existing first. Suggested logic: if the agent process is **running**, ignore `exit-reason.json` (it is the previous-exit history). If the agent is **not running**, `exit-reason.json` tells you why. Missing file means "never exited" (fresh agent) — render as a benign empty state.


grandamenium:
Sandbox audit (cortext-designer) — same scope-mismatch pattern as siblings #334/#336/#340.

**Stated scope:** `feat(daemon)` write `exit-reason.json` so dashboard can distinguish planned restarts from crashes.

**Actual diff vs current grandamenium/main:** 80 files changed, +6860 insertions, -218 deletions. Way more than a single exit-reason marker file feature.

**Likely cause:** stale base `009191bd` vs main `68055a36` dragging in unrelated commits.

**Ask:** rebase against current grandamenium/main, push, ping me for re-test. (Tonight's autonomy cycle is gating on a clean rebase before any squash into nightly-soak.)

PR #348, opened 2026-05-07
fix(env): enforce CTX_AGENT_DIR subordination to CTX_FRAMEWORK_ROOT (#313)

### LINKED ISSUE #313: security: sandbox daemon must pin CTX_FRAMEWORK_ROOT — isolation breach near-miss on PR #309 review
## Summary

During review/testing of PR #309, a near-miss isolation breach was discovered: when a sandbox daemon is spawned without `CTX_FRAMEWORK_ROOT` pinned in its environment, it falls through to the live agent config paths — exposing live `BOT_TOKEN` and agent config to the sandbox context.

The breach was prevented only by:
1. Telegram's 409 conflict lock (duplicate bot token rejected)
2. OS `posix_spawnp` failure

Both are accidental guards, not intentional isolation guarantees.

## Root cause

Sandbox daemon spawning does not assert `CTX_FRAMEWORK_ROOT` is set to an isolated path before starting. Without this pin, any bus command that resolves paths via `CTX_FRAMEWORK_ROOT` will silently use the live org's state directory.

## Impact

- **Bot token leak**: sandbox agent inherits live `BOT_TOKEN` and begins accepting Telegram messages intended for the production agent
- **Config corruption**: sandbox writes (heartbeat, task state, cron state) land in live agent state directory
- **Silent failure mode**: no error is thrown — the sandbox appears to work normally while operating against production state

## Required fix

Before spawning any sandbox/test daemon:
1. Assert `CTX_FRAMEWORK_ROOT` is explicitly set to an isolated workspace path
2. Assert `CTX_INSTANCE_ID` is unique and distinct from production
3. Assert `BOT_TOKEN` is either unset or a test-only token

Recommend a `preflight-sandbox-env.sh` check that validates all three before `daemon start`.

## References

- Discovered during PR #309 review (2026-05-05)
- Memory entry: `project_sandbox_ctx_framework_root.md`
- Related: issue #188 (`.env must exist before enable-agent.sh`)

grandamenium:
## Summary

Closes #313 — sandbox/live env-isolation leak class.

Adds two path-divergence canaries inside `resolveEnv()` (the shared env-assembly bottleneck for every CLI subcommand):

1. When both `CTX_FRAMEWORK_ROOT` and `CTX_AGENT_DIR` are set, the resolved `agentDir` must be subordinate to `frameworkRoot`. Otherwise throws.
2. When both `CTX_PROJECT_ROOT` and `CTX_FRAMEWORK_ROOT` are set, they must be equal. Otherwise throws.

Both canaries fail-fast at env-resolution time, before any filesystem write can land in the wrong tree.

## Why this is needed

The PR-242 TC-C precedent surfaced the actual leak vector: a CLI subprocess (`pty-send`, `bus send-message`, etc.) launched from a **live agent shell** inherits `CTX_AGENT_DIR` (and/or `CTX_PROJECT_ROOT`) from the parent. If the runner overrides only `CTX_FRAMEWORK_ROOT` thinking that fully redirects the install, the inherited per-agent vars silently point the resolved `agentDir` back at the live install — and writes (config edits, experiment files, hook outputs) land there instead of the sandbox.

The leak is hard to spot because every individual var is set to a plausible-looking value; only the *combination* is incoherent.

## Why this layer

`resolveEnv()` in `src/utils/env.ts` is the single bottleneck where every CLI subcommand assembles its env context. Catching it here is one diff that protects every code path.

A daemon-side check was considered and rejected: `CTX_AGENT_DIR` and `CTX_PROJECT_ROOT` aren't daemon-boot env vars — they're per-agent CLI subprocess env, set by the daemon at PTY-spawn time and inherited from the parent shell when invoked manually. The daemon never sees the leak; only the resolution layer does.

## Scope

- Narrow, additive: one import change + one new check block in `resolveEnv()`.
- No behavior change for correctly-configured installs (back-compat covered by TC-D).
- Preflight check (warn if any `CTX_*` is set when `--instance-id` flag forces an override) is deferred to a follow-up.

## Test coverage

`tests/sprint7-environment.test.ts` adds a new `Sandbox/live env subordination (issue #313)` describe block with 4 cases:

| TC | Scenario | Expected |
|----|----------|----------|
| TC-A | `agentDir` under `frameworkRoot` | resolves without throwing |
| TC-B | `agentDir` outside `frameworkRoot` (live-leak shape) | throws \`not under CTX_FRAMEWORK_ROOT\` |
| TC-C | `projectRoot` diverges from `frameworkRoot` | throws \`must equal CTX_FRAMEWORK_ROOT\` |
| TC-D | back-compat happy-path with derived `agentDir` | resolves; `agentDir` derives correctly |

## Test plan

- [x] \`npm run typecheck\` — exit 0
- [x] \`npm run lint\` — exit 0
- [x] \`npm run build\` (CLI) — exit 0
- [x] \`cd dashboard && npm run build\` — exit 0
- [x] \`npx vitest run tests/sprint7-environment.test.ts\` — 13/13 PASS (9 existing + 4 new)
- [ ] sandbox-executor self-validation against this PR (Phase 1-5)

### DIFF
--- a/src/utils/env.ts
+++ b/src/utils/env.ts
@@ -1,5 +1,5 @@
 import { readFileSync, existsSync, writeFileSync } from 'fs';
-import { join, basename } from 'path';
+import { join, basename, resolve as resolvePath, sep } from 'path';
 import { homedir } from 'os';
 import type { CtxEnv } from '../types/index.js';
 import { ensureDir } from './atomic.js';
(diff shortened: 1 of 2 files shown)


grandamenium:
Pipeline review (boris, nightly cycle 2026-05-09): 9/10. Closes #313. Sandbox PASS via cortext-designer, full result.md at `orgs/testorg/agents/cortext-designer/pr-reviews/313/result.md`. CLEAN/MERGEABLE against main `0220f138`. Ready for owner review/merge.

PR #355, opened 2026-05-07
fix(daemon): auto-update cron-state.json when daemon fires a cron

fil-up:
## Summary

- After `CronScheduler` fires a cron successfully, call `updateCronFire(stateDir, name, interval)` to stamp `cron-state.json`
- Eliminates the gap-detector nudge spam ("[SYSTEM] Cron gap detected") for crons that fire on schedule
- 12-line change in `src/daemon/cron-scheduler.ts`, failure is non-fatal (caught + suppressed)

## Problem

The gap-detection loop in `agent-process.ts` reads `cron-state.json` to determine when each named cron last fired. `cron-state.json` was only updated when an agent explicitly called `cortextos bus update-cron-fire <name>`. Daemon-scheduled crons (external crons in `crons.json` fired by `CronScheduler`) never called this.

Result: once a cron's gap exceeded 2× its interval, the gap-detector emitted a `[SYSTEM] Cron gap detected` nudge every 10 minutes — indefinitely, even when crons were firing normally.

In a 5-agent fleet this produced 60+ spurious nudges per day, consuming context and reducing signal-to-noise for all agents.

## Fix

After `fireWithRetry` succeeds in the CronScheduler tick loop, call `updateCronFire()` to write the fire timestamp to `cron-state.json`. The gap-detector then sees the cron as recently fired and suppresses the nudge until a real gap develops.

## Test plan

- [ ] Start daemon with an agent that has at least one external cron in `crons.json`
- [ ] Let the cron fire once
- [ ] Confirm `cron-state.json` in the agent's state dir is updated with the correct `last_fire` timestamp
- [ ] Wait 2× the cron interval — confirm no gap-detector nudge is injected


fil-up:
Closing in favour of a cleaner branch that contains only the cron-scheduler fix (1 file, 12 lines). New PR incoming.

PR #359, opened 2026-05-08
fix(qa): harden sweep timing — networkidle wait + retry-on-empty for tables

revopsglobal:
## Summary

- `waitForSettle` replaced fixed 2.5s timeout with `networkidle` wait (8s cap) — catches React Query data fetches that fire after `domcontentloaded`
- `interactTable` gets a retry loop: waits up to 5s for a visible row OR explicit empty-state element before recording N/A; if neither appears, waits 3s and retries once
- Eliminates false "No table rows found" on pages like `/companies` (642 rows), `/app/cortex/skills` (393 rows), `/app/fleet/agents` (4 agents) where data is real but hadn't loaded during the previous sweep

## Root cause

Pages using React Query mount the component first, then fire `supabase.from(...)` queries. A fixed 2.5s wait races against that fetch. The first QA sweep produced 5 spurious "No table rows found" entries for pages with confirmed live data. This PR makes the wait data-driven, not time-driven.

## Test plan

- [ ] Run `DISPLAY=:99 npx tsx scripts/hub-full-sweep.ts` — verify /companies, /app/cortex/skills, /app/fleet/agents, /app/agents show row data or explicit empty-state (not timing-ambiguous N/A)
- [ ] Confirm sweep completes without hanging (networkidle 8s cap prevents stalls)


revopsglobal:
Reopening against correct fork — grandamenium is upstream, PRs should target RevOps-Global-GIT/cortextos.

PR #383, opened 2026-05-09
feat(telegram): handle edited_message updates

storypixel:
## Summary

- Telegram's `getUpdates` emits edits as a separate `edited_message` update type. The poller's current `allowed_updates` whitelist excludes it, so Telegram never sends edits to the daemon — bots see edits land in the UI but can't react to them.
- This patch wires up edit handling end-to-end while keeping the channel opt-in.

## What changed

| File | Change |
|------|--------|
| `src/telegram/api.ts` | Add `'edited_message'` to `allowed_updates` |
| `src/types/index.ts` | Add `edited_message?: TelegramMessage` to `TelegramUpdate` with doc comment |
| `src/telegram/poller.ts` | Add `editedMessageHandlers` + `onEditedMessage(handler)` + dispatch in `pollOnce()` with the same offset-after-handler / fail-defers semantics as `message` |
| `tests/unit/telegram/poller.test.ts` | 4 new tests covering the new path |

## Why a separate handler list (not re-emitting through `onMessage`)

Most chat-style consumers reply when they see a message. If edits flowed through `onMessage`, every edit would re-trigger a reply to the same conversation point — annoying for users, costly for the agent. Making the edit channel opt-in via `onEditedMessage` lets consumers that DO care about edits (audit trails, content moderation, redact-after-send detection) subscribe explicitly, while existing consumers stay quiet on edits by default.

## Test plan

- [x] `npm run build` — clean
- [x] `npx vitest run tests/unit/telegram/poller.test.ts` — 12 passed (8 existing + 4 new)
- [x] New tests cover:
  - `onMessage` does NOT fire on `edited_message` (no double-reply)
  - `onEditedMessage` fires with correct payload
  - Offset advances after successful edit handler
  - Offset does NOT advance if edit handler throws (Telegram redelivers)

### DIFF
--- a/src/telegram/api.ts
+++ b/src/telegram/api.ts
@@ -359,7 +359,7 @@ export class TelegramAPI {
     return this.post('getUpdates', {
       offset,
       timeout,
-      allowed_updates: ['message', 'callback_query', 'message_reaction'],
+      allowed_updates: ['message', 'edited_message', 'callback_query', 'message_reaction'],
     });
   }
 
--- a/src/telegram/poller.ts
+++ b/src/telegram/poller.ts
@@ -5,6 +5,7 @@ import { TelegramAPI } from './api.js';
 import { ensureDir } from '../utils/atomic.js';
 
 export type MessageHandler = (msg: TelegramMessage) => void;
+export type EditedMessageHandler = (msg: TelegramMessage) => void;
 export type CallbackHandler = (query: TelegramCallbackQuery) => void;
 export type ReactionHandler = (reaction: TelegramMessageReaction) => void;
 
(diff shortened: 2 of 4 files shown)


storypixel:
Closing this PR — on second review, the scope was narrower than the framing.

The patch adds poller plumbing for `edited_message` updates, but no production daemon consumer subscribes to `onEditedMessage`, so the daemon would receive edits and silently acknowledge them without any agent ever seeing the change. The "handle edited_message updates" framing oversells what the code does.

Other gaps surfaced on review:
- `TelegramMessage` is missing `edit_date?: number` per the official Telegram Bot API; the new test had to use `as any` to include it
- The 4 new tests use a stub api that injects `edited_message` directly — they don't verify that `api.ts` `allowed_updates` actually includes `'edited_message'` (could remove that line and tests still pass)
- `edited_channel_post` and `edited_business_message` aren't addressed and weren't stated as non-goals
- Doc comment on `TelegramUpdate.edited_message` references `TelegramPoller.onEditedMessage` from the central type module (cross-layer leak)

Will re-cut as v2 with the type fixes + a direct allowed_updates test + either a daemon-side handler so edits are end-to-end or a tighter "add poller support for edited_message" framing. Apologies for the noise.

PR #402, opened 2026-05-12
telegram: stop HTML-escaping in plain-text mode

storypixel:
## Summary

Fix a one-line ordering bug in `src/telegram/api.ts` `markdownToHtml`: plain-text mode was HTML-escaping `&`, `<`, `>` as step 1 *before* the early-return for the `plainText` flag, so the escaped entities leaked into the rendered Telegram message instead of being sent as raw characters.

Real-world symptom: agents emitting `->` in plain-text bodies got `-&gt;` in the rendered chat; `5 > 4` rendered as `5 &gt; 4`. These leak into customer-facing copy and look like unprofessional gremlins.

## The fix

Move the early-return above the escape step.

```diff
   private markdownToHtml(text: string, plainText = false): string {
-    // Step 1: HTML-escape (& must be first to avoid double-escaping)
-    let html = text
-      .replace(/&/g, '&amp;')
-      .replace(/</g, '&lt;')
-      .replace(/>/g, '&gt;');
-
-    if (plainText) return html;
+    // Plain-text mode: skip both HTML-escaping AND Markdown conversion.
+    // sendChunk calls sendMessage without parse_mode, so Telegram renders
+    // the raw text. HTML-escaping in this path leaks visible &gt; / &lt; /
+    // &amp; into the rendered message.
+    if (plainText) return text;
+
+    // Step 1: HTML-escape (& must be first to avoid double-escaping)
+    let html = text
+      .replace(/&/g, '&amp;')
+      .replace(/</g, '&lt;')
+      .replace(/>/g, '&gt;');
```

Also updated the JSDoc above the function to reflect the corrected behaviour.

## Why this is safe

- Default path (`plainText = false`) is unchanged: full HTML-escape + Markdown conversion still runs identically.
- Plain-text path now genuinely sends raw text. `sendChunk` already passes `parseMode = null` for plain-text mode, so Telegram does no HTML parsing on the receiving end -- the raw `>` arrives as `>`.
- No new tests added: the bug is a simple ordering issue and the surrounding behaviour is unchanged for every non-plain-text path.

## Verified

Diagnosed end-to-end from a real production gremlin in agent output (DIY Daily newsletter project): `fingers` agent sent a plain-text Telegram message containing `->` and `&gt;` appeared in the user's Telegram client. Traced through `bus.ts send-telegram` command -> `TelegramAPI.sendMessage` -> `markdownToHtml` and confirmed the escape step ran before the plainText short-circuit. One-line move puts the function in the state the code's own JSDoc + the `--plain-text` flag's documented behaviour already promised.

### DIFF
--- a/src/telegram/api.ts
+++ b/src/telegram/api.ts
@@ -109,16 +109,25 @@ export class TelegramAPI {
    *   5. Italic (_..._) — word-boundary aware to avoid snake_case false positives
    *   6. Links ([text](url)) → <a href="url">text</a>
    *
-   * Pass `plainText: true` to skip conversion (just HTML-escape and send raw).
+   * Pass `plainText: true` to skip both Markdown conversion AND HTML-escaping.
+   * sendChunk calls sendMessage without parse_mode in that case, so Telegram
+   * renders the raw text -- escaping would leak visible `&gt;` / `&lt;` /
+   * `&amp;` into the message.
    */
   private markdownToHtml(text: string, plainText = false): string {
+    // Plain-text mode: skip both HTML-escaping AND Markdown conversion.
+    // sendChunk calls sendMessage without parse_mode, so Telegram renders
+    // the raw text. HTML-escaping in this path leaks visible &gt; / &lt; /
+    // &amp; into the rendered message — agents emitting "->" or "5 > 4" in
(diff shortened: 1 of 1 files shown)


storypixel:
Bumping this gently — happy to rebase or address anything that would help land it. Has been clean and merge-ready since 5/12.

PR #406, opened 2026-05-12
telegram: add react-telegram command for single-emoji acks

storypixel:
## Summary

Adds `cortextos bus react-telegram <chat-id> <message-id> [emoji]` so agents can acknowledge a Telegram message with a single emoji reaction instead of sending a full verbal reply. Today every agent response is a message, which clutters chats with low-content "got it / on it / sounds good" noise. Reactions are the right shape for one-bit acks.

## What changed

- **`TelegramAPI.setMessageReaction(chatId, messageId, emojis)`** in `src/telegram/api.ts` — wraps Telegram's `setMessageReaction` Bot API endpoint. Pass an empty array to clear the bot's reaction. Bots can only react within ~48h of message send (same window as `deleteMessage`).
- **`cortextos bus react-telegram <chat-id> <message-id> [emoji]`** in `src/cli/bus.ts` — CLI command mirroring `send-telegram` (same bot token resolution flow, same env-fallback, same error shape). Default emoji is `👍`. Pass empty string to clear.
- **Default agent templates** in `src/cli/init.ts` and `src/cli/add-agent.ts` — documented the new command alongside the existing `send-telegram` entry in the SYSTEM.md / USER.md scaffolds so newly-created agents inherit awareness.

## Why this is safe

- Pure additive: no existing code path touched, no signatures changed. Existing `send-telegram`, `sendMessage`, etc. unchanged.
- Bot token resolution mirrors `send-telegram` byte-for-byte (agent `.env` first, then `process.env`).
- Telegram's `setMessageReaction` is server-validated; invalid emoji or out-of-window message returns an explicit API error which the CLI surfaces.

## Verified end-to-end

```
> # send a probe message, capture its id
> message_id: 580
> cortextos bus react-telegram 5459602728 580 👍
Reacted 👍
```

Visible thumbs-up reaction confirmed on the recipient's Telegram client.

## What's NOT in this PR

The usage policy — *when* an agent should react vs send a verbal reply — lives in each agent's IDENTITY/SOUL prompt, not in framework code. This PR ships the capability; the prompt rules ship per-agent. Suggested policy shape for agent prompts:

> React with an emoji instead of sending a verbal reply when ALL of these are true: you have nothing to add beyond acknowledging the message; the message doesn't ask a question, carry a directive needing status, or deliver content you must engage with; silence would feel ambiguous but a verbal "got it" would be noise.
>
> Defaults: 👍 = got it / acked. 👀 = looking at it. 🔥 = endorsing a take. ✅ = done. When in doubt, verbal.

That belongs in agent dirs, not here.

### DIFF
--- a/src/cli/add-agent.ts
+++ b/src/cli/add-agent.ts
@@ -239,6 +239,7 @@ export const addAgentCommand = new Command('add-agent')
             '',
             '- Agent-to-agent: `cortextos bus send-message <agent> <priority> "<text>"`',
             '- Telegram to user: `cortextos bus send-telegram <chat_id> "<text>"`',
+            '- React to a Telegram message (single emoji ack, no verbal noise): `cortextos bus react-telegram <chat_id> <message_id> 👍`',
             '- Check inbox: `cortextos bus check-inbox`',
             '',
           ].join('\n');
@@ -447,5 +448,6 @@ Complete tasks: \`cortextos bus complete-task <id> --result "<text>"\`
 Log events: \`cortextos bus log-event <category> <event> <severity>\`
 Update heartbeat: \`cortextos bus update-heartbeat "<status>"\`
 Send Telegram: \`cortextos bus send-telegram <chat_id> "<text>"\`
+React to Telegram message (single emoji ack): \`cortextos bus react-telegram <chat_id> <message_id> 👍\`
 `;
 }
--- a/src/cli/bus.ts
+++ b/src/cli/bus.ts
@@ -1023,6 +1023,54 @@ busCommand
   });
 
 busCommand
+  .command('react-telegram')
+  .description("Set the bot's reaction on a Telegram message (single emoji ack)")
+  .argument('<chat-id>', 'Telegram chat ID')
+  .argument('<message-id>', 'ID of the message to react to')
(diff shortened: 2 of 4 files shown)


grandamenium:
Stacked onto `nightly-soak-2026-05-14` at 15:00Z — squash `3b7bd34e` (6th PR on stack: #423 `74fbba7b` · #411 `d032f9ed` · #412 `f8ff2b81` · #413 `63a92a9d` · #414 v2 `0e6a6c17` · #406 `3b7bd34e`). Designer PASS 9/10 from 2026-05-12T22:54Z (the wait was the now-deprecated 'HELD for restack call' pattern, killed in the new P0-P6 phasing). Clean squash, no conflicts with the prior 5 — touches `src/telegram/api.ts` + `src/cli/{add-agent,bus,init}.ts` only. Phase 4.5 all PASS; unit tests pre-existing cron-scheduler flake only. P3 daemon swap ~01:00Z 2026-05-15; P5 morning validation 13:00Z produces per-PR causal verdict for James merge call.

storypixel:
Bumping — happy to rebase or scope down if needed. Let me know what would help get this in.

PR #430, opened 2026-05-14
feat(bus): add task-comment for mid-task progress notes

hank9991:
Closes #429.

## Summary

Adds \`cortextos bus task-comment <id> <text>\`, a new bus command that appends a free-form note to a task's existing audit log without changing the task's status. Comments render in \`bus task-history\` chronologically alongside create / claim / update / complete entries.

## Motivation

Long-running tasks (multi-step agent work, multi-hour builds, blocked-on-PR-review states) emit two audit entries today: claim and complete. Any \"how's it going\" signal in between either pollutes the lifecycle (flipping status away and back) or goes to a parallel log channel that doesn't show up next to the task it belongs to. \`task-comment\` is the missing primitive — observability, not state.

Use cases I had in mind:

- Director dropping a heartbeat note every ~30 min on an in-progress task so an orchestrator sees \"still home, no blockers\".
- Watcher (operator, oncall, dashboard) annotating someone else's task without racing the owning agent's writes.
- Pre-completion notes for unusual outcomes that don't fit in \`--result\` (e.g. \"completed, but retry rate was 3× normal, see comment\").

## Change

- \`src/bus/task.ts\`: \`TaskAuditEntry.event\` extended with \`'comment'\`; new exported \`commentTask(paths, taskId, agent, text)\` helper that validates, then calls \`appendTaskAudit\`. No mutation of the task JSON.
- \`src/cli/bus.ts\`: new \`task-comment\` subcommand mirroring \`claim-task\`'s ergonomics (\`--agent\` falls back to \`CTX_AGENT_NAME\`). \`task-history\` description updated to mention comments.
- \`tests/unit/bus/task.test.ts\`: 6 new tests under a fresh \`describe('commentTask …')\` block covering: append+status-unchanged, multi-comment interleaved with transitions, empty/whitespace rejection, missing-agent rejection, cross-org not-found, and whitespace trimming.

Diff is \`+143 / -6\` across three files. No new dependencies.

## Backwards compatibility

Additive only.

- New audit entries carry \`event: 'comment'\`. Consumers of \`readTaskAudit\` should now expect this in the union; existing code that only checks against the previous four values will treat them as \"unknown event\" and most callers will still print the \`note\` because the display path is event-agnostic on text. The TypeScript type was updated so the compiler will flag any exhaustive switch that needs widening.
- Task JSON shape is unchanged. \`update-task\` / \`claim-task\` / \`complete-task\` ordering is unchanged.

## Tests

\`\`\`
[...]

Issue #429, opened 2026-05-14
feat(bus): mid-task progress notes — task-comment


hank9991:
## Context

\`cortextos bus\` records lifecycle events for every task — create / claim / update / complete — as append-only JSONL audit entries (\`src/bus/task.ts\`, \`TaskAuditEntry\`). \`bus task-history <id>\` replays them.

What's missing is a way for an agent to drop a free-form note *between* status transitions: "halfway through, no blockers", "waiting on PR review", "model ate 8 min, retrying with smaller chunk". Today the only options are:

1. Flip status to something else and back (pollutes the lifecycle), or
2. Use \`bus log-event\` (lives in a separate stream, not visible on the task's history).

## Use case

Driving a multi-agent setup (a chief-of-staff agent dispatching directors), the operator wants mid-task visibility without forcing the assignee to either complete prematurely or hand-roll a parallel log channel. Status transitions tell you when work started and ended; comments tell you whether anyone is still home in between.

## Proposal

Add a new bus command:

\`\`\`
cortextos bus task-comment <id> <text> [--agent <name>]
\`\`\`

Behaviour:

- Appends \`{ ts, event: 'comment', agent, note: <trimmed text> }\` to the existing audit log.
- Does *not* touch the task JSON — comments are observability, not state.
- Rejects empty / whitespace-only text and missing agent.
- Resolves agent from \`CTX_AGENT_NAME\` when \`--agent\` is omitted (same as \`claim-task\`).
- Comments render in \`bus task-history\` chronologically alongside status transitions via the existing display path.

## PR

Patch ready: I'll attach as a follow-up linked PR. Diff is contained:

- \`src/bus/task.ts\`: \`TaskAuditEntry.event\` union extended to include \`'comment'\`; new exported \`commentTask\` helper.
- \`src/cli/bus.ts\`: new \`task-comment\` subcommand.
- \`tests/unit/bus/task.test.ts\`: 6 new tests covering happy path, interleaved with status changes, multi-agent commentary, empty/whitespace rejection, missing-agent rejection, cross-org not-found, and whitespace trimming.

Out of scope for this issue: dashboard UI to render and post comments. That belongs in a follow-up that consumes \`readTaskAudit\` from the new shape.

PR #431, opened 2026-05-14
feat(dashboard): task history feed + comment input on task detail

hank9991:
Follow-up to #429 / #430. This patch surfaces the new \`task-comment\` audit event in the dashboard so operators see the full task lifecycle — status transitions and comments — in one chronological feed.

## Summary

- New \`GET /api/tasks/[id]/history\` returns the audit JSONL as JSON. Reads the file directly so the dashboard does not depend on the bus daemon being reachable.
- New \`POST /api/tasks/[id]/history\` appends a comment. Resolves agent from the NextAuth session (falls back to \`dashboard-user\`).
- New \`task-history-feed.tsx\` renders the feed: status transitions as compact one-liners, comments as agent-headed bubbles with relative timestamps (\`date-fns\` \`de\` locale).
- \`task-detail-sheet.tsx\` loads history on open, optimistic-updates on submit, surfaces errors inline.
- \`task-card.tsx\` gains optional \`commentCount\` + \`lastComment\` props and renders an \`IconMessage\` badge when there are comments. (See follow-up.)
- Dashboard-side \`TaskAuditEntry\` / \`TaskAuditEvent\` types mirror the bus shape, kept self-contained so this build is not blocked on #430 landing first.

## Why a separate PR from #430

The dashboard surface stands up to review independently of the bus command and ships behind its own test surface (14 new tests). Combining them would have made the diff harder to review without buying anything operationally — \`commentTask\` lands as a callable primitive from #430, and this PR is the dashboard consumer.

## Screenshots

- Detail sheet, history feed visible: see \`task-detail-history.png\` / \`task-detail-history-scrolled.png\` (uploaded in conversation; happy to host elsewhere if preferred).
- Detail sheet, comment input focused: \`task-detail-comment-input.png\`.

(I can attach via a follow-up comment if reviewer prefers them inline.)

## Tests

\`\`\`
$ npx vitest run --root . dashboard/src/app/api/tasks/__tests__/history.test.ts
 Test Files  1 passed (1)
      Tests  14 passed (14)
\`\`\`

\`npx tsc --noEmit\` is clean in \`dashboard/\`. Full dashboard suite still passes locally (no regression in 126 tests).

## Follow-ups intentionally not bundled

1. **Kanban badge data wiring.** \`task-card.tsx\` already accepts \`commentCount\` / \`lastComment\` and the badge renders when they're present, but \`kanban-board.tsx\` does not yet pass them through. Doing so requires aggregating per-task audit summaries somewhere — either an N+1 read on \`getTasks()\` or a cache field on the task JSON. Both have trade-offs and I'd rather take a maintainer steer than guess. Issue follow-up will spell out the two options.
2. **Realtime push.** The detail sheet polls on open and revalidates on POST; it does not subscribe to \`/api/events/stream\`. Out of scope here, low-cost follow-up.
3. **JSONL writer dedup.** \`appendComment\` in the dashboard data layer duplicates the JSONL line shape that \`commentTask\` in \`src/bus/task.ts\` produces. Both formats must stay in sync. The alternative is to spawn \`cortextos bus task-comment\` from the route (mirroring how \`/api/tasks/[id]\` shells out to \`update-task.sh\`). Happy to switch if you prefer the spawn pattern — kept the direct write for now because it's the same node process and one fewer moving piece in dev.

## Test plan

- [x] New history-route tests pass (14/14)
- [x] Dashboard tsc clean
- [x] Manual: open task detail, history feed renders both status transitions and comments chronologically
- [x] Manual: submit comment via the input, optimistic entry appears, server entry persists
- [x] Manual: error path (empty text) surfaces inline
- [ ] Manual: kanban badge after wiring follow-up #1


grandamenium (approved):
## QA Review — PASS

Dashboard companion to #430 (task-comment). API design is solid: GET reads audit JSONL directly (no daemon dependency), POST resolves agent from NextAuth session with fallback to 'dashboard-user'. New `task-history-feed.tsx` renders status transitions and comments in one chronological feed. Test coverage: 353 lines covering API routes. CI green.

PR #436, opened 2026-05-14
fix: address Codex auto-review findings from noogalabs/ascendops PR #24+#25 (3 patches)

noogalabs:
## Summary

Two small, targeted fixes for findings surfaced by Codex auto-review on the noogalabs/ascendops persona-template PRs (#24 Maintenance Director, #25 Leasing Coordinator). Each commit addresses one finding; convergent root cause is that runtime / org context wasn't being preserved across boundary crossings.

A third finding (codex-app-server thread resume on daemon restart) is more invasive — it touches startup-flow semantics across all codex agents — and will land in a separate PR per the spec in the noogalabs/ascendops fork.

A fourth finding (NON_CODEX_TEMPLATES allowlist extension for agent-maintenance-director / agent-leasing-coordinator) was intentionally scoped DOWN to noogalabs/ascendops only — those persona templates don't exist on grandamenium/cortextos, so the deny-list entries are inert here. Per the upstream-clean scope rule, PM-domain template names stay downstream.

## Commits

1. **378cd1f** fix(cli): preserve runtime field when importing exported agents — P1. import-agent rewrote config.json from scratch and never carried over the runtime field. Codex-app-server / hermes agents silently fell back to claude-code on import, breaking the daemon spawn.
2. **80da075** fix(bus): write task completion events to task owner's analytics tree — P2. completeTask supports cross-org completion via findTaskFile, but the post-write logEvent call reused the caller's paths.analyticsDir, misrouting the event into the caller's analytics dir. Build a paths variant scoped to the task's own org for the logEvent call.

## Test plan

- npx tsc --noEmit clean
- Unit-test coverage for Patch 2 (cross-org event analytics path) — Codie picking up during review
- Smoke test for Patch 1 (re-import a codex-app-server agent export, verify runtime survives)

## Context

Workflow-inversion: usually Codie writes and Collie reviews, but Codie's queue was 9 deep and these patches were small + clear, so Collie wrote them per David/Dane greenlight. Codie reviews this PR.

Generated with Claude Code (collie + codex-rescue collaboration).


grandamenium:
**Sentinel QA — PR #436**

**Verdict: PASS** | CI 3/3 green

Reviewed both patches.

**Patch 1 (import-agent runtime):** Fallback chain importedConfig?.runtime || manifest?.runtime || 'claude-code' is correct. Preserves the agent's own config first, manifest second, safe default third. Fixes silent fall-back to claude-code for codex-app-server and hermes agents on import.

**Patch 2 (cross-org task completion analytics):** Regex /[\/]orgs[\/](?<org>[^\/]+)[\/]tasks[\/]/ correctly extracts the task's org from its file path. Flat/test-harness layout (no /orgs/ in path) returns empty fileOrg and keeps original paths unchanged — the exemption is right. No impact on single-org installs.

**Gap:** PR description notes unit-test coverage for Patch 2 is pending (Codie picking up). Not in this PR. The logic is simple and the flat-layout fallback is safe, but a cross-org event routing test would be good hygiene as a follow-up.

Both patches are clean P1/P2 fixes with no unrelated changes. Ready to merge.

Filed by Sentinel (QA, dual mandate).

grandamenium (approved):
## QA Review — PASS

Both changes are correct and match the independent PR #437 which fixes the same two issues plus the codex-app-server thread resume fix.

- `src/bus/task.ts`: cross-org task completion analytics path fix — correct.
- `src/cli/import-agent.ts`: runtime field in import manifest — correct.

**Merge note:** PR #437 (davidhunter) contains these exact same changes plus the codex-pty fix. Whichever merges first makes the other redundant. If #437 merges, this PR's diff will go to zero. Suggest closing this in favor of #437 to avoid duplicate diffs, unless there's a reason to prefer the split approach.

Issue #433, opened 2026-05-14
`shouldContinue()` uses Claude JSONL signal for codex-app-server runtime, causing deterministic restart-crash for migrated agents


grandamenium:
## Summary

Two related bugs in the agent restart path cause a deterministic crash loop for any `codex-app-server` agent that previously ran on the Claude runtime.

## Location

- `src/daemon/agent-process.ts:514-552` (`shouldContinue()`)
- `src/pty/codex-app-server-pty.ts:477-521` (`startOrResumeThread`)

## Bug 1 — Wrong continue signal for non-Claude runtimes

`shouldContinue()` checks `~/.claude/projects/<launch-dir-with-dashes>/*.jsonl` to decide fresh vs continue mode. This is Claude Code-specific and runs for **all** runtimes including `codex-app-server`. Codex agents have their own thread state file (`<state-dir>/codex-app-server-thread.json`) which is the correct signal for that runtime.

## Bug 2 — Uncaught JSON-RPC error in second resume path

The persisted-resume call at `codex-app-server-pty.ts:481-491` has try/catch. The fallback via `findLatestThreadForCwd` at lines 497-509 does **not**. If `thread/resume` times out on the second attempt, the error propagates uncaught, the daemon classifies it as a crash, and backoff restart begins — creating a loop.

## Reproduction (deterministic)

1. Take any `codex-app-server` agent that previously ran on the Claude runtime (leaving a JSONL in the Claude projects dir)
2. Restart the daemon
3. Agent enters continue mode (wrong signal)
4. Tries `thread/resume` against a persisted threadId
5. Codex backend has evicted the thread
6. JSON-RPC timeout propagates uncaught
7. Daemon classifies as crash, backoff restart, loop

## Witnessed in Production

**2026-05-14T17:50Z** post-soak-swap (HEAD 23684f93):

- `codex-agent` (testorg, migrated from Claude runtime to codex-app-server): stale May 9 JSONL at `~/.claude/projects/-Users-cortextos-cortextos-orgs-testorg-agents-codex-agent/` forced continue mode. Crashed 2x with `Error: JSON-RPC request timed out: thread/resume` before stabilizing.
- `codex-worker` (lifeos, native codex-app-server, no Claude project dir): went straight to fresh mode via `thread/start`, clean boot.

## Workaround (operational)

Remove the stale JSONL for affected codex agents:
```bash
[...]

PR #437, opened 2026-05-14
fix(codex-pty): resume persisted codex thread on daemon restart

noogalabs:
## Summary
- resume persisted codex app-server thread state regardless of start mode
- keep  fallback gated to continue mode
- add unit coverage for persisted-state resume in fresh mode

## Why
Codex agents can restart with mode=fresh because upstream continue detection is Claude-history based; without this, persisted codex thread state is ignored and continuity is lost after restart.

## Validation
- 
 RUN  v4.1.2 /Users/davidhunter/cortextos

 Test Files  1 passed (1)
      Tests  63 passed (63)
   Start at  15:40:36
   Duration  172ms (transform 53ms, setup 0ms, import 70ms, tests 11ms, environment 0ms)
- 
 RUN  v4.1.2 /Users/davidhunter/cortextos

 Test Files  1 passed (1)
      Tests  40 passed (40)
   Start at  15:40:37
   Duration  183ms (transform 36ms, setup 0ms, import 49ms, tests 52ms, environment 0ms)

### DIFF
--- a/src/bus/task.ts
+++ b/src/bus/task.ts
@@ -491,8 +491,14 @@ export function completeTask(
       // findTaskFile, but the caller's `paths.analyticsDir` is scoped to
       // the caller's org. Rewrite the analytics path to the task's actual
       // org so dashboards/metrics see the completion under the right tree.
-      const eventPaths: BusPaths = taskOrg
-        ? { ...paths, analyticsDir: join(paths.ctxRoot, 'orgs', taskOrg, 'analytics') }
+      // Only rewrite analyticsDir when the resolved task path is in the
+      // nested cross-org layout: <ctxRoot>/orgs/<org>/tasks/<taskId>.json.
+      // Flat/single-org test harnesses use <ctxRoot>/tasks + <ctxRoot>/analytics
+      // and should keep the caller-provided analyticsDir unchanged.
+      const pathOrgMatch = filePath.match(/[\\/]orgs[\\/](?<org>[^\\/]+)[\\/]tasks[\\/]/);
+      const fileOrg = pathOrgMatch?.groups?.org || '';
+      const eventPaths: BusPaths = fileOrg
+        ? { ...paths, analyticsDir: join(paths.ctxRoot, 'orgs', fileOrg, 'analytics') }
         : paths;
       logEvent(eventPaths, assignee, taskOrg, 'task', 'task_completed', 'info', {
         task_id: taskId,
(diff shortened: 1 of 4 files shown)


grandamenium:
**Sentinel QA — #433 coverage review**

This PR addresses the primary crash scenario from #433 and is a solid improvement. Coverage notes:

**Bug 1 (shouldContinue() reads Claude JSONL for all runtimes):** Mitigated via workaround — persisted thread state is now checked before the `mode` gate, so codex agents resume correctly even when incorrectly forced into fresh mode. The root cause in `shouldContinue()` is not fixed but the crash is avoided. Acceptable for now.

**Bug 2 (uncaught JSON-RPC in findLatestThreadForCwd fallback):** *Still open.* The `findLatestThreadForCwd` path at lines ~497-509 does not have try/catch. If `thread/resume` times out on this path, it still propagates uncaught and the daemon will classify it as a crash. This path is now only hit in `continue` mode with no persisted state, so it is less likely to be triggered — but it is not fully safe.

**Tests:** New unit coverage for fresh-mode persisted resume is correct and sufficient for the change made.

**Verdict:** PARTIAL — merging this is a net improvement and closes the witnessed crash scenario from 2026-05-14T17:50Z. Recommend a follow-up to wrap `findLatestThreadForCwd` try/catch before calling this fully closed. Will update #433 accordingly.

Filed by Sentinel (QA, dual mandate).

grandamenium (approved):
## QA Review — PASS

**CI:** Build + Dashboard + Unit Tests all green.

### Main fix: `src/pty/codex-app-server-pty.ts`
Correct. Lifts persisted-thread resume above the `mode === 'continue'` guard so fresh-mode restarts also pick up serialized thread state. The old guard was too conservative — it caused continuity loss when `shouldContinue` returned false (e.g. due to stale Claude JSONL from a previous runtime). Catch-path behavior is preserved: resume failure falls through to `findLatestThreadForCwd` (continue mode) or starts fresh (fresh mode). Test coverage added for the new code path.

### Bundled fixes
- `src/bus/task.ts` (+14/-1): Cross-org task completion now writes analytics to the task's actual org directory, not the caller's org. Regex path match is safe. Filing a separate issue (#) for tracking since this bug is independent of the codex-pty fix.
- `src/cli/import-agent.ts` (+2/-0): `runtime` field threaded through import manifest. Correct, small.

### Ready to merge
No changes requested. Awaiting James sign-off before merge (per policy).

PR #440, opened 2026-05-15
feat(crons): fresh-session dispatch with PROTECTED_CRONS guard (Hermes Pattern 3)

mofoeden-star:
## Summary
Pattern 3 of the Hermes-inspired refactor: scheduled crons can opt into running as a FRESH `claude --new` session instead of inheriting the live agent's accumulated PTY history via `--continue`.

Eliminates context accumulation from scheduled chores → defends against instruction-following degradation.

**Stacked on #439 (Hermes Phase 1+2).**

## Design (codex-g impl after 2 ops-g/codex-g adversarial review passes — 20 findings incorporated)

- **`fresh_session: true` flag** on `CronDefinition` (opt-in, default false — preserves existing behavior).
- **`dispatchCron()` shared path**: scheduler + test-fire both route through it (eliminates divergence).
- **`buildAgentRuntimeEnv()` parity**: fresh sessions get the same env as live PTY (no env-skew bugs).
- **`timedOut` flag** on the spawn wrapper: timeout + graceful exit-0 correctly rejects.
- **`PROTECTED_CRONS` per-cron exemption list** in shared module:
  - `smart-g/*` blocked (analyst depends on session-level pattern recognition).
  - `top-g/morning-review`, `top-g/evening-review`, `top-g/weekly-review` blocked (need conversational continuity from yesterday).
  - `top-g/heartbeat` and `top-g/check-approvals` ALLOWED (stateless polls, re-derive from disk).
- **Fail-closed for unsupported runtimes**: if a runtime can't spawn fresh sessions, error rather than silently fall back.
- **`skill_file` → `--append-system-prompt`**: fresh session loads the relevant skill body as system prompt.
- **Cooldown applied on success only**: prevents rapid-retry storm on transient failures.

## Test plan
- [x] **28/28** fresh-session unit tests pass
- [x] **29/29** ipc-fire-cron tests pass
- [x] **35/35** agent-manager tests pass
- [x] **30/30** bus-crons tests pass
- [x] **328/328** daemon unit tests pass
- [x] 4 pre-existing integration failures NOT introduced by this change (confirmed same on baseline)
- [x] PROTECTED_CRONS list verified: smart-g/* and top-g review crons blocked; top-g heartbeat + check-approvals allowed
- [ ] CI passes
- [ ] Limited-blast rollout: dev-g heartbeat first → 1-2 cycles → verify via smart-g baseline → expand to other specialist crons → enable top-g/heartbeat + top-g/check-approvals last

## Ratchet protocol
[...]

PR #441, opened 2026-05-15
fix(crons): restore fresh session observability (Pattern 3 follow-up)

mofoeden-star:
## Summary
Follow-up patch to #440 (Pattern 3 fresh-session). Smart-g's post-flip audit on dev-g caught 3 observability regressions that block expansion to other crons. This PR restores parity with the legacy PTY-inject path.

**Stacked on #440.** Merge after #440 lands.

## Regressions fixed

**R1 — fresh fires now write to cron-execution.log**
Legacy fires write status/duration/error rows. Fresh fires were silent → smart-g's cron-watchdog would false-positive every fresh fire as a silent skip. Now `fireCronFreshSession` calls `writeExecLog` from close/error/pre-spawn handlers (success, non-zero exit, timeout, spawn error all covered).

**R2 — fresh fires now update crons.json `last_fired_at` + `fire_count`**
Uses actual `firedAt` timestamp (not `new Date()`) to prevent interval drift. Scheduler disk write gated on `!cron.fresh_session` (in-memory `nextFireAt` advancement stays unconditional — prevents refire loop, **critical correctness fix from codex-g adversarial review**).

**R3 — TZ env propagation in fresh sessions**
Else-if branch was dropping `TZ=process.env.TZ`. Memory entries from fresh fires had 3h30m timestamp discrepancy (Bangkok vs UTC). `CTX_CRON_FIRED_AT` also injected into fresh session env so skill templates can use a deterministic timestamp. R3 root cause is suspected (skill template may also have a bug); mitigation in place, template follow-on may be needed.

## Scheduler change
`maxAttempts=1` for fresh sessions (no retry blocking — fresh sessions are spawned isolated, retries don't make sense the same way). All PTY-inject paths unchanged.

## Test plan
- [x] **119/119** tests pass (existing fresh-session suite + new observability coverage)
- [x] codex-g adversarial review: 1 critical + 3 high findings, all incorporated
- [x] ops-g code review: APPROVED
- [x] Build clean
- [ ] CI passes
- [ ] Local validation: rebuild daemon, fire dev-g/heartbeat, verify cron-execution.log has new row + crons.json `fire_count` increments + memory timestamp matches actual fire time
- [ ] smart-g re-validates against original baseline (pre-flip vs post-patch fresh fire)

## Authorship
Spec: ops-g (Sonnet 4.6). Impl: codex-g (gpt-5.5 high). Cross-review: ops-g + codex-g via REVIEW: prefix. Final approval: top-g (Opus 4.7).

🤖 Generated with cortextOS Hermes Pattern 3 follow-up

### DIFF
--- a/src/cli/bus.ts
+++ b/src/cli/bus.ts
@@ -1,7 +1,7 @@
 import { Command } from 'commander';
 import { spawnSync, execFileSync } from 'child_process';
-import { existsSync, readFileSync } from 'fs';
-import { join } from 'path';
+import { existsSync, readFileSync, readdirSync } from 'fs';
+import { isAbsolute, join } from 'path';
 import { sendMessage, checkInbox, ackInbox } from '../bus/message.js';
 import { validateAgentName } from '../utils/validate.js';
 import { createTask, updateTask, completeTask, claimTask, readTaskAudit, checkTaskDependencies, compactTasks, listTasks, checkStaleTasks, archiveTasks, checkHumanTasks } from '../bus/task.js';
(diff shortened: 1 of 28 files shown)


grandamenium (changes requested):

PR #443, opened 2026-05-15
fix(crons): complete fresh session validation patch (Pattern 3 follow-up #2)

mofoeden-star:
## Summary
Second follow-up patch to PR #440 (Pattern 3 fresh-session). Smart-g's re-validation after PR #441 surfaced 3 remaining issues blocking expansion to other crons. This PR closes them.

**Stacked on #441.** Merge after #441 lands.

## Fixes

**Fix 1 — crashes.log false-positive (real regression)**
Every fresh-session subprocess exit was producing a phantom crashes.log entry (reason=none) because the daemon's crash-detection hook treated isolated child exits as parent agent crashes. `CTX_FRESH_SESSION_CRON` early-return now sits as the second line of `hook-crash-alert.ts main()` — before mkdirSync, crash counters, crashes.log, Telegram, everything. Hook exported for testability; VITEST guard prevents auto-execute on import.

**Fix 2 — memory write missing on fresh fires (Sean greenlit Option A)**
Fresh fires weren't executing the full HEARTBEAT.md operational checklist (mandatory Step 5: write daily memory). Root cause: `skill_file` pointed at `.claude/skills/heartbeat/SKILL.md` (skill definition, ~150 lines), not the agent-root `HEARTBEAT.md` (full checklist). Solution: new `protocol_file` field on `CronDefinition`. Fresh sessions now load both `skill_file` AND `protocol_file` via single `--append-system-prompt` (`systemPromptParts`). Wired through daemon dispatch + IPC validation (`validateFreshSessionFields`, `handleAddCron`, `handleUpdateCron`) + CLI (`add-cron --protocol-file`, `update-cron --protocol-file`). `dev-g/heartbeat` cron updated with `protocol_file: HEARTBEAT.md`.

**Fix 3 — last_fire_attempted_at not updated**
Legacy path sets `last_fire_attempted_at` BEFORE onFire (used to detect mid-fire crashes — see cron-scheduler.ts:387-388). Fresh path was skipping this. Now written after pre-spawn validation, before Promise constructor — covers IPC test-fire path; belt-and-suspenders for scheduler path.

## Test plan
- [x] **131 new tests** across 4 files, all green
- [x] Full suite: 0 new failures (27 pre-existing doc test failures unchanged)
- [x] codex-g adversarial review pass complete
- [x] ops-g code review: APPROVED
- [ ] CI passes
- [ ] Local validation post-rebuild: trigger fresh fire on dev-g/heartbeat → verify crashes.log NOT polluted, memory entry written with correct timestamp, last_fire_attempted_at updated
- [ ] smart-g re-validates against original baseline → green-light to remove WATCHDOG_SKIP_FRESH_SESSION_TESTBED env var → expand to other specialist crons

## Authorship
Spec: ops-g (Sonnet 4.6). Impl: codex-g (gpt-5.5 high). Cross-review: ops-g + codex-g via REVIEW: prefix. Final approval: top-g (Opus 4.7).

🤖 Generated with cortextOS Hermes Pattern 3 follow-up #2

### DIFF
--- a/src/cli/bus.ts
+++ b/src/cli/bus.ts
@@ -1865,8 +1865,8 @@ function validateFreshSessionCliOptions(
   agent: string,
   cronName: string,
   frameworkRoot: string,
-  opts: { freshSession?: boolean; skillFile?: string; freshSessionTimeout?: string },
-): { freshSession?: boolean; skillFile?: string; freshSessionTimeoutMs?: number } {
+  opts: { freshSession?: boolean; skillFile?: string; protocolFile?: string; freshSessionTimeout?: string },
+): { freshSession?: boolean; skillFile?: string; protocolFile?: string; freshSessionTimeoutMs?: number } {
   if (opts.skillFile !== undefined) {
     if (!opts.skillFile.trim()) {
       console.error('Error: --skill-file must be a non-empty relative path.');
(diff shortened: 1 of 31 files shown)


grandamenium (changes requested):

PR #446, opened 2026-05-15
fix(telegram+daemon): suppress local_file in photo + auto-recover image-poison crashes

storypixel:
## Summary

Two coupled fixes for the same recurring failure mode: cortextOS agents silently crash-loop when their conversation history contains a poisoned image attachment.

**Source fix** (commit 1 — original PR scope): suppress the `local_file:` path from photo injections in `FastChecker.formatTelegramPhotoMessage`. Without this, Claude Code reads the file Telegram delivered, base64-encodes it for Anthropic with a media_type that doesn't match the bytes, and Anthropic rejects with `API Error: 400 ... image format image/<fmt> not supported`. The agent exits clean (code 0) and the bad reference stays in conversation history, so every `--continue` restart re-crashes.

**Retroactive fix** (commit 2 — new in this update): detect that exact 400 signature in the agent's recent stdout after exit, auto-write `.force-fresh`, and respawn without counting against `max_crashes_per_day`. Closes the gap where agents poisoned BEFORE the source fix shipped will keep crash-looping until they hit the daily limit and the daemon halts them.

## Why both

Real failure mode observed in production:
1. User sends a screenshot via Telegram → agent receives it → injection includes `local_file: /path/to/screenshot.jpg`
2. Claude Code auto-reads the file, base64-encodes it for Anthropic
3. Media-type stamping bug somewhere in the pipeline labels it `image/png` when the bytes don't match
4. Anthropic returns HTTP 400 `image format image/png not supported`
5. Agent exits code 0 (clean)
6. Daemon counts as a crash, restarts with `--continue`
7. Saved conversation history still contains the poisoned base64 → instant re-crash
8. After 3-4 cycles, max_crashes_per_day trips and daemon halts the agent

Source fix alone (commit 1) prevents NEW poisoning. It doesn't help an agent that already has a bad image in saved history — that agent has to be manually `.force-fresh`'d, which an operator can only do if they know to look for the specific 400 signature in logs.

Retroactive fix (commit 2) makes the daemon self-healing for this specific failure mode.

### Commit 1 — `fast-checker.ts` (original)
- `formatTelegramPhotoMessage`: replace `local_file: ${imagePath}` with `[image attached — local path suppressed to prevent claude-code auto-attach crash; vision pre-call not yet wired]`
- Note added in source: imagePath stays in the signature for a future vision-side-channel that produces a text description, currently unused

### Commit 2 — `agent-process.ts` (new)
- `handleExit`: read last 16KB of stdout.log via new `tailStdoutLog` helper
- New `detectImagePoisonCrash`: returns true when recent output contains `API Error: 400` + `image.source.base64`, OR the literal `image format image/<fmt> not supported`
- New `armForceFresh`: writes `.force-fresh` marker to state dir
- New path in `handleExit`: when exit code 0 + image-poison signature detected, log `IMAGE_POISON_RECOVERY` and respawn after 5s WITHOUT incrementing `crashCount`
- `appendCrashToRestartsLog` accepts new `IMAGE_POISON_RECOVERY` kind, formatted distinctly so operators can grep for it in audit logs

## Test plan

- [x] Manual: send a problematic screenshot to an agent → confirm new fast-checker injection omits local_file path
- [x] Manual: pre-poison an agent's conversation with a bad image, restart → confirm handleExit detects, arms force-fresh, respawns fresh (no `--continue`), crash count unchanged
- [x] Manual: trigger an unrelated crash (e.g. PTY error) → confirm normal CRASH path still increments counter
- [x] Unit: existing `fast-checker.test.ts` updated for the new format

Tested in production across 9 cortextOS agents that had been collectively hitting this bug 4-6 times per day.

### DIFF
--- a/src/daemon/agent-process.ts
+++ b/src/daemon/agent-process.ts
@@ -376,7 +376,80 @@ export class AgentProcess {
 
   // --- Private methods ---
 
+  /**
+   * Read the tail of this agent's stdout.log without loading the whole file.
+   * Used by handleExit() to inspect recent output for known-crash signatures
+   * (e.g. the image-poison API 400 pattern) so it can decide whether the
+   * exit is a real crash or a recoverable upstream artifact.
+   *
+   * Returns an empty string if the log doesn't exist or can't be read.
+   */
+  private tailStdoutLog(maxBytes: number): string {
+    const logPath = join(this.env.ctxRoot, 'logs', this.name, 'stdout.log');
+    try {
+      if (!existsSync(logPath)) return '';
+      const stats = statSync(logPath);
+      const start = Math.max(0, stats.size - maxBytes);
+      const len = stats.size - start;
+      // Synchronous read of the tail; small and bounded so the cost is fine
+      // even in the exit handler.
+      const fd = require('fs').openSync(logPath, 'r');
+      try {
+        const buf = Buffer.alloc(len);
+        const read = require('fs').readSync(fd, buf, 0, len, start);
+        return buf.toString('utf-8', 0, read);
+      } finally {
+        require('fs').closeSync(fd);
+      }
+    } catch {
+      return '';
+    }
+  }
+
+  /**
+   * Match the API 400 image-poison signature in recent stdout.
+   *
+   * Two variants observed in Anthropic's Messages API responses:
+   *   `API Error: 400 messages.N.content.M.image.source.base64.data: Image format image/<fmt> not supported`
+   *   `API Error: 400 ... image.source.base64.data: ...`
+   *
+   * Matching the prefix `image.source.base64` is robust to wording changes
+   * in Anthropic's error string; matching `image format image/<fmt>` is the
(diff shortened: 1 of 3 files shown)


grandamenium (approved):
## QA Review — PASS

Correct diagnosis and fix. Claude Code auto-attaches local image paths found in user input, causing a 400 API error that permanently wedges the session. Suppressing the path from the message and replacing with an explanatory text note is the right approach. The `void imagePath` annotation prevents unused-variable warnings. Test updated to match. CI green.

storypixel:
Approved + clean. Ready when you are.

grandamenium:
Resolved as a split: the image-poison auto-recover half landed via #552 (rebased onto current main + sandbox-validated, @storypixel co-author credit preserved). The photo local_file-suppression half is superseded by the Track-2 byte-sniff mime fix (which addresses the root cause instead of suppressing the path) and is intentionally not carried forward. Closing — thanks @storypixel.

PR #455, opened 2026-05-16
fix(cron-scheduler): use created_at as referenceMs floor for pre-first-fire schedule drift

mofoeden-star:
## Root Cause

A cron that has never fired has all-null timestamps (`last_fired_at`, `last_fire_attempted_at`, `cron-state.json.last_fire`). The scheduler's `referenceMs` computation fell back to `now`, so `next_fire_at = now + interval` on every daemon restart — the cron drifted forward indefinitely and could never reach its first fire slot.

## Evidence

Reproduces on any fleet cron created with a multi-hour interval:
- Cron created at T=0, interval=4h
- Daemon restarts at T=2h (no fire yet): `referenceMs=now` → `next_fire_at=now+4h=T+6h`
- Daemon restarts again at T=5h: `referenceMs=now` → `next_fire_at=now+4h=T+9h`
- Cron never fires

## Fix

Use `created_at` as the fallback `referenceMs` **only** when the candidates pool (fire history) is empty:

```typescript
const referenceMs = candidates.length > 0
  ? Math.max(...candidates)
  : (def.created_at ? new Date(def.created_at).getTime() : now);
```

`created_at` must **not** enter the `max()` pool — doing so would suppress catch-up for crons with old `last_fired_at` but recent `created_at` (e.g. a just-restarted daemon). The pool contains actual fire evidence; `created_at` is a reference floor only when that evidence is absent.

This commit also removes the stale `if (!cron.fresh_session)` guard around the post-fire `updateCron` call. That guard is the subject of PR #453 (already open against main); it is re-applied here so this branch builds correctly without depending on PR #453 merging first.

## Test Plan

- [x] `npm run build` — TypeScript compiles cleanly
- [x] `npm test tests/unit/daemon/cron-scheduler.test.ts` — 34/34 pass
- [x] Updated existing `fresh_session` test: now asserts `last_fired_at + fire_count` ARE persisted (was asserting the opposite — the bug)
- [x] Added new test: cron created 2h ago with 4h schedule → `nextFireAt ≈ now+2h` (not `now+4h`)
- [x] All existing catch-up, reload, and retry tests still pass (no regression from created_at placement)

## Related

- PR #453: `fresh_session` persistence fix (guard removal already open; re-applied here for standalone build)

PR #456, opened 2026-05-16
feat(kb): route queryKnowledgeBase + ingestKnowledgeBase through gbrain

mofoeden-star:
## Root Cause / Motivation

mmrag.py (ChromaDB + Gemini Embeddings) is the legacy KB backend. gbrain is now the primary knowledge store (Supabase Postgres + pgvector, hybrid RRF search). The two systems need to coexist until everything migrates; agents should hit gbrain by default without any configuration change on machines where it is already installed.

## What Changed

**`src/bus/knowledge-base.ts`** — gbrain fast-path behind a feature flag:

When `/root/.gbrain/config.json` exists (gbrain initialized) AND `GBRAIN_KB_BACKEND` is not `false`:

- **`queryKnowledgeBase()`** → spawns `gbrain query <q> --limit <topK>`, parses scored text output (`[score] slug -- preview`), then fetches full page content per result via `gbrain get <slug>`. Returns `KBQueryResponse` in the same shape callers already expect.
- **`ingestKnowledgeBase()`** → derives a slug from each file basename, pipes file content to `gbrain put <slug>`.

Rollback: `GBRAIN_KB_BACKEND=false` in the agent's env, or remove `/root/.gbrain/config.json`. mmrag path is untouched.

**`tests/unit/bus/knowledge-base.test.ts`** — test adjustments:
- `mockMissingKbConfig` and `mockConfiguredKb` updated to return `false` for `.gbrain/config.json` paths, keeping existing mmrag tests on the legacy path.
- New `mockConfiguredKbWithGbrain` helper for gbrain tests.
- 3 new tests: query happy path, query error handling, ingest put-per-file.

## Test Plan

- [x] `npm run build` — TypeScript compiles cleanly
- [x] `npm test tests/unit/bus/knowledge-base.test.ts` — 8/8 pass (5 existing + 3 new)
- [x] Full suite — pre-existing failures unchanged (27 in phase3-docs, sprint1-templates, phase5); 0 new failures
- [x] Smoke test on live fleet: `cortextos bus kb-query "what is cortextos" --org 1evo` → 5 results with scores and full content via gbrain path ✓

## Notes

- N+1 shell spawns per query (1 `gbrain query` + N `gbrain get` for topK results). Acceptable for topK=5 (~3s) but worth watching if topK is raised.
- gbrain CLI injects `[ai.gateway]` warnings to stderr — these are suppressed via `stdio: ['pipe','pipe','ignore']` in the execFileSync calls, keeping stdout parsing clean.

### DIFF
--- a/src/bus/knowledge-base.ts
+++ b/src/bus/knowledge-base.ts
@@ -8,7 +8,130 @@ import { normalizeOrgName } from '../utils/org.js';
 /**
  * Knowledge base integration — calls mmrag.py directly (cross-platform,
  * no bash dependency).  Previously wrapped kb-*.sh bash scripts.
+ *
+ * gbrain fast-path: when /root/.gbrain/config.json exists (gbrain initialized),
+ * queryKnowledgeBase and ingestKnowledgeBase route through gbrain instead of
+ * mmrag.py.  Set GBRAIN_KB_BACKEND=false to force the legacy path.
  */
+
+const GBRAIN_CLI = '/usr/local/bin/gbrain';
+
+/**
+ * True when gbrain is installed, initialized, and not explicitly disabled.
+ * Checks binary + config file existence so cold machines without gbrain
+ * fall back to mmrag transparently.
+ */
+function gbrainAvailable(): boolean {
+  return (
+    existsSync(GBRAIN_CLI) &&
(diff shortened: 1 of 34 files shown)


grandamenium (changes requested):

PR #464, opened 2026-05-16
fix(bus): stop truncating task IDs in list-tasks display

asachs01:
## Summary

`bus list-tasks` chopped task IDs to 26 chars via `substring(0, 26).padEnd(26)`. Task IDs were bumped from `task_<13ms>_<3-digit>` (22 chars) to `task_<13ms>_<8-digit>` (27 chars) in a prior `createTask` fix to avoid same-millisecond random-suffix collisions (see `task.ts:42` comment). The display column wasn't bumped along with the data layer, so **the last digit of every new-format ID was silently dropped** in the table output.

## Symptom

Operators copy-pasted the visible 26-char ID into `update-task` / `complete-task`, which then correctly reported `'Task <id> not found in any org under ...'` because `findTaskFile` looks for the on-disk filename — which has the full 27-char ID.

Two confirmed instances (verified on disk):

| Display (truncated) | On-disk (actual) |
|---|---|
| `task_1778933649360_7150617` | `task_1778933649360_71506170.json` |
| `task_1778936188789_8921793` | `task_1778936188789_89217936.json` |

## Fix

`src/cli/bus.ts:384` — stop truncating task IDs in display. Apply the same principle to assignee (drop the substring; keep `padEnd` as min-width).

```diff
-    const id = t.id.substring(0, 26).padEnd(26);
-    const assignee = (t.assigned_to || '-').substring(0, 16).padEnd(17);
+    const id = t.id.padEnd(28);
+    const assignee = (t.assigned_to || '-').padEnd(17);
```

Plus matching header padding so column alignment stays clean (header `ID` column gets +2 spaces).

## Principle

Never `substring` an identifier in display. The data layer is the source of truth; the display must preserve identifier precision so copy-paste from the table cannot produce a non-existent ID. ID column pad bumped to 28 (current max 27 + 1 headroom) so the next ID-format change won't reintroduce this class of bug. Assignee column drops the substring but keeps the existing 17-char min-width since no current assignee name exceeds 16.

## Scope

- **No data drift.** The store was never inconsistent — purely a display-layer bug. `create-task` and `list-tasks` were always writing/reading the full ID; only the column rendering chopped it.
- **No mutation-layer changes.** `findTaskFile` was correct all along; it just got handed truncated IDs by humans copy-pasting from the broken display.

## Verification

- `npm run build` clean.
- `npm test` — same pre-existing failure count on `main` vs this branch (11 files fail, 1 test fail, 1538 pass, 48 skipped). Failures unrelated to this change.
- Manual visual check:

```
  Status  Pri  ID                          Assignee         Title
  ---------------------------------------------------------------
  ●       🔵   task_1778937502052_05411243 murph            Ship Entra ID MCP vendor-config to mcp-gateway mai
  ●       🔵   task_1778699061604_275      murph            Pair with forge on Azure + Entra ID MCP wiring int
```

Both new-format (27-char) and old-format (22-char) IDs render in full, column alignment intact.

## Discovery context

Surfaced by boss after dev hit "Task X not found in any org" on two old tasks (`task_1778933649360_*` and `task_1778936188789_*`). The truncated `~/.cortextos/de` in boss's bug report title was boss's own write-side truncation when transcribing — the actual error log shows the full path. The real bug was the display, not the path resolution.

## Test plan

- [x] `npm run build` clean
- [x] `npm test` — no new failures
- [x] Manual visual: both new-format and old-format IDs render in full
- [ ] Maintainer review + merge


thehoff:
Heads-up on a duplicate: I independently opened #626 for the same bug (#587) before spotting this PR — apologies for the overlap, **this PR predates mine**.

Same core fix (stop truncating, widen the id column to 28). #626 adds two things that might be worth folding in here, whichever the maintainer prefers:

- a shared `ID_W` constant driving **both** the header column and the row, so the two widths can't drift the way the hardcoded 26 did vs the 27-char id;
- an integration regression test (create a task → read the full id back out of the `list-tasks` **table** → confirm `complete-task` accepts it).

No strong preference on which lands — flagging so there's only one merge for this. cc @asachs01

PR #465, opened 2026-05-16
fix(daemon): write .session-refresh marker so SessionEnd hook classifies planned refresh

saurav-yirifi:
## Summary

- `hook-crash-alert.ts` reads `.session-refresh` to classify a PTY exit as a planned session refresh; without the marker, the SessionEnd hook defaults to `crash` and writes a spurious entry to `crashes.log` on every daemon-internal session refresh.
- `AgentProcess.sessionRefresh()` was missing the writer side — `stop() + start()` ran with no marker write.
- Fix: write the marker at the top of `sessionRefresh()` before `stop()`, mirroring `src/bus/system.ts:65` (hardRestart).

## Postmortem — the asymmetry was the bug

The marker type `.session-refresh` was wired into the **reader** (`src/hooks/hook-crash-alert.ts:185`) but the **writer** was never implemented. Daemon-driven session refresh executed correctly at the process level, but the SessionEnd hook had no way to distinguish a planned refresh from a crash, so it defaulted to `crash` (line 180) and emitted a misleading entry. Reader-without-writer is the inverse of the more common primitive-without-callers trap — both fail the same way: each half passes its own unit tests, and the gap only shows up in live operation.

## Observed in the wild

`boss/crashes.log` at `2026-05-16T06:52:38Z`: `type=crash reason=none last_task=online` — PM2 logs from the same window show a clean planned-restart sequence. The daemon did the right thing; the hook misclassified.

## Test plan

- [x] `npx tsc --noEmit` — clean
- [x] `npx vitest run tests/unit/daemon/agent-process.test.ts` — 34/34 pass including new ordering test
- [x] `npm test` — 2273 passed / 0 failed
- [x] `npm run build` — clean
- [ ] After merge: verify the next daemon-driven session refresh (6h `session-refresh` cron) no longer produces a `crashes.log` entry

## Scope

- Single-phase fix. `code-evaluator` ran on the commit — LGTM with one nit (clarifying test comment), addressed.
- Explicitly out of scope per dispatch: `hook-crash-alert.ts` changes, marker file format redesign, historical `crashes.log` cleanup.


saurav-yirifi:
wrong base; re-targeting to fork main (saurav-yirifi/sb-cortextos-fork) — that's the deploy path for our running fleet. Will cherry-pick to upstream as a separate clean PR after production validation.

PR #475, opened 2026-05-17
fix(daemon): make spawn-codex cron dispatch async to unblock event loop

revopsglobal:
## Summary
- `spawnCodex()` in `src/bus/spawn-codex.ts` used `spawnSync` which blocked the daemon event loop for 50–116s per fire (every 30min on kanban-sync cron)
- Added `spawnCodexAsync()` using non-blocking `child_process.spawn`; sync path preserved for CLI compatibility
- `dispatchCronFire` made async, `agent-manager.ts` awaits it

## Root Cause
`spawnSync` on the Node.js main thread blocks all timers, IPC, and watchdog heartbeats — causing false watchdog restarts and IPC timeouts that appeared to be separate bugs.

## Test Plan
- [ ] Unit tests updated to mock `spawnCodexImpl` and async-ify `dispatchCronFire` calls
- [ ] `npm test` passes
- [ ] `npm run build` compiles cleanly


revopsglobal:
Wrong base repo — this fix targets RevOps-Global-GIT fork where cron-fire-dispatch + spawn-codex architecture is in use.

PR #480, opened 2026-05-17
fix(daemon): write .daemon-stop markers synchronously in SIGINT/SIGTERM handler

revopsglobal:
## Problem

PM2 sends SIGINT/SIGTERM to the daemon's **entire process group**. PTY children (Claude Code CLI sessions) receive the signal at the same instant as the daemon — before the Node.js event loop can execute the async `shutdown()` → `stopAll()` sequence that wrote `.daemon-stop` markers.

Without markers on disk at signal time, `agent-process.ts:handleExit()` mis-classifies every PTY exit as a crash → SESSION CONTINUATION injected per agent per daemon restart. Observed: 30-40 fake restarts/hour = measurable Claude API spend.

## Fix

- Add `writeAllDaemonStopMarkersSync()` to `AgentManager` — synchronous marker-write called in the signal handler before any async work.
- Call it in `handleSignal()` before the `shutdown()` async call.
- `stopAll()` retains its own marker-write loop as belt-and-suspenders.

## Test plan

- [ ] `npm run build` passes ✅
- [ ] After `pm2 restart cortextos-daemon`, no SESSION CONTINUATION flood — agents get clean `--continue` restarts
- [ ] No BUG-011 regression check firings in PM2 out log after daemon restart


revopsglobal:
Opened against wrong repo — closing. Correct PR will be in RevOps-Global-GIT/cortextos.

PR #483, opened 2026-05-18
fix(daemon): cron --continue race + heartbeat watchdog env

soletraderai:
## Summary

Two related production bugs that surface together during a Claude Code `--continue` restart of an agent process:

### Bug 1 — Cron prompt-injection race

`CronScheduler.fireWithRetry` exhausts after 4 attempts inside a ~21s window (`RETRY_DELAYS_MS = [1_000, 4_000, 16_000]`). During a `--continue` rehydration the child `AgentProcess` flips `status === 'running'` the instant the PTY spawns, but `AgentProcess.injectMessage` still no-ops:

```ts
// src/daemon/agent-process.ts
injectMessage(content: string): boolean {
  if (!this.pty || this.status !== 'running') {
    return false;
  }
  // …
}
```

The PTY is alive, but stdin is still streaming a multi-MB transcript replay before the agent is ready to read injected input. With ~21s of total retry budget, a slow rehydration silently drops the cron — the execution log records `failed` and no operator alert fires.

**Fix:** extend the retry shape to `1+2+4+8+16+32+64s` (8 attempts, ~127s window). Same exponential-backoff shape, longer ceiling. This is a stop-gap; the real fix is to drain queued cron firings against an explicit bootstrap-completion event from `AgentProcess`, which I'd suggest tracking as a follow-up.

### Bug 2 — Heartbeat watchdog rejects daemon agent name

`FastChecker` spawns a `cortextos bus update-heartbeat` every 15s via `execFile`. With no explicit `env` option the child inherits `process.env`, and the daemon process does **not** always have `CTX_AGENT_NAME` set — it can derive identity from `CTX_AGENT_DIR`. When `resolveEnv()` runs in the child:

```ts
// src/utils/env.ts
const agentName =
  overrides?.agentName ||
  process.env.CTX_AGENT_NAME ||
  envFile.CTX_AGENT_NAME ||
  basename(process.cwd());  // ← falls back to "MudBot" (capital M)
```

It falls through to `basename(daemon-cwd)` and produces a name like `MudBot` that `validateAgentName` (`/^[a-z0-9_-]+$/`) rejects. Result: `Invalid agent name 'MudBot'` spammed in the watchdog log every 15s, no heartbeat ever recorded.

There's no agent-side workaround — `update-heartbeat` has no `--agent` flag, and the daemon spawn ignores per-agent `.env` files. The fix has to live in the daemon.

**Fix:** explicitly pass `CTX_AGENT_NAME: agentName` in the `execFile` env override. One-line change at the call site; `FastChecker` already holds the canonical agent name.

## Evidence

Real-world impact captured against a fork running an agent in a parent dir named `MudBot`:

- `daily-pnl-summary` cron fired at `07:33 BNE` (after a `--continue` restart) but never delivered; execution log shows `failed` after 4 retries inside the 21s window.
- Daemon log shows `Heartbeat watchdog error: Invalid agent name 'MudBot'` repeating every 15s on the same agent. Agent reports `[CRON FIRED …]` but the heartbeat in the dashboard hasn't moved since the restart.

## Changes

**Code (2 files):**

- `src/daemon/cron-scheduler.ts` — `RETRY_DELAYS_MS = [1_000, 2_000, 4_000, 8_000, 16_000, 32_000, 64_000]`; rationale comment block; max-attempts comment and log-message templates updated from hardcoded `4` to `\${maxAttempts}`.
- `src/daemon/fast-checker.ts` — heartbeat-watchdog `execFile` call now passes `{ env: { ...process.env, CTX_AGENT_NAME: agentName } }`.

**Tests (5 files):**

- `tests/unit/daemon/cron-scheduler.test.ts` — retry count `3 → 7`, `failingFire` calls `4 → 8`, advance-time totals updated.
- `tests/unit/daemon/fast-checker.test.ts` — existing test updated for 4-arg `execFile` signature; new regression test asserts `options.env.CTX_AGENT_NAME === 'my-agent'` via a watchdog-call filter (so bootstrap-check tests stay independent).
- `tests/integration/phase5-failure-modes.test.ts` — \"first 3 then succeeds on 4\" → \"first 4 then attempt 5\" with `retried×4 + fired×1`; \"all 4 attempts\" → \"all 8\" with `deadCallCount=8`, advance `135s`.
- `tests/integration/phase1-backtesting.test.ts` — \"exhausts 4 attempts\" → \"exhausts 8 attempts\" (`retried.length=7`, `failed.attempt=8`); retry-timing test rewritten from `[1s,4s,16s]` to `[1,2,4,8,16,32,64]s` with array-based gap assertions.
- `tests/integration/phase5-e2e-simulation.test.ts` — comment block at scenario 5 updated for the new budget; tests 5c/5d updated for 8 attempts and the new spacing.

## Test plan

- [x] `npm run build` clean
- [x] `npm test` — 1545 passed / 1 failed; the failure is `dashboard/.../next/server` pre-existing on `upstream/main` (verified by running the same test on a clean checkout of `upstream/main` — same failure). 11 dashboard files fail to import `next/server` on both `main` and this branch.
- [x] Targeted unit tests: `cron-scheduler` + `fast-checker` 74/74 pass.
- [x] Targeted integration tests: phase1-backtesting, phase5-failure-modes, phase5-e2e-simulation retry/heartbeat assertions all pass.

## Follow-up (not in this PR)

Bug 1's real fix is event-driven, not time-window-driven: have `AgentProcess` emit a `bootstrap-complete` event after the rehydration stream finishes, and have `CronScheduler` queue firings against that event for newly-spawned agents instead of racing the PTY. Happy to open a follow-up issue if useful.

### DIFF
--- a/src/daemon/cron-scheduler.ts
+++ b/src/daemon/cron-scheduler.ts
@@ -167,7 +167,14 @@ function computeNextFireAt(cron: CronDefinition, referenceMs: number): number {
 // Retry helper
 // ---------------------------------------------------------------------------
 
-const RETRY_DELAYS_MS = [1_000, 4_000, 16_000];
+// Exponential back-off across 8 attempts (initial + 7 retries) for a cumulative
+// wait of ~127 s before giving up. The previous 4-attempt window (1 + 4 + 16 s
+// = 21 s) was reliably exhausted when a cron fired during a Claude Code
+// `--continue` restart: status flips to 'running' on PTY spawn but the session
+// is still rehydrating its transcript, so `injectMessage()` returns false until
+// bootstrap completes. Heavy transcripts (multi-MB) can take well over 21 s,
+// causing the prompt to be lost. 127 s covers observed worst-case rehydration.
+const RETRY_DELAYS_MS = [1_000, 2_000, 4_000, 8_000, 16_000, 32_000, 64_000];
 
 async function fireWithRetry(
   cron: CronDefinition,
@@ -175,7 +182,7 @@ async function fireWithRetry(
   onFire: (c: CronDefinition) => Promise<void> | void,
   logger: (msg: string) => void,
 ): Promise<boolean> {
-  const maxAttempts = RETRY_DELAYS_MS.length + 1; // 4 attempts total
+  const maxAttempts = RETRY_DELAYS_MS.length + 1; // 8 attempts total
   for (let attempt = 0; attempt < maxAttempts; attempt++) {
     const start = Date.now();
     try {
@@ -196,7 +203,7 @@ async function fireWithRetry(
         const delay = RETRY_DELAYS_MS[attempt];
         logger(
           `[cron-scheduler] onFire failed for "${cron.name}" ` +
-          `(attempt ${attempt + 1}/4, retrying in ${delay}ms): ${errMsg}`
+          `(attempt ${attempt + 1}/${maxAttempts}, retrying in ${delay}ms): ${errMsg}`
         );
         appendExecutionLog(agentName, {
           ts: new Date().toISOString(),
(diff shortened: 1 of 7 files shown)


soletraderai:
Closing — going with a local workaround instead. Thanks.

PR #487, opened 2026-05-18
fix(cron-migration): accept type:"cron" + schedule alias on auto-load

soletraderai:
## Summary

`CronEntry.type` previously only accepted `'recurring' | 'once' | 'disabled'`. Authors who wrote `type:"cron"` plus `schedule:"0 18 * * *"` in their agent `config.json` (a natural convention given the type name) watched the daemon silently skip the entry on boot — neither field name was recognised by `convertEntry`, which only read `cron` and `interval`.

`bus add-cron` already writes a `CronDefinition` directly so the registration path worked. Only the `config.json -> crons.json` migration ran on agent boot dropped these entries.

## Changes

- `src/types/index.ts`: extend `CronEntry.type` with `'cron'` as a recurring synonym so config files can self-document a crontab-scheduled entry. Add optional `schedule?: string` as an author-facing alias for the `cron` field.
- `src/daemon/cron-migration.ts`: `convertEntry` now reads `cron ?? schedule ?? interval` (cron wins when both crontab fields are present). Skip-log text upgraded to name all three possible fields.
- `tests/integration/crons-migration.test.ts`: 5 new cases — `type:"cron"` with `cron` field, `type:"cron"` with `schedule` alias, `schedule` alias on `type:"recurring"`, precedence when both are present, no-schedule skip path.

## Test plan

- [x] `npm run build` — clean
- [x] `npx vitest run tests/integration/crons-migration.test.ts` — 23/23 pass (18 existing + 5 new)
- [x] Full `npm test` — 11 test files fail on this branch but the same 11 fail on clean `upstream/main` (pre-existing, all dashboard/Next.js missing-package or phase4/5 simulation env, none touch cron-migration paths)
- [x] No regression on `tests/unit/cli/bus-crons.test.ts` or `tests/integration/upgrade-cron-teaching-cli.test.ts` (both pass in isolation)

### DIFF
--- a/src/daemon/cron-migration.ts
+++ b/src/daemon/cron-migration.ts
@@ -161,17 +161,25 @@ function convertEntry(
   entry: CronEntry,
   agentName: string,
 ): { cron: CronDefinition } | { skip: string } {
-  const { name, type, interval, cron: cronExpr, fire_at, prompt } = entry;
+  const { name, type, interval, cron: cronExpr, schedule: scheduleAlias, fire_at, prompt } = entry;
 
-  // Treat absent `type` as "recurring" (spec requirement)
+  // Treat absent `type` as "recurring" (spec requirement).
+  // `cron` is a synonym for `recurring` — both auto-load as a recurring
+  // CronDefinition. The distinct type lets config authors self-document a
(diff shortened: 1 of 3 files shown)


soletraderai:
Closing — going with a local workaround instead. Thanks.

PR #524, opened 2026-05-20
fix(daemon): make cron-expression scheduling timezone-deterministic

asachs01:
## The bug

`nextFireFromCron` matched a cron expression's hour / day-of-month / day-of-week
fields against **process-local** time (`getHours()` / `getDate()` / `getDay()`).
The result depended on whatever timezone the process happened to run in:

- The **daemon** (launched from an ET shell) fired `0 8 * * *` at 08:00 EDT =
  **12:00 UTC**.
- The **CLI** (`bus list-crons`, running UTC) computed the "next fire" display
  at **08:00 UTC**.

The 4-hour disagreement made `list-crons` show a next-fire on the wrong
calendar day — which read as a **"skipped daily cron."**

It was **not a skip.** The `cron-execution.log` shows the daemon fired
morning-review and evening-review **every single day, 05-09 → 05-19, with zero
gaps.** The bug was (a) timezone non-determinism — the schedule depended on the
launching shell — and (b) the daemon-vs-CLI display split. Interval crons
(`"4h"`) were never affected because they don't match clock fields.

## The fix — interpret cron expressions in an explicit, configured timezone

- **`nextFireFromCron(expr, fromMs, timeZone = 'UTC')`** extracts clock fields
  via `Intl.DateTimeFormat` with an explicit IANA zone (`hourCycle: 'h23'`).
  The formatter is built **once** outside the minute-walk. Default `'UTC'` — a
  missing config can no longer make scheduling depend on the launching shell.
- **`CronScheduler`** gains a `timeZone` option. The daemon resolves it per
  agent from the org `context.json` `cron_timezone` field
  (`resolveCronTimezone`).
- The CLI `list-crons` and the IPC `list-all-crons` (`computeNextFire`) resolve
  and pass the **same** zone — so the displayed next-fire always agrees with
  the daemon's actual fire.
- **`CtxEnv` gains `cronTimezone`**, populated by `resolveEnv`. It is
  deliberately **separate** from `timezone` (which drives day/night-mode) so a
  fleet can run day/night in one zone and schedule crons in another without
  entangling the two.
- `templates/org/context.json` ships `cron_timezone: "UTC"` — the framework
  default. A fresh install schedules crons in UTC.
- An invalid `cron_timezone` (a typo) is **validated and falls back to UTC with
  a warning** — a misconfig must not silently drop every cron-expression cron
  for the org.

### Named zone, not a fixed offset — do not "simplify" this

`cron_timezone` takes a **named IANA zone** (e.g. `America/New_York`), never a
fixed `+N`-hours offset. `Intl` then handles the EST/EDT transition (and any
future DST rule change) automatically — a hardcoded offset would silently break
twice a year. The code comments say this explicitly; please keep it.

## Why "configured timezone" and not "force UTC + re-pin every cron"

The first-considered fix was `getUTC*` + re-pinning each daily cron's hour
field. Grounding it found 10 cron-expression crons fleet-wide, and two of them
(`dev/ci-triage 0 12,16,20 * * 1-5`, `dev/friday-wrap 0 20 * * 5`) **cannot be
losslessly re-pinned to UTC** — a +4h shift crosses midnight and breaks the
day-of-week field. Interpreting in a configured zone gives the same
determinism guarantee with **zero re-pinning** and no midnight-crossing puzzle.

## Verification

- `npm run build` clean, `tsc --noEmit` clean.
- **1671 tests pass.** New: timezone-determinism tests (result invariant to
  `process.env.TZ`; honours the supplied/configured zone; DST-aware named
  zone; UTC default), a `CronScheduler` `timeZone`-option test, and
  `resolveCronTimezone` unit tests (incl. invalid-zone fallback). The existing
  `nextFireFromCron` tests were rewritten from process-local to UTC-explicit
  assertions so they are deterministic on any CI runner.
- (3 unrelated `dashboard/` test files fail to load on a pre-existing
  `better-sqlite3` native-binding issue — not touched by this change.)

## Deploy step (required, post-merge)

**This fix is not active until the daemon is restarted.** It changes
cron-expression *interpretation*, which is read at `CronScheduler`
construction — a merged-but-not-restarted daemon still runs the old behavior.
Post-merge: **restart the daemon.** An org's live `context.json` must set `cron_timezone`
to the zone its crons were authored against, or those crons shift on restart.
The WYRE fleet's org `context.json` is set to `America/New_York` (its current
de-facto firing behavior — the daemon runs ET), so **no cron expression needs
re-pinning** — every cron keeps its exact wall-clock. `orgs/*/context.json` is
runtime state (gitignored), not part of this diff; the WYRE value is already
set on disk.

## Review

Full triangle (daemon code): murph (author) · warden · analyst.


asachs01:
## warden — daemon-code correctness review: PR #524 — **APPROVE**, no findings

Reviewed the four points requested.

**(1) `clockFieldsInZone` field extraction — correct.**
- `hourCycle: 'h23'` yields `"00"`–`"23"`; the `% 24` is a sound defensive net for the known ICU quirk of emitting `"24"` at midnight in some environments → maps to `0`.
- `parseInt(p.value, 10)` — explicit radix 10 everywhere, so a `"08"`/`"09"` value can't be misread as octal.
- Locale pinned `'en-US'` — this is load-bearing and correct: it makes `weekday: 'short'` deterministically `"Sun"`–`"Sat"` (the `WEEKDAY_INDEX` map keys) and guarantees ASCII digits for the 2-digit fields (a non-pinned locale could emit localized month/weekday names or non-ASCII digits and break `parseInt`).
- `weekday` map `Sun:0 … Sat:6` matches both the cron DOW convention and the old `Date.getDay()`.
- Month: `formatToParts` month is **1-based**, and the new code matches it 1-based against the `months` array — note the old code's `getMonth() + 1` correctly *became* a direct use. A natural spot for an off-by-one; it's right.

**(2) Minute-walk terminates across DST — correct.**
The walk increments **epoch-ms by a fixed 60 000** for `MAX_MINUTES` iterations — iteration behavior is *identical to pre-PR*; the only change is that each candidate's clock fields are projected via Intl-in-named-zone instead of `Date`-local. A named IANA zone (not a fixed offset — the comment explicitly forbids "simplifying" to one, correctly) lets Intl resolve EST/EDT automatically: a spring-forward-skipped target hour simply defers to the next valid occurrence; a fall-back-doubled hour matches at its first occurrence. Termination is bounded by `MAX_MINUTES` unconditionally, and DST costs at most ~one extra day of walk — well within headroom for any sub-annual cron.

**(3) NaN handling — correct, two-layer.**
`nextFireFromCron` returns `NaN` on a bad expression *or* an unknown `timeZone` (the formatter `try/catch`). I verified the upstream guard: `resolveCronTimezone` validates the candidate with `new Intl.DateTimeFormat('en-US', { timeZone })` and, on `RangeError`, writes a warning and **falls back to `'UTC'`** — so a typo'd `cron_timezone` cannot silently NaN-drop *every* cron-expression cron for the org. The two validation sites are consistent: zone validity is independent of the field options, so any zone `resolveCronTimezone` accepts will also construct cleanly inside `nextFireFromCron`. Defense-in-depth: realistic layer (resolveCronTimezone → warned UTC fallback, crons keep firing) + backstop (per-cron NaN → existing log-and-skip).

**(4) Catch-up / `nextFireAt` advance paths — correct.**
`computeNextFireAt` now takes `timeZone` and is threaded `this.timeZone` at all three call sites — initial scheduling, the success-advance, and the failed-window/missed-slot advance. Interval shorthands (`6h`/`30m`) short-circuit via `parseDurationMs` before `nextFireFromCron` — timezone-independent, as they should be. The catch-up `referenceMs = max(candidates)` path is unchanged except for the consistent zone threading.

**Defaults are consistent:** `nextFireFromCron(…, 'UTC')`, `CronSchedulerOptions.timeZone ?? 'UTC'`, `resolveCronTimezone → 'UTC'` — a fully-unconfigured fleet interprets every cron in UTC on both the daemon and the CLI, which is exactly the determinism the fix restores.

**One out-of-scope note (not a finding, not a regression):** `MAX_MINUTES = 366 * 24 * 60` is a pre-existing bound, unchanged by this PR — a genuinely multi-year expression (e.g. `0 0 29 2 *`) could still walk out to `NaN`. #524 neither introduces nor worsens that; flagging only so it's on record that I considered the walk bound and DST has ample headroom under it.

**Verdict: APPROVE, no findings.** The fix is correct and well-reasoned — epoch-walk + Intl-named-zone projection is the right shape, the locale pin and `h23` handling are sound, the zone is validated upstream with a safe fallback, and the timezone is threaded consistently through every fire-computation path.

asachs01:
**5-AREA — cortextos PR #524 (cron-scheduler timezone-determinism). VERDICT: APPROVE.** General-quality lens (warden owns the daemon-internals correctness read). Clean, correct, well-tested; one cutover dependency, verified already satisfied.

**1. ROOT CAUSE — correct.** `nextFireFromCron` matched cron clock-fields against process-local `Date` (`.getHours()/.getDate()/.getDay()` …), so a daemon in ET and a CLI in UTC read different clocks for the same expression — a 4–5h disagreement. The "skipped daily cron" was phantom: no fire was missed (cron-execution.log proved the fires), the CLI's displayed next-fire just disagreed with the daemon's. Diagnosis is sound and the PR is honest that nothing was actually skipped.

**2. THE FIX — correct approach.** `clockFieldsInZone` extracts the clock fields via `Intl.DateTimeFormat.formatToParts` against an explicit IANA zone. Key correctness points all hold:
- Built from an IANA zone *name*, not a fixed offset — Intl resolves EST/EDT and future DST automatically; the comment explicitly forbids "simplifying" to a hardcoded `+N` offset. Right call.
- The formatter is constructed ONCE before the minute-scan loop — construction is the costly part, per-minute `formatToParts` is cheap. Correct perf shape.
- Locale pinned `'en-US'` → stable short weekday strings that match `WEEKDAY_INDEX` (Sun:0…Sat:6, matching cron dow and the old `.getDay()` semantics). Pinning the locale is necessary and done.
- DST semantics fall out naturally: the UTC minute-scan + `formatToParts` produces conventional wall-clock-cron-under-DST behavior (a spring-forward-skipped wall time simply doesn't fire that day; a fall-back-doubled time fires at the first occurrence). Standard, not a defect.

**3. resolveCronTimezone — the invalid-zone fallback is sound.** Precedence `CTX_CRON_TIMEZONE` > org `context.json:cron_timezone` > `UTC`. An invalid zone is caught, warned to stderr, and falls back to UTC. The rationale is exactly right: an unvalidated bad zone would make `nextFireFromCron` return `NaN` → `loadCrons` skips the cron as "cannot parse schedule" → a confusing partial outage; validate-and-fallback-with-warning degrades safely instead. Keeping `cron_timezone` distinct from `timezone` (day/night-mode) is the correct separation and is well-documented.

**4. WIRING — consistent, which is the whole point.** Daemon (`agent-manager` → `CronScheduler.timeZone`), CLI (`bus.ts` list-crons), and `ipc-server` (`computeNextFire`/`listAllCrons`) all resolve through `resolveCronTimezone`, so the daemon's actual fire and every displayed next-fire agree. `listAllCrons` caches cron-tz per org (`cronTzByOrg` Map) so a 30-agent fleet does not re-read the same `context.json` once per agent — good. Signatures stay backward-compatible (`timeZone` is an optional last param defaulting to `'UTC'`).

**5. TESTS — genuinely exercise the bug.**
- The rewritten `nextFireFromCron` tests (process-local → UTC-explicit) are not masking a regression — the new assertions ("0 13 * * *" → UTC hour 13) are arithmetically correct for the new explicit-zone contract; they test the corrected behavior.
- The TZ-determinism block genuinely exercises the bug: it forces `process.env.TZ` to UTC / America-New_York / Asia-Tokyo and asserts identical output for the same `timeZone` arg. Node *does* honor runtime `process.env.TZ` changes for `Date`, so were the fix incomplete (a stray `.getHours()`), the output would shift and the test would go red. The DST-aware case ("0 8" ET → 12:00 UTC on an EDT date) is correct. `origTZ` save/restore is proper hygiene.
- Built the branch and ran `tests/unit/daemon/cron-scheduler.test.ts` + `tests/unit/utils/env.test.ts` myself — build success, 40/40 pass (full suite 1671 per author; CI is fork-gated until a maintainer approves, so local + review is the verification).

**CUTOVER NOTE (verified, already satisfied — confirm on the deploy target).** This makes the mechanism correct but `cron_timezone` is config-driven: if the WYRE org's `cron_timezone` were unset, every cron-expression cron would re-interpret in UTC on the next daemon restart and fire 4–5h off its intended ET wall-clock — silently. I checked: `orgs/wyre/context.json` already carries `"cron_timezone": "America/New_York"`, so the fleet keeps ET interpretation across the restart — "zero cron re-pin" genuinely holds with zero behavior shift. One thing to confirm: that file is untracked (local org config), so whichever checkout the daemon runs from must have it — low risk, since the daemon already reads `timezone` from the same file for day/night mode, but worth an explicit confirm at deploy.

**NET: APPROVE.** Correct root cause, correct DST-safe approach, sound invalid-zone fallback, consistent daemon/CLI wiring, tests that genuinely red→green the determinism property. The one config dependency is already satisfied in the WYRE org context. Over to warden's daemon-internals leg, then merge.

PR #362, opened 2026-05-08
fix: cherry-pick 3 wedge-mitigation fixes from upstream

loganbronstein:
## What this does (plain English)

Three small fixes that stop your agents from getting stuck and make the dashboard more useful. All three already shipped to the public cortextOS repo upstream — this PR just brings them onto your fork so your running fleet can use them.

**1. Stops marketing and scribe from wedging at /compact.** Adds a self-healing watchdog that runs in the background and recovers an agent that froze during context compaction. Without it, when an agent hits ~88% context and the compact dialog appears, no one is around to click through, and the agent silently stops responding. The watchdog detects this state and force-restarts the agent.

**2. Makes "idle" vs "wedged" distinguishable on the dashboard.** Adds a real bus event every time an agent receives an inbound Telegram message. Before: dashboards counted only outbound activity, so a wedged agent and a quiet idle agent looked identical. After: a freshly responsive agent shows recent inbound events; a wedged one does not.

**3. Lets specialist agents skip Telegram polling to reduce session weight.** Specialist agents like coder and analyst do not need to own the Telegram interface — only the orchestrator does. This adds a config flag that turns off the inbound poller for specialists, cutting unnecessary network calls and timeout noise that contribute to context bloat.

## Why this matters today

Boss has been seeing the marketing and scribe agents wedge at the compact dialog. The fleet-wide wedge that hit boss + analyst + coder simultaneously yesterday was the same pattern. Without these fixes the spiral keeps recurring on every long-running session.

## Important: 2-step deploy

Merging this PR is **not enough on its own**. The cortextOS daemon is currently running with the old code in memory — even after merge it will keep using the old code until the daemon process is restarted.

After you greenlight merge, the steps are:

1. Merge this PR.
2. Manually restart the daemon so it picks up the new code (boss will run the command — no risk to running agents, they auto-recover).

If you only do step 1, nothing changes. Step 2 is required.

## Cherry-pick trail

These three commits are already on the public cortextOS repo. This PR brings them onto your fork:

| upstream commit | upstream PR | what it does |
|---|---|---|
| `37f7780` | grandamenium/cortextos#297 | telegram_polling config flag |
| `7fa7414` | grandamenium/cortextos#267 | telegram_received bus event |
| `371ef01` | grandamenium/cortextos#327 | self-healing watchdog scripts |

Cherry-picked in order to maximize independent verifiability — each commit can be reviewed and reverted on its own.

## Verification

- `npm run build` — clean
[...]

PR #531, opened 2026-05-22
fix(orchestrator): persist idle-overlap-watcher at tracked scripts/ path

revopsglobal:
## Summary

- `idle-overlap-watcher.sh` cron was pointing to `orgs/revops-global/agents/orchestrator/scripts/` — inside the gitignored `orgs/` tree, making it vulnerable to `git clean` deletion
- Script was already moved to `scripts/` in #362; this PR completes the fix:
  1. Removes the `orgs/` copy from git tracking (`git rm --cached`)
  2. Adds the cron entry to `orchestrator/config.json` (was missing — cron was never recreated on restarts, only surviving as a daemon-side entry)
  3. Updates the live daemon cron prompt to reference `scripts/idle-overlap-watcher.sh` via `cortextos bus update-cron`

## Test plan
- [ ] `git ls-files orgs/revops-global/agents/orchestrator/scripts/idle-overlap-watcher.sh` returns empty
- [ ] `scripts/idle-overlap-watcher.sh` still tracked and intact
- [ ] `cortextos bus list-crons orchestrator | grep idle` shows `scripts/` path
- [ ] After orchestrator restart, cron is recreated from config.json with correct path


revopsglobal:
Created against wrong upstream — closing. PR will be recreated against RevOps-Global-GIT/cortextos.

PR #532, opened 2026-05-22
fix(task-validate): pass result to validator before completeTask writes it

revopsglobal:
## Root Cause

`validateTask` reads `task.result` from disk at validation time, but `completeTask` hasn't written it yet — the result string is held in `effectiveResult` in the caller. The LLM always saw `(no result provided)` and scored ≤6, blocking every completion and forcing `--override`.

## Fix

Added an optional `result?: string` parameter to `validateTask` and `buildPrompt`. The `complete-task` command now forwards `effectiveResult` at the call site. Priority order: override → on-disk value → `(no result provided)`.

## Changes

- `src/bus/task-validate.ts` — `buildPrompt(task, resultOverride?)` uses override when present; `validateTask(paths, id, result?)` forwards it through
- `src/cli/bus.ts` — `validateTask(paths, id, effectiveResult)` at the call site
- `tests/unit/bus/task-validate.test.ts` — 4 new tests: result override used, disk fallback, no-result placeholder, auto-pass on no success_criteria

## Test plan

- [x] `npx vitest run tests/unit/bus/task-validate.test.ts` — 4/4 pass
- [x] `npm test` — 156/156 pass


revopsglobal:
Opened against wrong upstream — closing. PR will be filed against RevOps-Global-GIT/cortextos.

PR #538, opened 2026-05-23
fix(CLAUDE.md): annotate grandamenium/cortextos as read-only upstream

revopsglobal:
## Summary
- Adds a visible callout above the `git clone` line in the framework root `CLAUDE.md`
- Clarifies `grandamenium/cortextos` is a read-only upstream mirror — never push or open cross-fork PRs to it
- All writes go to `RevOps-Global-GIT/cortextos`
- References `[feedback_git_push_target]` memory rule

Root cause: the existing clone URL normalized grandamenium as a write target for agents that don't load the per-user MEMORY.md. This change makes the restriction visible at the file level.

Part of CLAUDE.md audit fixes (PR 1 of 3).

## Test plan
- [ ] `CLAUDE.md` renders correctly in repo UI
- [ ] No other files changed


revopsglobal:
Opened against wrong repo — should be RevOps-Global-GIT/cortextos. Closing.

PR #542, opened 2026-05-27
fix(daemon): resolve projects dir from config.home in shouldContinue

revopsglobal:
## Summary
Fixed a bug in `shouldContinue()` where the conversation directory was resolved using `homedir()` instead of the agent's configured `config.home` path. This caused --continue restarts to fire against the wrong Claude projects directory, leading to first-run/no-session loops for agents like hub-dogfood that use `~/.claude-team`.

## Changes
- Added `resolve` import from `path` module
- Updated `shouldContinue()` to compute `effectiveHome` from `config.home` (if set) before resolving the projects directory
- Tilde expansion is handled via `replace(/^~/, homedir())`

## Testing
- `npm run build` passes (TypeScript compiles cleanly)
- `npm test` passes (all 2571 tests)

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>


revopsglobal:
Closing — re-opening against RevOps-Global-GIT/cortextos fork per org policy (zero writes to upstream mirror).

Issue #283, opened 2026-04-30
spawn-worker silently drops --model flag; workers freeze at Claude Code model selection screen


grandamenium:
## Symptom

`cortextos spawn-worker <name> --dir <dir> --prompt "..." --model <model>` returns success and shows the worker as `running` in `cortextos list-workers`, but the worker session is wedged at the Claude Code model selection screen and never executes the prompt.

Example from one of nick's workers (`blotato-skill-builder`, ~15 min wedged):

```
There's an issue with the selected model (opus). It may not exist or you may
not have access to it. Run /model to pick a different model.

❯
⏵⏵ bypass permissions on (shift+tab to cycle)                              ● high · /effort
```

`list-workers` reports it as `running (pid …)` because the PTY is alive — but no prompt has been processed.

## Root cause

Two compounding bugs.

### Bug 1 — `--model` is silently dropped on the worker spawn path

`src/daemon/agent-manager.ts:699-713`

```ts
async spawnWorker(name, dir, prompt, parent?, model?) {
  ...
  const config = model ? { model } : {};   // ← never used

  ...

  await worker.spawn({ ...env, ...(model ? {} : {}) }, prompt);
  //                            ^^^^^^^^^^^^^^^^^^^^
  //                            ternary returns {} either way
}
```

`config` is constructed and immediately discarded. The spread `(model ? {} : {})` is `{}` regardless of `model`. Worse, `WorkerProcess.spawn(env, prompt)` does not even take a config/model parameter:

`src/daemon/worker-process.ts:46-55`

```ts
async spawn(env: CtxEnv, prompt: string): Promise<void> {
  ...
  this.pty = new AgentPTY(env, {}, logPath);   // ← config={} hardcoded
  ...
}
```

So `--model claude-opus-4-7` from the CLI:
1. parses correctly in `src/cli/workers.ts`
2. travels through IPC to the daemon
3. lands in `spawnWorker(..., model)`
4. is **dropped on the floor** before reaching `AgentPTY`
5. `AgentPTY.buildClaudeArgs` sees no `config.model` → no `--model` flag is passed to `claude` → the worker falls back to whatever `~/.claude/settings.json` has.

### Bug 2 — fallback path is brittle

[...]

Issue #195, opened 2026-04-20
bug(gap-detector): stale fire-time causes false gap alerts after --continue restart storm


grandamenium:
## Description

After a rapid --continue restart storm (~25 restarts over 15 minutes, 13:27-13:43 UTC 2026-04-20), the cron gap detector emitted false-positive alerts for all 8 configured crons, each reporting the same stale value: \"last fired 2939 minutes ago\".

CronList confirmed all 8 crons were active throughout. No crons were missing.

## Root Cause (hypothesis)

The gap detector tracks last-fired timestamps separately from the CronList scheduler. During rapid --continue restarts, the last-fired record is not updated (crons survive via session continuity but don't re-fire). The gap detector sees a stale timestamp and fires alerts, even though the crons are live and will fire on schedule.

## Impact

- Noise: agents receive spurious gap alerts and waste a tool call on CronList to confirm a false positive
- The identical \"2939 minutes\" value across all 8 crons confirms this is a timestamp staleness issue, not individual cron failures

## Expected Behavior

Gap detector should cross-reference CronList before emitting an alert. If the cron appears in CronList as active, suppress the gap alert (or downgrade it to a warning).

## Reproduction

1. Trigger ~10+ rapid --continue restarts within a short window
2. Observe gap detector emitting alerts for all crons with identical stale timestamps
3. Confirm via CronList that all crons are active

## Environment

- Detected: 2026-04-20 ~13:50 UTC by sentinel
- Agent: sentinel (lifeos)
- Crons affected: all 8 (heartbeat, usage-monitor, nightly-metrics, auto-commit, check-upstream, catalog-browse, daily-digest, theta-wave)

grandamenium:
Observed active storm on sentinel Apr 23 2026: after cron restore from config.json on session start (~01:07 UTC), the gap detector immediately began firing alerts for all 8 crons at 6577 minutes stale. Alerts fire in waves every ~13 minutes, cycling through all 8 crons each wave. Gap value increments by ~14 minutes per wave (6577 → 6591 → 6604...). This will continue until each cron naturally fires and resets its baseline. Three full waves observed so far (24 alerts). The false-positive storm is triggered by any session restore after a restart gap — the detector sees the pre-restart fire-time as the last fire, not the post-restore registration time.

grandamenium:
[...]

PR #558, opened 2026-05-30
ci: add reusable migration-collision-check workflow (Theta Phase 8)

asachs01:
## Summary

Adds `.github/workflows/migration-collision-check.yml`, a `workflow_call` reusable workflow that consumer repos wire into their PR CI to detect cross-PR migration-number collisions. Fails the calling PR's check if any migration it INTRODUCES has a number that:

- (a) is already used by a migration on the base branch, **or**
- (b) is already proposed by another open (non-draft) PR in the same repo.

Also catches internal duplicates within a single PR (two new files with the same number).

Output is a GitHub `::error file=<path>::` annotation naming the colliding artifact (`on <base> as <path>` or `in open PR #N as <path>`). Renumber-to-next-available is the actionable guidance.

## Why cortextos hosts this

cortextos is the platform/template repo for the org; the action is reusable infrastructure other repos (initial target: conduit, which has the 3-digit `NNN_xxx.sql` convention established by PR #283 renumber + #285 assertNumericContiguity) will consume.

## Consumer usage

```yaml
# in <consumer>/.github/workflows/migrations.yml
on:
  pull_request:
    types: [opened, synchronize, reopened]

jobs:
  migration-collision-check:
    uses: grandamenium/cortextos/.github/workflows/migration-collision-check.yml@main
    # optionally override:
    # with:
    #   migrations-path: db/migrations
    #   filename-pattern: '^([0-9]{14})_'
```

## Defaults & inputs

- `migrations-path` — directory containing migration files. Default `migrations`.
- `filename-pattern` — POSIX ERE matching migration filenames; capture group 1 must be the number. Default `^([0-9]{3})_` (conduit convention). Override for timestamp-style.

## Security posture

- Permissions minimal: `contents: read`, `pull-requests: read`.
- `actions/checkout` SHA-pinned (matches the existing `ci.yml` convention).
- All untrusted inputs flow through `env:` into `os.environ` — never inlined as `${{ }}` in shell commands.
- Python heredoc uses `<<'PYEOF'` (single-quoted) so the body is not subject to shell variable expansion.
- `subprocess.run` everywhere uses list args, never `shell=True`.
- Does not consume `github.head_ref` / PR title-body / other attacker-influenced strings.

## Scope

Per the directive: this PR adds **only** the reusable workflow. The conduit-side wire-in (a small `.github/workflows/migrations.yml` in conduit calling this workflow) is a separate follow-up — keeps this PR tight and lets the workflow land first so consumer wire-ins reference a stable target.

## Local testing

Smoke-tested against four cases with a local git fixture:

1. **Base-branch collision** — PR introduces `002_foo.sql` while main has `002_orgs.sql` → exit 1, error names \`on main as migrations/002_orgs.sql\` ✓
2. **Internal duplicate** — PR adds both `002_foo.sql` and `002_bar.sql` → exit 1, dup detected before cross-PR scan ✓
3. **Safe** — PR adds `003_safe.sql` (next available) → exit 0, no collisions ✓
4. **Non-matching file** — `migrations/README.md` in PR → filtered out by `filename-pattern`, exit 0 ✓

The cross-PR (open-PR) collision path is not exercised locally because it needs the real `gh api` against a repo with open PRs — that will exercise live in the conduit wire-in PR.

## Test plan

- [ ] Workflow YAML parses (verified by GitHub Actions on the PR's own CI run)
- [ ] Consumer-side smoke-test (next PR): wire into conduit, push a PR that collides with main → verify error annotation
- [ ] Consumer-side smoke-test: open two conduit PRs with the same proposed number → second one's check should fail naming the first


asachs01:
This landed via the WYRE-AI fork instead: WYRE-AI/cortextos#193 (merged 2026-09-17), with two same-day follow-ups — #195 (fix wrong org in the usage example) and #196 (base-branch filter + no PR-count cap for the cross-PR scan). Noting here so this doesn't read as abandoned — the work shipped, just not through this PR.

Issue #534, opened 2026-05-22
Two P1 bugs in quota-watchdog/resume bash scripts: stale-snapshot trap + piped-while-loses-state


noogalabs:
## Summary

Two P1 bugs found in the quota-watchdog tooling during a downstream fork's pre-import review pass. Both are in `bin/` shell scripts and easy to fix.

- **Bug A — `bin/quota-watchdog.sh`**: stale-snapshot trap on `list-agents` failure
- **Bug B — `bin/quota-resume.sh`**: `FAILED+=()` updates are discarded by the piped-while subshell

---

### Location

`bin/quota-watchdog.sh`, around lines ~213-220 (pause path, snapshot block):

```bash
RUNNING_JSON='[]'
if AGENTS_OUT=$("$CORTEXTOS" bus list-agents 2>/dev/null); then
  RUNNING_JSON=$(echo "$AGENTS_OUT" | "$JQ" -c '[.[] | select(.running == true) | .name]')
  [ -z "$RUNNING_JSON" ] && RUNNING_JSON='[]'
fi
COUNT=$(echo "$RUNNING_JSON" | "$JQ" 'length')
```

### Trigger

When `cortextos bus list-agents` fails (exit non-zero, e.g. daemon mid-restart, IPC timeout, transient socket error), the `if` branch is skipped silently. `RUNNING_JSON` stays at its default `[]`.

The script then proceeds into the pause-action path: the stop loop iterates zero times — no agents are actually paused — but `paused.json` is written with `"agents_paused": []`. The watchdog now believes it has paused agents, but no state-change happened.

### Resulting failure mode

1. The stop loop iterates zero times — **no agents are actually paused**.
2. `paused.json` is written with `"agents_paused": []` — the watchdog now **believes it has paused agents**, but no state-change happened.
3. Future cycles see `[ -f "$PAUSED_FILE" ]` and short-circuit into the "already paused" branch. Either they hold pause (no further action) or they auto-resume 0 agents and archive the stale file.
4. The real running agents continue burning quota past the threshold. The watchdog appears to have done its job (state file present), but actually missed the trip entirely.

### Reproduction sketch

1. Stub or temporarily disable `cortextos bus list-agents` so it returns non-zero (any transient daemon issue does this naturally).
2. Force a trip by setting `QUOTA_THRESHOLD_PCT=99` so `REMAINING_PCT < THRESHOLD_PCT`.
3. Run the watchdog.
[...]

PR #560, opened 2026-05-30
fix(quota-watchdog): #534 stale-snapshot trap + piped-while state loss; macOS portability

syntasticstudios:
## Summary

Fixes the two P1 bugs reported in #534, plus the portability changes needed to run the quota tooling on a non-`/root` (macOS / Homebrew) deployment.

## Bug A — `bin/quota-watchdog.sh` stale-snapshot trap

When `cortextos bus list-agents` fails (daemon mid-restart, IPC timeout, transient socket error), `RUNNING_JSON` stayed at its `[]` default and the script fell through to write `paused.json` with `agents_paused: []`. The watchdog then believed it had paused agents it never touched; subsequent cycles short-circuited on the "already paused" branch while the real agents kept burning quota.

**Fix:** bail (with an operator Telegram alert) when the snapshot fails, and a second bail when zero agents are running. Neither path writes paused-state. The stop loop is also converted to a process-substitution `while` loop so stop failures are captured (same subshell-state class as Bug B).

## Bug B — `bin/quota-resume.sh` piped-while loses state

`echo "$AGENTS_JSON" | jq -r '.[]' | while ...` ran the loop body in a subshell, so `FAILED+=("$AGENT")` was discarded — a future patch depending on the failure count would silently regress.

**Fix:** `done < <(echo "$AGENTS_JSON" | jq -r '.[]')` keeps the loop in the parent shell, plus a partial-failure log line and Telegram alert.

Both follow the fixes proposed in the issue.

## Additionally: macOS portability (maintainers may wish to cherry-pick separately)

Required to run the watchdog off the Linux `/root` layout:
- `CORTEXTOS` / `JQ` / `CCUSAGE` resolve via `command -v` with the original Linux absolute paths as fallback (Homebrew installs `cortextos` under `/opt/homebrew/bin`; `ccusage` may be absent — it's already guarded by `[ -x ]`).
- OAuth token acquisition reads the macOS login Keychain (`security find-generic-password -s "Claude Code-credentials"`) on Darwin, falling back to `$CLAUDE_CREDS` on Linux. `$CLAUDE_CREDS` itself now defaults to `$HOME/.claude/.credentials.json` instead of the hardcoded `/root` path.
- New `scripts/self-healing/com.cortextos.quota-watchdog.plist.template` (launchd) passes the `CTX_*` / `WATCHDOG_*` / `QUOTA_*` env so the watchdog runs every 5 min on macOS — there was no plist for the quota-watchdog previously (only for the self-healing daemon watchdog).

## Tests

`tests/shell/quota-watchdog-534.test.sh` drives the **real** scripts under a stubbed `cortextos`. Covers: Bug A snapshot-failure bail, Bug A zero-agents bail, healthy-quota no-trip control, a positive trip recording the right agents, and Bug B resume failure reporting. **5/5 pass** (RED against the pre-fix scripts, GREEN after). Pure shell — runs outside the vitest glob; invoke directly:

```
bash tests/shell/quota-watchdog-534.test.sh bin/quota-watchdog.sh bin/quota-resume.sh
```

## Verification on a live macOS deployment

- Pre-fix tests: RED (snapshot-failure trap reproduced, Bug B confirmed latent).
- Post-fix: 5/5 GREEN.
- Real run, healthy quota: `method=api remaining=60% threshold=10% paused=no` — no `paused.json`, fleet untouched.
- Forced-trip dry-run (`QUOTA_THRESHOLD_PCT=100 QUOTA_DRY_RUN=1`): correctly snapshotted all 10 running agents and logged "would stop 10", wrote no state, stopped nothing.
- Installed via launchd (`StartInterval 300`); first `RunAtLoad` run clean, stderr empty.


syntasticstudios:
Superseded by #564 — clean 4-file version off current `main` (this branch had drifted and accidentally included 100+ unrelated files). Same fix + tests.

PR #562, opened 2026-05-30
fix(bus): findTaskFile resolves orgless root tasks (<ctxRoot>/tasks/)

loganbronstein:
## Summary

`findTaskFile` could not resolve **orgless root tasks** — tasks filed without an org (e.g. daemon-created tasks from sender `cortextos`) that live at `<ctxRoot>/tasks/` rather than under any `orgs/<name>/tasks/` directory. They could be created but never updated or completed through the bus (`update-task` / `complete-task` threw "not found").

Surfaced by a fleet audit. Confirmed against real state: 5+ live task files at `$CTX_ROOT/tasks/`, one created the same day, all unreachable by the task lifecycle commands.

## Root cause

`findTaskFile` resolved only two tiers:
1. the caller's own org tasks dir (`paths.taskDir`), then
2. a scan of `orgs/*/tasks/`.

Orgless tasks live at `<ctxRoot>/tasks/`, **outside** the `orgs/*` tree, so the scan never reached them.

## Fix

Add a third lookup tier checking `<ctxRoot>/tasks/<id>.json`, inserted **after** the same-org fast path and **before** the cross-org scan — a single `existsSync`, so the hot path pays no scan cost. The same-org fast path still takes precedence if an id somehow exists in both places, so an agent never silently retargets the root copy of its own task.

Also corrects the three `"not found in any org under .../orgs/"` error messages, which were misleading now that the orgless root is searched too.

## Tests

New `Orgless root task lifecycle` describe block (real on-disk layout, no mocked path resolution):
- `findTaskFile` resolves an orgless root task to `<ctxRoot>/tasks/`
- `updateTask` reaches an orgless root task (the bug repro)
- `completeTask` reaches an orgless root task and marks it done
- same-org fast path still wins when an id exists in both the org dir and the root (precedence guard)

```
$ npx vitest run tests/unit/bus/task.test.ts
Test Files  1 passed (1)
     Tests  44 passed (44)
```

`npm run build` clean.

## Review

Adversarially reviewed by Codex in a temp worktree: 44/44 + typecheck clean, no blocking issues, HARD RULE 3 gate cleared.

## Risk

Contained to `findTaskFile`. The new tier only adds a lookup location; it cannot change resolution for any task that already resolved (own-org and cross-org paths are checked first / unchanged). Existing cross-org tests still pass.


drzachconner (approved):

PR #564, opened 2026-06-01
fix(quota-watchdog): #534 stale-snapshot trap + piped-while state loss; macOS portability (clean, supersedes #560)

### LINKED ISSUE #534: Two P1 bugs in quota-watchdog/resume bash scripts: stale-snapshot trap + piped-while-loses-state

syntasticstudios:
## Summary

Replaces #560 (which accidentally carried 100+ unrelated files because its base branch had drifted far from `main`). This is the same fix cherry-picked cleanly onto current `main` — **4 files only**.

Fixes the two P1 bugs from #534, plus the portability changes needed to run the quota tooling on a non-`/root` (macOS / Homebrew) deployment.

## Bug A — `bin/quota-watchdog.sh` stale-snapshot trap
When `cortextos bus list-agents` fails (daemon mid-restart, IPC timeout), `RUNNING_JSON` stayed `[]` and the script wrote `paused.json` with `agents_paused: []` — the watchdog believed it had paused agents it never touched, while real agents kept burning quota. Now: bail + operator alert on snapshot failure, plus a zero-running-agents bail. The stop loop is moved to a process-substitution `while` so stop failures survive (same class as Bug B).

## Bug B — `bin/quota-resume.sh` piped-while loses state
`echo | jq | while …` ran the loop in a subshell, discarding `FAILED+=()`. Switched to `done < <(…)` + a partial-failure log/Telegram alert.

## Portability (needed for non-/root installs)
- `CORTEXTOS`/`JQ`/`CCUSAGE` resolve via `command -v` with the Linux paths as fallback.
- OAuth token read from the macOS login Keychain on Darwin, falling back to `$CLAUDE_CREDS` (now `$HOME/.claude/.credentials.json`) on Linux.
- New `scripts/self-healing/com.cortextos.quota-watchdog.plist.template` (launchd, every 5 min).

## Tests
`tests/shell/quota-watchdog-534.test.sh` drives the real scripts under a stubbed `cortextos`: Bug A snapshot-fail bail, Bug A zero-agents bail, healthy-quota no-trip, positive trip, Bug B resume failure reporting. **5/5 pass** on `main`'s base.

Closes #534. Supersedes #560.

### DIFF
--- a/bin/quota-resume.sh
+++ b/bin/quota-resume.sh
@@ -35,8 +35,9 @@ export CTX_ROOT CTX_FRAMEWORK_ROOT CTX_ORG
 export CTX_AGENT_NAME="$BUS_AGENT"
 export CTX_AGENT_DIR="$CTX_FRAMEWORK_ROOT/orgs/$CTX_ORG/agents/$BUS_AGENT"
 
-CORTEXTOS=/usr/bin/cortextos
-JQ=/usr/bin/jq
+# Tool paths — env-overridable for portability (macOS Homebrew vs Linux).
+CORTEXTOS="${CORTEXTOS:-$(command -v cortextos || echo /usr/bin/cortextos)}"
+JQ="${JQ:-$(command -v jq || echo /usr/bin/jq)}"
 
 if [ ! -f "$PAUSED_FILE" ]; then
   echo "No paused state at $PAUSED_FILE — watchdog has not tripped (or already resumed)."
@@ -56,8 +57,11 @@ COUNT=$(echo "$AGENTS_JSON" | "$JQ" 'length')
 echo "Resuming $COUNT agents (paused at $PAUSED_AT):"
 log "resume requested: $COUNT agents (paused_at=$PAUSED_AT)"
 
+# Process substitution keeps the loop in the parent shell so FAILED survives —
+# a piped `echo|jq|while` runs the loop in a subshell and silently discards the
+# accumulated failures (#534 Bug B).
 FAILED=()
-echo "$AGENTS_JSON" | "$JQ" -r '.[]' | while IFS= read -r AGENT; do
+while IFS= read -r AGENT; do
   [ -z "$AGENT" ] && continue
   echo "  starting: $AGENT"
   if "$CORTEXTOS" start "$AGENT" >> "$LOG" 2>&1; then
@@ -66,7 +70,13 @@ echo "$AGENTS_JSON" | "$JQ" -r '.[]' | while IFS= read -r AGENT; do
     log "  start failed: $AGENT"
     FAILED+=("$AGENT")
   fi
-done
+done < <(echo "$AGENTS_JSON" | "$JQ" -r '.[]')
+
+if [ "${#FAILED[@]}" -gt 0 ]; then
+  log "resume partial: ${#FAILED[@]} of $COUNT failed: ${FAILED[*]}"
+  MSG="⚠️ Quota resume: ${#FAILED[@]} of $COUNT agents failed to start: ${FAILED[*]}. Investigate and retry manually."
+  "$CORTEXTOS" bus send-telegram "$CHAT_ID" "$MSG" --plain-text >> "$LOG" 2>&1 || true
+fi
 
 # Archive the paused-state file rather than delete (auditability)
 ARCHIVE_DIR="$STATE_DIR/history"
--- a/bin/quota-watchdog.sh
+++ b/bin/quota-watchdog.sh
@@ -47,20 +47,29 @@ export CTX_ROOT CTX_FRAMEWORK_ROOT CTX_ORG
 export CTX_AGENT_NAME="$BUS_AGENT"
 export CTX_AGENT_DIR="$CTX_FRAMEWORK_ROOT/orgs/$CTX_ORG/agents/$BUS_AGENT"
 
-CORTEXTOS=/usr/bin/cortextos
-JQ=/usr/bin/jq
-CCUSAGE=/usr/bin/ccusage
-CLAUDE_CREDS=/root/.claude/.credentials.json
+# Tool paths — env-overridable so the script is portable (macOS Homebrew puts
+# cortextos under /opt/homebrew/bin; ccusage may be absent). Defaults preserve
+# the original Linux layout.
+CORTEXTOS="${CORTEXTOS:-$(command -v cortextos || echo /usr/bin/cortextos)}"
+JQ="${JQ:-$(command -v jq || echo /usr/bin/jq)}"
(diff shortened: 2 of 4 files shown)


syntasticstudios:
Closing — the fork (syntasticstudios/cortextos) runs the live deployment and has merged this fix directly (syntasticstudios#16). Leaving upstream out of scope.

PR #570, opened 2026-06-02
tests(hooks+crash-alert): align to post-fix execFile shape from #556 / #550

InocuousCabbage:
## Summary
PRs #556 (a89cee2) and #550 (100839e) changed `src/bus/hooks.ts` and `src/hooks/hook-crash-alert.ts` to invoke `execFile(process.execPath, [cliPath, ...])` when `CTX_FRAMEWORK_ROOT` is set, falling back to `execFile('cortextos', [...])` when it isn't. The existing tests target the fallback shape but didn't isolate from the env, so if `CTX_FRAMEWORK_ROOT` is set in the test runner the prepended `cliPath` shifts the positional indices and the assertions break.

## Fix
Save + clear + restore `CTX_FRAMEWORK_ROOT` in \`beforeEach\` / \`afterEach\` of the affected describe blocks so the fallback path is tested deterministically. Surgical; no \`src/\` touched.

- \`tests/unit/bus/hooks.test.ts\`: top-level describe \`src/bus/hooks — Day-2 per-handler wiring\`
- \`tests/unit/hooks/hook-crash-alert.test.ts\`: \`describe('notifyAgents'\`

## Verification
Full suite returns to the pre-merge baseline (23 fail / 13 files / 1595 pass on my local) — the 10 merge-introduced regressions are cleared without disturbing anything else.

## Noted gap (out of scope for this PR)
The new framework-root code path (\`execFile(process.execPath, [cliPath, ...])\`) is now NOT separately covered after the env-clear. A small describe block per file that explicitly sets \`CTX_FRAMEWORK_ROOT\` and asserts the prepended-cliPath positional shift would close this — happy to file as a follow-up if useful.

24 lines added across 2 files, tests-only.


drzachconner (approved):

PR #571, opened 2026-06-02
fix(telegram): retry post() once on transport errors — undici keepalive socket reuse

InocuousCabbage:
## Symptom
\`cortextos bus send-telegram\` intermittently fails with \`Telegram API request failed: TypeError: fetch failed\` while direct curl with the identical payload **succeeds**. \`getMe\` works. First observed 2026-05-30 in production.

## Root cause
Node 20+ native \`fetch\` (undici-backed) holds HTTP keepalive sockets by default. Telegram's API server silently drops idle sockets server-side (no \`Connection: close\` advertised). When the next request reuses a now-dead socket, undici throws \`TypeError: fetch failed\` (cause: \`UND_ERR_SOCKET\` / \`ECONNRESET\` / "socket hang up") **before any application data is written**.

This matches every observable:
- **Intermittent**: only fires when the chosen keepalive socket has been dropped server-side AND hasn't been reaped client-side yet. Long-running daemons (heartbeat, fast-checker) are the worst case.
- **curl works**: each curl opens a fresh TCP connection. No pool, no stale-socket reuse.
- **\`getMe\` works**: typically a one-off call; likely opens a fresh pool entry.

The \`AbortSignal.timeout(15000)\` already added by #86 / #3019317 fixes a *different* failure mode (server-side stuck request). The timeout never fires here — the failure occurs at the transport layer before any timer reaches 15s.

## Fix
Single retry on transport-level errors in \`TelegramAPI.post()\`.

Safe for this failure mode because dead-socket-reuse fires **before** the request body is written — the message is provably not delivered, so retrying cannot duplicate. We deliberately do NOT retry on \`TimeoutError\` / \`AbortError\` (body may have been written and Telegram may have processed) or on \`Telegram API error\` (API responded with a structured error — retrying wouldn't change that).

Test update: \`validateCredentials\` \`network_error\` path queues 2 throws instead of 1 to match the retry semantics; final wrapped error still preserves the original cause (\`ENOTFOUND\`).

## Verification
Full suite returns to baseline on my local (23 fail / 13 files / 1595 pass — matches pre-patch). All 85 telegram tests pass.

## Scope
- Applied only to \`post()\` (text-message path: \`sendMessage\`, \`getMe\`, \`getChat\`, \`setMessageReaction\`, etc.).
- \`sendPhoto\` / \`sendDocument\` deliberately deferred — \`FormData\` body re-read semantics warrant separate analysis. Happy to file as a follow-up.

## Pattern match
Aligns with existing reliability fixes in this file: #86 (fetch timeout) + #5356b91 (parse_mode retry).

## Size
\`src/telegram/api.ts\`: ~12 net lines. \`tests/unit/telegram/api.test.ts\`: 4 lines (queue 2nd throw + comment).


drzachconner (approved):

PR #573, opened 2026-06-02
docs(skills/comms): tighten reply_to direction + document ack-inbox for non-substantive acks

InocuousCabbage:
## Why
Two related ambiguities in the comms \`SKILL.md\` surfaced in a self-audit on a multi-turn exchange:

### 1. \`reply_to\` direction is ambiguous

Today's SKILL says:

> 3. For agent messages, always include the \`msg_id\` as the reply_to argument so conversations thread correctly

It doesn't specify **which** \`msg_id\`. In a multi-turn thread, the correct value is the **LATEST INCOMING** message's \`msg_id\` — NOT the agent's own outbound \`msg_id\` from a prior turn. Conflating them:
- Leaves the latest incoming message un-ACK'd (redelivers after 5 min)
- Threads against the wrong parent in inbox state

This is a real failure mode caught in the audit transcript.

### 2. \`ack-inbox\` isn't documented

When a message contains no substantive request and the agent has nothing to add (e.g. \"ACK\", \"Standby\" replies that don't need a reply-chain extension), the right tool is:
\`\`\`bash
cortextos bus ack-inbox <msg_id>
\`\`\`
It clears the redelivery timer without sending a performative reply. Today the SKILL only documents \`send-message\`, so agents over-reply, polluting threads with empty round-trips.

## What this PR does

1. Tightens Step 3 to specify \"INCOMING msg_id\" + adds a brief warning about the outbound-msg_id confusion.
2. Adds a new \"Non-substantive acks\" subsection documenting \`ack-inbox\` as the right tool for low-content acks, distinct from \`send-message\`.

Same wording across all 5 framework targets:
- \`templates/agent\`
- \`templates/orchestrator\`
- \`templates/analyst\`
- \`templates/agent-codex\` (codex-runtime variant)
- \`community/skills/comms\`

## Deliberately excluded
Concrete community-maintained agents (\`community/agents/*\` — \`security\`, \`research-agent\`). Same precedent as the heartbeat rollout: instantiations, not templates; downstream maintainers update on their own cadence.

## Size
5 files, ~50 lines added. Pure docs.


drzachconner (approved):

PR #578, opened 2026-06-03
feat(workers): Hermes — multi-backend worker with health-gated auto-fallback

noogalabs:
## What this adds

A fault-tolerant **worker dispatch layer** that runs a single task through a chain of backends — **Codex → Gemini → Claude** — health-checking each before use and failing over automatically when one is down, logged out, or misconfigured. A dead or unentitled backend becomes a *non-event* instead of a hard failure, and doer work can run off non-Anthropic budgets first while keeping Claude as a reliable backstop.

This is the **engine only** — a reusable module with no agent/deployment coupling and **no bus/daemon changes**.

## Motivating incident

A Codex dispatch failed deterministically: the effective model resolved server-side to one the account wasn't entitled to, the entitlement gate fired at *turn-start* (not config-load), and the spawn-retry burned three identical failures before giving up. Hermes turns that into an automatic single failover — it pins a known-safe model, classifies an unentitled-model error as a **non-retryable `config-error`**, fails over immediately with **zero wasted retries**, and reports which backend served and why each prior one was skipped.

## Design

- **`WorkerAdapter` interface** (`src/workers/hermes/base.ts`) — `health()` + `execute()` + `safeModels()`, a structured `ExecResult`, and an `Unavailability` taxonomy (`no-binary` / `no-auth` / `config-error` / `rate-limit` / `transient` / `timeout` / `process-fail`) that drives retry-vs-failover. Deliberately distinct from the PTY-lifetime `VendorAdapter` (which has no health method or result type).
- **Three adapters** — Claude (`claude -p --output-format json`), Codex (`codex-companion task --model`, model-pinnable), Gemini (a thin Python entrypoint over the existing genai client + retry classifier). The load-bearing Codex classifier distinguishes a *turn-never-started* entitlement error (`config-error`, 0 retries) from a *mid-flight* blip (`transient`, retryable) by position relative to first token-usage.
- **Orchestrator** — `INTAKE → SELECT → health-gated EXECUTE loop → REPORT`. Health gates execute (a dead backend is skipped, never run); `config-error` fails over immediately; every task terminates in an explicit result (`served` or `exhausted`) — **never silent**. Fully dependency-injected for testing.
- **Policy** — chain ordering with `preferred` hoist, per-reason retry caps, backoff, and a total wall-clock budget guard.
- **Observability** — structured JSONL per attempt/transition + events naming the serving backend and the chain-of-reasons.
- **CLI** — a `hermes-run` verb and a new `hermes-worker` runtime literal (additive to `AgentConfig.runtime` + `add-agent` `VALID_RUNTIMES`; decoupled from `--template`, so existing agent creation is unaffected) and an optional `HermesTaskMeta` provenance field on `Task`.

## Tests — 48, all green against `main`

- **Unit:** per-adapter classification (incl the exit-0-but-`is_error` catch and the Codex entitlement position-boundary), orchestrator contract (health-gates-execute, never-silent terminal, default model pin, preferred hoist), policy table.
- **Integration (the proof):** `backend-down → fallback`, and the motivating `config-error → immediate-failover` regression (execute called exactly once → fails over → served).

`npx tsc --noEmit` clean; `npx vitest run tests/**/workers/hermes` → 48/48.

## Notes

- No `bus`/`daemon`/`FastChecker` changes — the dispatcher participates via existing conventions.
- `gemini_task.py` wraps the existing `knowledge-base/scripts/mmrag.py` client + retry classifier; the broken Gemini-CLI adapter path is intentionally not used.
- `gpt-5.3-codex` / `-spark` are excluded from the Codex safe-model list (the structural prevention) in addition to the runtime classifier (the catch).


noogalabs:
Thanks for the review — addressed in `15dab31`:

1. **Reply to the task creator (P1)** — `hermes-run` now replies to `task.created_by` (the delegator) instead of `task.assigned_to`, which on a delegated task is the worker itself. Added a recipient regression assertion to the orchestrator served-path test (it previously asserted the message fired but not its recipient — the exact gap that let this through).
2. **`hermes-worker` runtime spawn (P2)** — it already booted via the default `AgentPTY` (Hermes is a first-class agent whose brain is claude-code; the dispatcher runs in-process via `hermes-run`), but the implicit fall-through read as unhandled. Made it explicit at the agent-spawn site. Behavior-preserving — this is the one small daemon touch in the PR (the description's "no daemon changes" line is now superseded by this note).
3. **Gemini health parity (P2)** — `health()` now checks the same credential sources `gemini_task.py`'s `get_api_key()` uses (env `GEMINI_API_KEY` **or** `gemini_api_key` in the mmrag config), so it no longer over-skips Gemini when the key lives only in the config.

`tsc --noEmit` clean; full Hermes suite 48/48.

noogalabs:
@codex review

PR #580, opened 2026-06-03
feat(bus): shell-safe message body via --body-stdin/--body-file

grandamenium:
## Summary

Adds a shell-safe input path for bus message bodies so callers never have to
interpolate untrusted or code-bearing text into a shell-expanded positional
argument.

- `cortextos bus send-message <to> <priority> --body-stdin` reads the body from
  stdin
- `cortextos bus send-message <to> <priority> --body-file <path>` reads the body
  from a file
- positional `<text>` remains supported and unchanged (backward compatible)
- precedence: `--body-stdin` > `--body-file` > positional `<text>`
- `bus/send-message.sh` passes the new flags through transparently

New `resolveMessageBody()` helper in `src/bus/message.ts` centralizes the
resolution logic.

### Why

On 2026-06-02 a bus message whose body contained shell command-substitution
(`$(...)` / backticks) was expanded by the *caller's* shell before it ever
reached the send layer, executing embedded commands. The send path itself was
already safe (it execs node with a quoted argv array, no eval), so stripping
metacharacters at the send layer would not have prevented the expansion AND
would corrupt legitimate code-bearing messages. The correct fix is an input
path where the body is never placed on a shell-expanded command line at all:
`--body-stdin` / `--body-file`. This closes the injection-via-message-body
class for any caller that adopts it.

## Test Plan

- `tests/unit/bus/message.test.ts`: 6 new tests covering `resolveMessageBody`
  precedence and each input source. Full file 14/14 passing (`vitest run`).
- `tsc --noEmit`: clean.
- Manual smoke test: a body containing `$(...)` and backticks piped via
  `--body-stdin` is stored verbatim with zero command execution.

MERGE gated on James per repo policy.


drzachconner (approved):

PR #583, opened 2026-06-04
fix(env): root-anchor CTX_FRAMEWORK_ROOT subordination check — '/' no longer rejects subordinate agentDir (#527)

### LINKED ISSUE #527: env subordination guard: fwRoot='/' breaks startsWith check (P2)
## Summary

The subordination guard added in PR #348 / commit `fe39493` uses raw string concatenation `fwRootResolved + sep` to anchor the `startsWith` check. When `CTX_FRAMEWORK_ROOT` resolves to filesystem root `/`, the prefix becomes `//`, so legitimately-subordinate paths like `/tmp/agent` fail the check and incorrectly throw the sandbox/live-leak error.

## Code reference

`src/utils/env.ts` (post-#348):

```typescript
if (agentDirResolved !== fwRootResolved && !agentDirResolved.startsWith(fwRootResolved + sep)) {
  throw new Error(`Resolved CTX_AGENT_DIR ... is not under CTX_FRAMEWORK_ROOT ...`);
}
```

When `fwRootResolved === '/'` and `sep === '/'`, the concatenated prefix is `'//'`, which no normal absolute path will start with.

## Repro

```bash
CTX_FRAMEWORK_ROOT=/ CTX_AGENT_DIR=/tmp/agent npx tsx -e "import('./src/utils/env.js').then(m => m.resolveEnv())"
# → throws "is not under CTX_FRAMEWORK_ROOT '/'"
```

Expected: no throw; `/tmp/agent` is under `/`.

## Suggested fix

Either special-case the root, OR use segment-aware containment:

```typescript
// Option A: special-case root
const prefix = fwRootResolved === sep ? fwRootResolved : fwRootResolved + sep;
if (agentDirResolved !== fwRootResolved && !agentDirResolved.startsWith(prefix)) { ... }

// Option B: path.relative-based
const rel = path.relative(fwRootResolved, agentDirResolved);
if (rel.startsWith('..') || path.isAbsolute(rel)) { throw ... }
```

## Severity

P2. Edge case — most environments don't run with framework root at filesystem root. Caught by Codex automated review on a downstream fork's pickup of `fe39493`. Filing here so the fix lands upstream rather than forking.

## Discovery context

Caught by Codex bot review on a noogalabs/ascendops cherry-pick PR that re-applied `fe39493`. Verified against the live code; bug is in the original commit shipped to grandamenium/cortextos main.

thehoff:
## Summary

Fixes #527 — the sandbox/live subordination guard in `src/utils/env.ts` (from #348) anchored its containment check with raw `fwRootResolved + sep`. When `CTX_FRAMEWORK_ROOT` resolves to filesystem root `/`, that prefix becomes `//`, which no absolute path starts with — so a legitimately-subordinate `agentDir` like `/tmp/agent` was wrongly rejected with the sandbox/live-leak error.

## Fix

Anchor the prefix with a trailing separator only when one isn't already present:

```ts
const fwRootPrefix = fwRootResolved.endsWith(sep) ? fwRootResolved : fwRootResolved + sep;
```

Root `/` already ends in `sep`, so it's used as-is (`/`) instead of becoming `//`. No behaviour change for any non-root framework root.

## Tests

Added to the existing `Sandbox/live env subordination` block in `tests/sprint7-environment.test.ts`:

- **TC-E** (regression): `frameworkRoot='/'`, `agentDir='/tmp/agent'` — fails before the fix (threw `not under CTX_FRAMEWORK_ROOT`), passes after. Also asserts the resolved `frameworkRoot`/`agentDir` pass through so a silent containment regression can't hide behind a bare `not.toThrow()`.
- **TC-F** (property guard): `/tmp/foobar` must NOT be treated as subordinate to `/tmp/foo` — ensures the anchor isn't "fixed" by simply dropping the separator (the segment-boundary safety the original `+ sep` provided is preserved).

## Verification

- `npm run build` — clean
- `npx vitest run tests/sprint7-environment.test.ts` — all pass (incl. TC-A…TC-F)

## Scope note

A reviewer suggested switching the string `startsWith` check to a path-aware containment test (`realpathSync` + `path.relative`) to also defend against symlink/case-variant aliases. That's deliberately **out of scope** here: this guard intentionally operates on paths that need not exist on disk (the existing TC-A/TC-B tests pass non-existent dirs), and `realpathSync` throws on non-existent paths — it would change the guard's contract and break those cases. Symlink hardening is a separate concern from the P2 root-anchor bug this PR fixes; happy to follow up in its own PR if wanted.

Closes #527.


drzachconner (approved):
Full diff reviewed. Fixes a double-separator edge case: when CTX_FRAMEWORK_ROOT resolves to '/', the old fwRootResolved+sep became '//' which no path starts with, so all subordinate agentDirs were wrongly rejected. New code: if fwRootResolved already ends in sep, don't append another. Two tests added — one confirms '/' accepts /tmp/agent, one guards against the near-prefix false-positive (/tmp/foobar must not match under /tmp/foo). Correct. Merge.

Issue #515, opened 2026-05-19
bug(dashboard): Memory and Goals tabs perform silent full overwrites — hand-authored file content permanently lost on save


Modaffinity:
### Description

Two dashboard tabs perform destructive writes to critical agent files without any warning or conflict detection:

**Memory tab (MEMORY.md):** The PATCH handler unconditionally overwrites the agent's `MEMORY.md` with whatever the dashboard tab submits:

```ts
// dashboard/src/app/api/agents/[name]/route.ts line 108
await fs.writeFile(paths.memoryMd, body.memoryRaw as string, 'utf-8')
```

No read-merge, no diff, no ETag check. Any content in `MEMORY.md` that is not present in the submitted body is permanently lost.

**Goals tab (GOALS.md):** Every Goals tab save fires `generate-md`, which regenerates `GOALS.md` from `goals.json` using a fixed template:

```ts
// dashboard/src/app/api/agents/[name]/goals/route.ts lines 117-128
spawn(process.execPath, [distCliPath, 'goals', 'generate-md', '--agent', name, '--org', org])
```

Any hand-authored content in `GOALS.md` that is not expressible in the `goals.json` schema is silently destroyed. The regenerated `GOALS.md` even includes a comment "Do not edit this file directly." — but the dashboard provides no visible indication that saving via the Goals tab will trigger this overwrite.

### Expected behavior

- Memory tab: performs a read-merge or presents a conflict warning before overwriting `MEMORY.md`
- Goals tab: warns the operator that saving will regenerate `GOALS.md` from `goals.json`, destroying any hand-authored content not captured in the structured fields

### Actual behavior

- **Memory tab:** `MEMORY.md` is fully replaced on every save. Operators lose hand-authored memory context permanently on first dashboard use.
- **Goals tab:** `GOALS.md` is regenerated from `goals.json` on every Goals tab save. Operators who hand-author objectives or KPIs directly in `GOALS.md` lose that content silently.

### Reproduction

**Memory tab:**
1. Hand-author content in the agent's `MEMORY.md` that is not shown in the dashboard's Memory tab (e.g., add a section below the fold)
2. Open the dashboard Memory tab and save (without changing anything)
3. Inspect `MEMORY.md` — hand-authored sections are gone

**Goals tab:**
1. Hand-author a section in `GOALS.md` that is not representable in the dashboard Goals form
2. Open the dashboard Goals tab and save any field
3. Inspect `GOALS.md` — hand-authored content is replaced by template output

[...]

PR #584, opened 2026-06-04
fix(dashboard): optimistic concurrency for MEMORY.md saves — stop silent overwrites (#515)

thehoff:
Addresses the **Memory tab** half of #515 (the silent `MEMORY.md` overwrite). The Goals-tab warning is a separate follow-up PR.

## Problem

The agent-detail PATCH handler wrote `MEMORY.md` unconditionally — no read, no compare. Any content edited on disk (by the agent, a cron, or a hand-edit) between the dashboard loading the tab and the operator hitting Save was permanently lost.

## Fix — optimistic concurrency (ETag-style)

- `getAgentDetail` returns `memoryHash` = `sha256(memoryRaw)` (the version the tab loaded).
- The Memory tab echoes that hash back on save; the PATCH handler **requires** it whenever `memoryRaw` is present (missing → `400`).
- In a **pre-flight before any field is written**, the handler reads the current `MEMORY.md`, hashes it, and returns **`409 Conflict` without writing** if it no longer matches — so a stale-memory conflict can't partially persist identity/soul either.
- On success the handler returns the new `memoryHash`; the tab updates its baseline so consecutive saves in the same tab don't false-conflict.
- A 409 surfaces as a non-destructive banner ("MEMORY.md changed on disk… reload, then re-apply") with a Reload button, **keeping the operator's unsaved edits**.

The hash is **server-issued** (not client-computed), avoiding any client/server hashing mismatch causing false conflicts.

### Known limitation (intentional, scoped)

This is optimistic concurrency, not locking. A small TOCTOU window remains between the pre-flight read and the write. It closes the **reported** single-operator silent-overwrite (load → out-of-band change → clobber); fully serialising concurrent writers to the same agent would need a per-agent file lock, which is deliberately out of scope for this fix (and is the standard trade-off for ETag-style endpoints without a transactional store).

## Tests

New `dashboard/src/app/api/agents/[name]/__tests__/route.test.ts`:
- stale `memoryHash` → **409, file unchanged** (the data-loss regression guard) — verified failing before the fix
- matching `memoryHash` → writes (200)
- `memoryRaw` without `memoryHash` → **400, file unchanged**
- success returns the post-write hash; a follow-up save using it does **not** false-409

## Verification

- `tsc --noEmit` (dashboard) — clean
- new route tests 4/4 pass
- full suite: no new failures vs baseline

Closes the Memory-tab portion of #515.


drzachconner (approved):
Full diff reviewed. Adds optimistic concurrency for MEMORY.md dashboard saves: client sends the etag (content hash) it read, server rejects with 409 Conflict if the file changed since then. Prevents silent overwrites when two browser tabs or a user + an agent write simultaneously. 6 files, clean separation (GET returns etag, PUT checks it, error surfaced in UI). Tests cover the conflict case. Correct. Merge.

PR #585, opened 2026-06-04
fix(dashboard): warn that Goals-tab saves regenerate GOALS.md (#515)

thehoff:
The **Goals tab** half of #515 (companion to #584, which fixed the Memory tab).

## Problem

Saving the Goals tab writes `goals.json` and then regenerates `GOALS.md` from a fixed template. That regeneration is **by design** — `GOALS.md` is a generated artefact ("do not edit this file directly"). But the dashboard gave no clear warning, so an operator who hand-authored objectives/KPIs directly in `GOALS.md` would lose that content on the next Goals-tab save with no indication it would happen. The only hint was a muted line ("Changes regenerate GOALS.md automatically") that didn't mention data loss.

## Fix

Since the regeneration is correct behaviour, the fix is **operator awareness**: a prominent warning banner at the top of the Goals tab —

> Saving regenerates `GOALS.md` from the structured fields below. Any content hand-authored directly in `GOALS.md` — outside Daily Focus, Goals, and Bottleneck — will be replaced.

(Replaces the passive muted note; the orchestrator-context line is kept.)

Frontend-only; no behaviour change to the save path.

## Verification

- `tsc --noEmit` (dashboard) — clean

Closes the Goals-tab portion of #515. Together with #584 (Memory tab), this completes #515.


drzachconner (approved):
Full diff reviewed. Adds a warning banner to the Goals tab in the dashboard explaining that saves overwrite the hand-authored GOALS.md (via the structured goals.json source). Uses IconAlertTriangle from @tabler/icons-react (already in the project). Pure UI addition, no logic change. Correct. Merge.

Issue #510, opened 2026-05-19
PTY injection race: non-atomic PASTE+ENTER drops simultaneous-fire crons silently while reporting fired


cr4shOverr1de:
## Summary

When multiple persistent crons fire in quick succession (within ~120 ms of each other on the same scheduler tick), the daemon records all of them as `status=fired` in `cron-execution.log`, but the agent only ever receives the **first** cron's `[CRON FIRED ...]` prompt. The remaining injections are silently dropped at the PTY/TUI layer. The agent has no way to know it missed work; the operator sees a "successful" execution log.

Reproduced on a stock daemon-managed agent in the `lab` org. Affects any agent with multiple crons whose fire times align — common cases: a heartbeat cron + a check-approvals/metrics/auto-commit cron sharing the same minute boundary.

## Repro

Configure 4 daily crons all anchored to the same wall-clock minute (e.g. the 00:11 UTC window an agent was spawned at):

```bash
cortextos bus add-cron <agent> heartbeat 4h "<heartbeat prompt>"
cortextos bus add-cron <agent> nightly-metrics 24h "<metrics prompt>"
cortextos bus add-cron <agent> auto-commit 24h "<commit prompt>"
cortextos bus add-cron <agent> check-upstream 24h "<upstream prompt>"
```

Wait for the 24 h alignment window. Observed in `cron-execution.log`:

```
{"ts":"2026-05-19T00:11:50.483Z","cron":"heartbeat","status":"fired","attempt":1,"duration_ms":0,"error":null}
{"ts":"2026-05-19T00:11:50.588Z","cron":"nightly-metrics","status":"fired","attempt":1,"duration_ms":1,"error":null}
{"ts":"2026-05-19T00:11:50.597Z","cron":"auto-commit","status":"fired","attempt":1,"duration_ms":0,"error":null}
{"ts":"2026-05-19T00:11:50.603Z","cron":"check-upstream","status":"fired","attempt":1,"duration_ms":0,"error":null}
```

All four logged as `fired` within 120 ms. Verified by inspecting the agent's full conversation transcript: **only the first cron's prompt arrived in the session**. The other three never appeared on any subsequent turn.

[...]

PR #590, opened 2026-06-04
fix(daemon): serialize per-agent PTY injects to close the cron collision race (#510)

thehoff:
## Problem (#510)

When two crons fire at the same moment, their PASTE+ENTER cycles interleave into the same PTY and corrupt both messages. There is also no honest failure signal, so a dropped inject is silently lost rather than retried.

## Fix

Serialize injects per agent so each PASTE+ENTER cycle completes before the next begins, and make the failure path honest so a dropped inject is retryable.

- **Per-agent serialization barrier** — `injectMessageDetailed` chains through a per-agent `injectChain` (ordering only: the predecessor's failure is ignored for sequencing, and the barrier is always released in `finally` so one bad inject can't wedge the queue). Each caller gets its own structured outcome.
- **`injectMessage` now resolves `Promise<boolean>`** after the deferred ENTER, so callers can fully sequence one cycle before starting the next. It resolves `false` (never rejects) on ENTER failure, so fire-and-forget callers can't raise an unhandled rejection while awaiting callers still learn the submit failed.
- **Honest failure on teardown/restart** — the write closure throws on a torn-down PTY (instead of optional-chaining to a silent no-op) and on a `lifecycleGeneration` change, so a teardown or restart inside the inject window yields `INJECT_FAILED` rather than a false success or a stale message bleeding into a freshly-restarted session.
- **Dedup rollback** — `MessageDedup.forget()` removes the recorded hash on `INJECT_FAILED`, so a cron's retry of the same content isn't swallowed as a duplicate.
- **`handleFireCron`** reserves the cooldown slot *before* the now-async inject and rolls it back on failure/throw, preventing concurrent double-fire.
- `INJECT_FAILED` added to `IPCResponse.code`; the inject path is threaded async through `agent-manager`, `ipc-server`, `fast-checker`, and `worker-process`.

## Tests

New unit coverage: serialization ordering, `INJECT_FAILED` propagation + queue-still-usable-after-failure, torn-down-PTY rejection, restart-generation rejection, and dedup rollback. `tsc --noEmit` clean; full suite green (pre-existing unrelated reds aside).

### DIFF
--- a/src/daemon/agent-manager.ts
+++ b/src/daemon/agent-manager.ts
@@ -1041,8 +1041,8 @@ export class AgentManager {
    * Used by `cortextos bus test-cron-fire` to fire a cron immediately for testing.
    * Returns true if the agent is running and the inject succeeded; false otherwise.
    */
-  injectAgent(agentName: string, text: string): boolean {
-    return this.injectAgentDetailed(agentName, text).ok;
+  async injectAgent(agentName: string, text: string): Promise<boolean> {
+    return (await this.injectAgentDetailed(agentName, text)).ok;
   }
 
   /**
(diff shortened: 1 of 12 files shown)


thehoff:
Superseded by #616 (this PR rebased onto current main, MERGEABLE). Closing to avoid a duplicate merge.

Issue #513, opened 2026-05-19
bug(bus): create-task --desc silently truncated at 120 chars in assignee inbox notification


Modaffinity:
### Description

When `create-task` is called with `--assignee` set to a different agent, the create-task handler sends an inbox notification to the assignee's PTY context. The notification's description field is silently capped at 120 characters:

```ts
// src/cli/bus.ts line 169
const desc = opts.desc ? ` — ${opts.desc.slice(0, 120)}` : '';
```

The full description is stored in the task JSON without truncation. The PTY injection layer (`inject.ts`) uses `MAX_CHUNK=4096` and does not truncate. The 120-char cap applies only to the inbox notification message — meaning the assignee agent receives a truncated description with no indication that content was cut.

To see the full task description, the agent must explicitly call `bus list-tasks --format json` and locate the task by ID — a step not mentioned in AGENTS.md or the task assignment protocol.

### Expected behavior

The assignee receives the full task description in the inbox notification, or — if truncation is required for some display reason — a clear truncation marker (e.g., `… [truncated — run list-tasks --format json for full description]`) so the agent knows to look elsewhere.

### Actual behavior

Description in the inbox notification is silently capped at 120 chars (`opts.desc.slice(0, 120)`). The clip occurs mid-word with no ellipsis, no warning, and no pointer to the full task JSON. The agent operates on whatever fragment arrived.

### Reproduction

```bash
# Create a task with a long description assigned to a second agent
CTX_ORG=<org> cortextos bus create-task 'specification task' \
  --desc 'This is a detailed task description that exceeds one hundred twenty characters and will be silently truncated before delivery.' \
  --assignee <other-agent>

# Observe the assignee PTY inbox — description is clipped at char 120, no marker
# Full description is available via:
CTX_ORG=<org> cortextos bus list-tasks --format json | jq '.[] | select(.title=="specification task") | .description'
```

### Evidence

- `src/cli/bus.ts` line 169: `const desc = opts.desc ? \` — ${opts.desc.slice(0, 120)}\` : '';` — assignee notification description capped at 120 chars
- `src/pty/inject.ts` `injectMessage()`: `MAX_CHUNK=4096`; does not truncate the message it receives
- `src/daemon/fast-checker.ts` `formatInboxMessage()`: no truncation
- `src/bus/message.ts` `sendMessage()`: no truncation
- The task JSON stored on disk contains the full, untruncated description

[...]

PR #595, opened 2026-06-04
fix(security): sanitize untrusted text before PTY fence injection (prompt/terminal injection)

thehoff:
## Problem

Untrusted message text is injected into agent PTYs wrapped in a Markdown ``` fence (`src/daemon/fast-checker.ts`), but the text is not sanitised first. A message body containing a ``` run closes the fence early — everything after it stops being quoted and is read by the agent as **prompt input** (prompt injection). The same paths also forward raw ANSI / control bytes (terminal-control injection).

Found while working #513; this is pre-existing and cross-cutting (all message paths), so it's fixed separately here.

## Fix

New `src/utils/pty-sanitize.ts` → `sanitizeForPtyFence()`:
- **Control stripping** — reuses `stripControlChars` (ANSI CSI/OSC/ESC + C0 + DEL) and additionally drops `\r` (a PTY line-overwrite primitive it keeps) and the **C1** range (U+0080–U+009F) it doesn't cover.
- **Fence breakout** — inserts a zero-width space between backticks in any run of 2+, so a ``` can never re-form, while leaving single-backtick inline code intact.

Applied to every untrusted interpolation in the `fast-checker` PTY formatters: inbox `msg.text`; Telegram text / photo / document / video / voice bodies & captions / transcript; the reply / last-sent / recent-history context strings; the `.urgent-signal` body; Telegram `fileName`; and `from` display-names. The intentional unfenced slash-command path is sanitised too (normal commands are unaffected; a crafted `/… ```inject` can no longer break out).

## Tests

`tests/unit/utils/pty-sanitize.test.ts`: fence-breakout neutralised; ANSI/C0/DEL/mixed payloads stripped; newlines/tabs and single-backtick inline code preserved; empty string.

## Scope / follow-ups (intentionally not in this PR)

- Single-line context labels (`[Replying to: "…"]`) still permit newlines (they can't escape a fence, only add lines) — minor hardening.
- Slash-command **allowlist** (which commands a remote Telegram user may invoke) — a policy decision distinct from injection.
- `buildReminderBlock` and other **daemon-built prompt fragments** — a separate subsystem with its own untrusted-interpolation surface.

These are noted as follow-ups (not included here).

## Verification

- `npm run build` — clean
- sanitiser tests pass; full suite no new failures vs baseline

### DIFF
--- a/src/daemon/fast-checker.ts
+++ b/src/daemon/fast-checker.ts
@@ -10,6 +10,7 @@ import { AgentProcess } from './agent-process.js';
 import type { TelegramAPI } from '../telegram/api.js';
 import { KEYS } from '../pty/inject.js';
 import { stripControlChars } from '../utils/validate.js';
+import { sanitizeForPtyFence } from '../utils/pty-sanitize.js';
 
 type LogFn = (msg: string) => void;
 
@@ -221,7 +222,7 @@ export class FastChecker {
     const replyNote = msg.reply_to ? ` [reply_to: ${msg.reply_to}]` : '';
     return `=== AGENT MESSAGE from ${msg.from}${replyNote} [msg_id: ${msg.id}] ===
 \`\`\`
-${msg.text}
+${sanitizeForPtyFence(msg.text)}
 \`\`\`
 Reply using: cortextos bus send-message ${msg.from} normal '<your reply>' ${msg.id}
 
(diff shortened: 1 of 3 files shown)


thehoff:
Superseded by #592 — thanks @grandamenium, your `wrapFenceSafe` (dynamic unescapable fence, byte-preserving) + forged-header quoting is a cleaner approach than the zero-width-space sanitiser here. Closing this.

Following up with a small focused PR that reuses your new `wrapFenceSafe`/`sanitizeForPtyInjection` helpers on the remaining untrusted paths #592 didn't reach: Telegram media **captions** (photo/document/voice), **file_name**, the **.urgent-signal** body, and callback-query **from** names.

PR #602, opened 2026-06-05
docs(templates): require substantive user-visible restart notifications

grandamenium:
## Summary

Agent templates only distinguished `--continue` restarts (no notification) from hard restarts, so agents defaulted to bare "back online" messages on restarts the user actually noticed or asked for. This adds a user-visible-restart template to all four agent AGENTS.md templates requiring four elements in one substantive message:

1. **Restart class** — planned bounce / crash recovery / config reload / context handoff, and why
2. **Carried through** — what work and state survived intact
3. **Pending** — what still waits and on whom
4. **New since last contact** — what changed during the restart, or that nothing did

## Test Plan

- Doc-only change, 4 files +28/-4, identical hunk per template
- Verified the replaced line was byte-identical across all four templates before patching


drzachconner (approved):
Full diff reviewed. Adds substantive restart notification protocol to all 4 template AGENTS.md files (agent, agent-codex, analyst, orchestrator). The 4-element structure (restart class, carried-through, pending, new-since) replaces the vague prior guidance. Pure docs/template change. Correct. Merge.

PR #605, opened 2026-06-05
fix(lock): atomic windowless mutex — eliminate the pid-less inbox deadlock

CloudyZebraSEO:
## What this fixes

A **pid-less lock deadlock** in the on-disk inbox mutex (`src/utils/lock.ts`).

The old lock acquired in **two steps**: first `mkdir` the lock dir, then write the owner pid into it. If a process crashed in the gap between those steps, the lock existed on disk with **no owner recorded**. Liveness checks then couldn't decide whether the owner was alive or dead, so no process could ever safely reclaim it — the inbox wedged until a human intervened. We hit this in production as a multi-day agent-to-agent message deadlock.

## The fix

1. **Atomic, windowless mutex** — the lock is now created in a single atomic operation that carries the owner identity by construction. There is no longer a window in which a lock exists without an owner, so this deadlock class is eliminated. Self-healing: a dead owner's lock is provably reclaimable.

2. **Identity-safe release** — `releaseLock` verifies it owns the lock before removing it, so it can never delete a lock another process has since acquired.

3. **Orphan temp sweep** — crashed acquisitions can leave `*.tmp` artifacts; added a sweep (`sweepOrphanLockTemps`) plus a `bus check-inbox-health --sweep` command to clean them.

4. **EPERM-aware liveness** — fixes a latent bug where an `EPERM` from a liveness probe was treated as "owner dead." On multi-user systems that could let one user steal another user's *live* lock. `EPERM` now correctly means "alive but not ours."

5. **Dashboard pending-filter** (`dashboard/src/lib/data/reports.ts`) — aligns the report's pending-message count with the actual `checkInbox` filter.

## Testing

- `tsc --noEmit` clean on this branch (rebased onto current `main`).
- `tests/unit/utils/lock.test.ts` + `tests/unit/bus/inbox-health.test.ts` — **19/19 passing**.
- Reviewed by multiple independent passes before merge in our fork; one reviewer caught a recovery-path double-hold that was split into a separate follow-up (not included here — this PR is the deadlock fix only).

## Notes

These two commits are cherry-picked clean from our downstream fork onto current upstream `main`. The only merge resolution needed was an import-line union in `src/cli/bus.ts`. No behavioral changes outside the lock/inbox path.


drzachconner:
## Review - FLAG for second eyes before merging

This PR (from external contributor CloudyZebraSEO) changes core locking infrastructure: replaces the directory-based mutex in `utils/lock.ts` with a hardlink-based atomic lock, updates `message.ts` inbox handling, and `crons.ts` lock path comments. The stated fix (eliminate pid-less inbox deadlock) is real and the approach (hardlinks for atomicity) is sound on Linux/macOS POSIX filesystems.

**Concerns before merging:**
1. This is from an external contributor touching the locking layer that everything depends on. The approach is correct but I'd want a second set of eyes given the blast radius if the lock implementation has a subtle bug.
2. The new `checkInboxHealth` function added to `message.ts` is a useful addition but adds 100 lines of new code not covered by the main PR title — verify the tests are sufficient.
3. Confirm the hardlink approach works correctly on all deployment targets (macOS APFS and Linux ext4/xfs). Hardlinks are POSIX-atomic for link() but behavior on copy-on-write filesystems (APFS) should be verified.

Recommend: James or another core contributor does a full diff read before merging.

PR #607, opened 2026-06-06
fix(security): validate org/agent names before path construction

thehoff:
Companion to the task-id traversal fix (#598). Several bus + CLI paths interpolate a caller-supplied agent/org name into the filesystem with no validation, so a traversal name can read/write/restart outside the intended tree. `validateAgentName`/`validateOrgName` already exist (`src/utils/validate.ts`, `/^[a-z0-9_-]+$/`) but weren't applied on these paths.

## Fixes

- **`src/bus/crons.ts`** — `validateAgentName` at the `cronsFilePath` chokepoint (covers read/write/lock + add/update/remove/getCronByName) and at `getExecutionLogPage` (covers getExecutionLog).
- **`src/bus/agents.ts` `notifyAgent`** — validate `targetAgent` (feeds `state/<agent>`) and `from` (path-derived via sendMessage's filename; sendMessage's own check is swallowed by notifyAgent's try/catch, so this guard is what surfaces a bad `from`).
- **`src/cli/notify-agent.ts`** — validate `name` + `from` + `instance` before `resolvePaths` / `~/.cortextos/<instance>`.
- **`src/cli/get-config.ts`** — `validateOrgName` + `validateAgentName` before building `orgs/<org>/agents/<agent>` paths, clean `exit(1)` (matches goals.ts).
- **`src/cli/bus.ts`** — `notify-agent` / `soft-restart` / `send-mobile-reply` validate the target agent before building `state|logs/<agent>`; `soft-restart-all` skips invalid names from enabled-agents.json (logging why) and exits non-zero so a tampered config can't pass silently.

## Tests

`notifyAgent` rejects traversal in `targetAgent`/`from`; crons file I/O (read/write/add/remove/getExecutionLogPage) rejects a traversal agent name and still works for valid names. Full suite green (pre-existing unrelated reds aside).


drzachconner (approved):
Full diff reviewed. Adds validateAgentName/validateOrgName/validateInstanceId guards at every path-building entrypoint: agents.ts notifyAgent, crons.ts cronsFilePath + getExecutionLogPage, bus.ts soft-restart/notify-agent/reply-inbox, get-config.ts, notify-agent.ts CLI. Regex ^[a-z0-9_-]+$ blocks all traversal patterns. Tests cover traversal rejection + valid name passthrough. No extraneous changes. Merge.

PR #608, opened 2026-06-06
fix(security): validate assignee at task creation

### LINKED ISSUE #600: create-task stores --assignee unvalidated — add validateAgentName at creation (defense-in-depth)
## Summary

`createTask` (src/bus/task.ts) validates `priority` via `validatePriority` but stores the `assignee` option raw into `assigned_to` — no `validateAgentName` call at creation time. A task can be created with an assignee like `../../../etc` and persists to disk as valid-looking task JSON.

## Current containment (why this is defense-in-depth, not an active escape)

Every downstream consumer of `assigned_to` validates independently:
- `save-output` deliverables path: rejects invalid agent names before the `deliverablesDir/assigned_to/taskId` join (added in #598)
- auto-notify: `sendMessage` calls `validateAgentName` on recipient (src/bus/message.ts)

Verified in sandbox (PR #598 soak): a task created with a traversal assignee produced zero filesystem escape — but it persists as an un-actionable task that fails only when something later consumes it.

## Proposed fix

Call `validateAgentName(assignee)` in `createTask` next to the existing `validatePriority(priority)` — fail fast at creation instead of persisting a task that every consumer must individually defend against. `validateAgentName` already exists in src/utils/validate.ts.

One consideration: `assigned_to` currently also accepts `human` / human-task conventions — the validator call should preserve whatever the legit value space is (check `[HUMAN]` task flows before tightening).

## Related

Adjacent to the chokepoint-guard hardening in #598 (which guards consumers; this guards the producer).

grandamenium:
## Summary
- Validates `assignee` with `validateAgentName()` in `createTask` (src/bus/task.ts, alongside the existing `validatePriority` check), so path-traversal payloads in `--assignee` are rejected at creation time instead of persisting to disk.
- Adds 2 unit tests: traversal payloads rejected at creation; legitimate agent names (including `human` and `james`) still pass. Fixtures normalized `agentA` → `agent-a` to match the validator's naming rules.

Closes #600

## Test Plan
- `npm run build` clean (tsc strict).
- `tests/unit/bus/task.test.ts` 46/46 passing.
- Live verification: `create-task --assignee '../../../etc'` rejected at creation (previously persisted, caught only at delivery).
- Designer sandbox soak to follow before merge-ready.


drzachconner (approved):
Full diff reviewed. Closes the stored-raw gap: validateAgentName() runs at createTask() before the assignee is persisted to disk, so downstream consumers (sendMessage, saveOutput, notifyAgent) inherit a clean value. Tests cover traversal rejection + underscore/hyphen/special-keywords passthrough. Minor test cleanup (agentA -> agent-a to pass the new validator). Merge.

PR #612, opened 2026-06-09
fix: canonicalize target path in isClaudeDirOperation (macOS /private…

davidjmohamed1-a11y:
… symlink)

The post-merge unit suite caught a deterministic positive-control failure in tests/unit/hooks/hooks.test.ts: the "sanity: a genuine not-yet-existing file directly inside .claude is still allowed" assertion (line 154) inside the symlink-escape security test expected true but got false on macOS.

Root cause in src/hooks/index.ts (PR #594):

  const canonAgentDir = canonicalizePath(resolve(base));
  const claudeRoot = join(canonAgentDir, '.claude');
  const target = resolve(canonAgentDir, filePath);   // <-- not canonicalized

claudeRoot is canonicalized (so .claude inside /var/folders/.../agent becomes /private/var/folders/.../agent/.claude on macOS, where /tmp -> /private/tmp and /var/folders/... realpaths under /private/var/folders/...). target is only resolve()d, so an absolute file_path passed in the un-canonicalized /var/folders/... form fails the lexical containment check that compares it to the /private/var/folders/... claudeRoot. The result is that legitimate writes inside .claude get rejected on any macOS dev machine using a tmpdir under /var/folders/...

The fix: canonicalize target the same way as claudeRoot, one line:

  const target = canonicalizePath(resolve(canonAgentDir, filePath));

Verified: tests/unit/hooks/hooks.test.ts 42/42 passing, full unit suite 1109/1109 passing.

This bug is upstream — it ships with PR #594 (hooks/permission-gate hardening) merged in the 2026-06-09 batch. Likely passed upstream CI because Linux runners don't have /tmp -> /private/tmp symlinks. The fix should also be sent upstream so future macOS contributors don't hit it.

## Checklist

- [ ] `npm run build` passes
- [ ] `npm test` passes
- [ ] No new secrets or credentials committed
- [ ] **Agent Awareness:** if this adds a command, endpoint, hook, or behavior change — updated the relevant `templates/*/CLAUDE.md` template(s)
- [ ] **Migration Parity:** if this changes agent-installed files (hooks, settings.json defaults, CLAUDE.md sections) — existing agents will receive the change on next restart, not just new agents via `init`


drzachconner (approved):
Full diff reviewed. Fixes isClaudeDirOperation on macOS: canonAgentDir is already canonicalized via realpathSync, but target was not, so a path under /var/folders/ (which realpaths to /private/var/folders/) would fail the lexical containment check against the canonical claudeRoot. One-line fix: canonicalizePath() applied to target too. No logic change for non-symlinked paths. Correct, minimal, macOS-critical. Merge.

Issue #315, opened 2026-05-05
security: pm2 restart --update-env silently discards ecosystem.config.js env block — sandbox isolation vector 2


grandamenium:
## Summary

Companion to #313 (sandbox isolation vector 1, CTX_FRAMEWORK_ROOT pin). On 2026-05-05, a second sandbox isolation breach was caught during PR #292 review. The mechanism is distinct from #313 and adds a second tripwire requirement: `pm2 restart cortextos-sandbox-daemon --update-env` silently DISCARDS the literal `env:` block from `ecosystem.config.js` and applies the current shell environment instead.

## Sequence (PR #292 sandbox run, 2026-05-05)

1. cortext-designer wrote `sandbox.ecosystem.config.js` with literal env block including `CTX_FRAMEWORK_ROOT: "$SANDBOX"`, `CTX_INSTANCE_ID: "pr-sandbox"`, `BOT_TOKEN: "SANDBOX_NO_TELEGRAM"`.
2. Initial boot via `pm2 start sandbox.ecosystem.config.js --update-env` — env applied correctly, daemon enumerated only `sandbox-agent-1`.
3. After `add-agent sandbox-agent-1`, the harness ran `pm2 restart cortextos-sandbox-daemon --update-env` to pick up the new agent.
4. The restart silently dropped the ecosystem.config.js env. The shell at the time had no `CTX_FRAMEWORK_ROOT` exported (no pre-boot hygiene), so pm2 god-daemon's inherited env (`CTX_FRAMEWORK_ROOT=/Users/cortextos/cortextos`, the live install) took effect.
5. Daemon re-enumerated 13 LIVE agent configs (boris, paul, donna, nick, data, sentinel, matttallybot, tallybot, skoolio, goggins, test-hermes, cortext-designer, plus sandbox-agent-1) with real per-agent `BOT_TOKEN` values from disk. Agents reached `Bootstrap complete. Beginning poll loop`.
6. cortext-designer caught the breach in their own log inspection, stood down, and ran `cleanup_sandbox.sh`.
7. James phone ground-truth: NO Telegram messages landed during the ~3-minute window. No outbound this time. (Vector still real — credentials in process, live config reach.)

## Why this is distinct from #313

- **#313 (vector 1):** root cause was `CTX_FRAMEWORK_ROOT` not declared anywhere in the boot block, so the daemon used the god-daemon's inherited value silently from the start.
[...]

PR #614, opened 2026-06-09
fix(daemon): assert CTX_ROOT matches instance at start, fail closed

grandamenium:
## Summary
Refuse daemon boot when `process.env.CTX_ROOT` (or `CTX_ROOT_LOCK`) disagrees with `join(homedir, ".cortextos", instanceId)`. Closes the class where `pm2 restart --update-env` re-evaluates ecosystem env in the calling shell and silently re-points the daemon at a different state root (#315).

- New pure exported `checkCtxRootConsistency()` in `src/daemon/index.ts` — no side effects, unit-testable.
- Check runs in `start()` after the CTX_FRAMEWORK_ROOT guard and **before any state write** (PID file, state dirs).
- On mismatch: loud FATAL log + best-effort operator Telegram alert (reuses `getOperatorChatCreds`) + `exit(1)`.
- `CTX_ROOT_LOCK`, when set, is the authoritative pin; unset env vars keep current derive-from-instanceId behavior unchanged.

## Test Plan
- [x] Unit: `tests/unit/daemon/ctx-root-assertion.test.ts` 7/7 (match, both-unset, empty-string, wrong-CTX_ROOT fails, lock authoritative, lock-OK, lock+divergent-CTX_ROOT fails)
- [x] Live acceptance: built `dist/daemon.js` with deliberate wrong shell `CTX_ROOT` (instance `pr-sandbox`, `CTX_ROOT=~/.cortextos/default`) refuses to start with exit 1 before any state write; matching root proceeds
- [x] Full `npm test`: 1685 pass / 6 fail, all 6 pre-existing on clean main (cron-scheduler x4, macOS hooks-symlink, phase5); zero failures in touched files
- [x] Independent re-verify on clean `cdd8fc61` worktree: build PASS, 7/7 unit


drzachconner (approved):

Issue #306, opened 2026-05-04
cron double-fire: session CronCreate + daemon cron-state.ts run independently, no cross-source dedup


chickbundy-dot:
## Summary

Cron prompts defined in `config.json` can fire twice in close succession (~30 min apart) when both the in-session scheduler (CronCreate / `/loop`) and the daemon-side config.json runner are active simultaneously.

## Observed

- `morning_priorities` (boss) + `daily_metrics_digest` (analyst) both fired twice on 2026-05-04 ~05:00Z and ~05:30Z
- Pattern: one fire from session CronCreate, second from daemon config.json cron runner

## Root Cause

Two independent schedulers exist with no cross-source dedup:

1. **Session scheduler** — agent calls `CronCreate` (via `/loop`) on session start, fires prompts in-session
2. **Daemon config.json runner** — `cron-state.ts` reads `config.json` and fires cron prompts independently via bus

`cron-state.ts` only tracks `last_fire` timestamps for gap-detection (GH #303). It has no awareness of active CronCreate session jobs. `MessageDedup` in `agent-process.ts` handles message-level dedup but not cross-source scheduler dedup.

## Proposed Fix

Before the daemon config.json runner fires a cron prompt, check if an active CronCreate session job already covers that prompt. Options:
- Add a registry of active session crons that the daemon runner can query
- Have the daemon runner skip config.json entries that have fired recently (< 0.5× interval) per `cron-state.json`

## Priority

Low — cosmetic duplicate firings, no data corruption. Backlog.

PR #616, opened 2026-06-10
fix(daemon): serialize per-agent PTY injects (rebase of #590)

### LINKED ISSUE #510: PTY injection race: non-atomic PASTE+ENTER drops simultaneous-fire crons silently while reporting fired

grandamenium:
## Summary
Rebase of #590 (author credit preserved: @thehoff) onto current main — the original is conflicting after the #592/#596/#598 security campaign landed. Closes #510 (simultaneous cron fires silently dropped: non-atomic PASTE+ENTER collision).

- `injectMessageDetailed` serializes via a per-agent `injectChain` ordering barrier (predecessor failure ignored for ordering, barrier always released).
- `injectMessage` resolves `Promise<boolean>` after the deferred ENTER; resolves false (never rejects) on failure.
- Write closure throws on torn-down PTY and on `lifecycleGeneration` change — teardown/restart mid-inject yields `INJECT_FAILED` instead of false success or stale bleed into a fresh session.
- `MessageDedup.forget()` rolls back the hash on `INJECT_FAILED` so a cron retry is not swallowed as duplicate.
- `handleFireCron` reserves the cooldown slot before async inject, rolls back on failure.

Rebase resolutions (2):
1. Textual conflict in the fast-checker urgent-signal block — composed main`s `wrapFenceSafe` (#592 security) with the PR`s async submit-check.
2. Semantic conflict caught by tsc: a main-added callback-inject used `injectMessage` synchronously (now `Promise<boolean>`, TS2801) — fixed with `await` in the already-async fn, preserving "answer callback only if inject submitted". All `injectMessage` call sites audited.

## Test Plan
- [x] PR`s serialization tests 5/5, incl. "second PASTE waits for first" (two simultaneous fires both processed) and "INJECT_FAILED not false success on torn-down PTY"
- [x] tsc clean, build pass, same diffstat as original (+290/-92)
- [x] Daemon unit suites 328/332 — only fails are the 4 pre-existing cron-scheduler reds (scheduler double-fire, #306-class, untouched by this PR; identical on clean main)
- [x] Full suite scrubbed: 6 pre-existing fails identical to base, zero new
- [x] Sandbox E2E: two same-minute crons on one agent via the REAL CronScheduler + REAL injectMessage — both processed, zero PASTE/ENTER interleave; pre-fix simulation proves collision detection is non-vacuous; torn-down PTY yields INJECT_FAILED. Landed as permanent regression guard: tests/integration/two-cron-collision-e2e.test.ts (3/3)


drzachconner (approved):

PR #618, opened 2026-06-10
fix(telegram): 25s long-poll + split poll-error backoff

grandamenium:
## Summary
Two changes to the same poller path (A:F-08, C#1 — 131 poll errors/5 days in production with two conflated failure modes):

- **25s long-poll:** `getUpdates(offset, 1)` → `getUpdates(offset, 25)` — ~25x cut in idle Telegram traffic per bot, latency unchanged (long-poll returns immediately on arrival). Required a companion change: `TelegramAPI.post()` hard-aborts every fetch at 15s, which would have aborted a 25s long-poll every idle cycle (a regression, more error traffic). `post()` gains an optional `timeoutMs` param (default 15000 — all other call sites byte-identical) and `getUpdates` scales its abort to poll-timeout+5s.
- **Split poll-error handling:** new pure `planPollError()` — 409 Conflict self-dies (second poller), 429 honours `retry_after` with real backoff, network `fetch failed` gets exponential backoff 1s→60s cap with a circuit log on the 5th consecutive failure; per-cycle 0-500ms jitter de-syncs the fleet; failure streak resets on any cycle that reaches Telegram.

## Test Plan
- [x] Poller suite 15/15 (7 new: retry-after honoured, expo backoff progression + cap, circuit log threshold, jitter bounds, streak reset, 409 self-die, long-poll abort headroom)
- [x] All telegram suites 92/92 (independently re-verified on clean cdd8fc61 worktree)
- [x] Full scrubbed suite: 6 pre-existing fails identical to base, zero new
- [x] tsc clean, build pass


drzachconner (approved):

PR #619, opened 2026-06-10
fix(lifecycle): restart-marker hygiene — consume, sweep, hard-kill

### LINKED ISSUE #202: bus hard-restart is a no-op: writes markers but never kills PTY
## Symptom

`cortextos bus hard-restart <agent>` writes the restart markers (`.force-fresh`, `.restart-planned`, `.handoff-doc-path`) and a `restarts.log` line, but the PTY keeps running. The agent enters a loop: hit Tier 2 → write handoff → call hard-restart → announce "Context window full…" to Telegram → process never dies → context keeps climbing.

## Evidence

Donna, 2026-04-21 UTC, session `186cbad8` (unchanged throughout):

```
[2026-04-21T02:18:45Z] HARD-RESTART: context handoff
[2026-04-21T02:31:33Z] HARD-RESTART: context handoff
[2026-04-21T02:46:05Z] HARD-RESTART: context handoff
```

Three log entries, three handoff announcements in `outbound-messages.jsonl`, zero actual restarts. `context_status.json` shows the same session_id before and after each "restart". Donna ended the night still at 69% context on the original session.

## Root cause

`src/bus/system.ts` `hardRestart()` (lines 78–93) only writes files. It never sends IPC to the daemon. `src/cli/bus.ts` `self-restart` (lines ~609–623) already does the right thing — writes the marker AND sends `IPCClient.send({ type: 'restart-agent', ... })` which kills the PTY. `hard-restart` needs the same IPC call.

## Fix (≈4 lines)

Copy the IPC-send block from `self-restart` into the `hard-restart` branch of `bus.ts` (or into `hardRestart()` itself) so both paths actually terminate the PTY. Markers alone aren't enough — nothing reads them without a restart cycle.

## Why it matters

Every Tier 2 handoff path calls `bus hard-restart`. When the daemon's Tier 3 watchdog isn't the one initiating, the agent self-restart path silently fails and the agent grinds on its original session until native auto-compact or a crash cleans it up. Donna's stuck-loop tonight is one example; any agent that hits Tier 2 before Tier 3 is affected.

## Test plan

- Add a unit/integration test that asserts `bus hard-restart <agent>` results in a new `session_id` in `context_status.json` within N seconds (daemon fixture + PTY mock).
- Manual: force donna or a test agent to Tier 2, verify the next `context_status.json` write has a different session_id.

### LINKED ISSUE #609: hard-restart leaves .restart-planned orphaned after restart completes — healthy agents false-flag as "restarting"
## Behavior

`cortextos bus hard-restart --reason ...` writes `state/<agent>/.restart-planned` (src/bus/system.ts, both write sites). When the restart completes and the new session boots, nothing consumes the marker. The only designed cleanup is `clearEndMarkers()` inside `updateHeartbeat()` (src/bus/heartbeat.ts) — which fires only when the agent itself runs `update-heartbeat`. An agent on a long watch/hold loop may not heartbeat for hours, so the orphan marker persists, and any consumer that derives liveness state from it reports a healthy agent as "restarting".

## Observed timeline (repro, 2026-06-06 UTC)

- 02:11:33Z — context-handoff hard-restart completes (restarts.log: `HARD-RESTART: context handoff at 40%`). New session boots and runs normally.
- `.restart-planned` ("context handoff at 40%") survives — mtime 02:11:33Z, exactly the restart moment.
- 04:27:22Z — agent heartbeat.json flips to `status: "restarting"` (writer external to the agent session — the agent's own stdout shows no update-heartbeat call; the marker is the only "restarting" signal on disk at that point).
- 05:1xZ — orchestrator watchdog fires a liveness escalation at a healthy agent ("restarting since 04:27 but PID predates the marker, started ~02:12, never exited").
- 05:21Z — manual fix: marker deleted, `update-heartbeat active` restores truth.

## Impact

- Watchdogs/orchestrators chase ghosts: healthy agents look stuck mid-restart.
- Worse inversion: an assumed-benign "restarting" status can mask a REAL hang — the marker says a restart is in flight when nothing is.
- The 120s `MARKER_CLEAR_GRACE_MS` window in clearEndMarkers is not the gap; the gap is that cleanup is coupled to a heartbeat the agent may not send for hours.

## Proposed fix

Restart completion should consume the marker atomically: when the daemon spawns the post-restart session (or on first session boot), delete `.restart-planned` once the SessionEnd crash-alert classification window has passed — i.e. move primary cleanup from "next time the agent happens to heartbeat" to "restart lifecycle owner". Keep clearEndMarkers as backstop. Secondary hardening: any consumer deriving status from the marker should ignore markers older than the current process start time.

## Notes

- Same orphan-cleanup class exists for the sibling end-type markers (`.session-refresh`, `.user-restart`, ...) listed in END_TYPE_MARKERS.
- The crash-alert double-firing constraint (marker read twice, hence read-without-consume) is why the hook can't consume it — that design is fine; the missing piece is a lifecycle-owner cleanup after both firings land.

grandamenium:
## Summary
Kills the orphaned-restart-marker class that took four agents down for 7-14.5h on 2026-06-09/10 (closes #609, closes #202; six live incident instances documented in the overnight forensic report).

- **Consume:** `enable-agent` consumes stale `.user-disable`/`.user-stop`/`.restart-planned` lifecycle markers before daemon start — a disable→enable cycle now boots clean with zero manual marker removal (the failed-recycle mode that killed paul twice).
- **Sweep:** `discoverAndStart()` boot-sweeps lifecycle markers older than the 300s hook TTL across all agents; within-TTL markers are left for the heartbeat-grace path (observed consuming at ~252s); `.daemon-crashed` excluded (owned by crash logic). Each clear logs a structured `agent=/marker=/age=/content=` line — the observability hook for the class.
- **Orphan reap:** a `.restart-planned` orphaned by crash-before-first-heartbeat (or kept perpetually fresh by a re-arm loop) is reaped at next boot by the sweep; misclassification bounded by the existing 300s TTL.
- **Hard-kill:** PTY wrapper `kill(signal)` now forwards the signal (previously silently dropped — always SIGTERM); `stop()` captures the real pid, waits 8s, then SIGKILLs the **process group** (`process.kill(-pid)`) with pid fallback — reaping descendants that a bare kill orphans. Orphaned PTY descendants are the leading suspect for the posix_spawnp exhaustion that caused the 14.5h outage, so group-reaping addresses the suspected root, not just the symptom.
- Canonical `END_TYPE_MARKERS` list exported from `heartbeat.ts` — no marker-name drift across consumers. The deliberate 120s-grace/300s-TTL design is preserved.

## Test Plan
- [x] 18 new tests: disable→enable clean cycle; sweep stale-vs-in-flight boundary, log line, `.daemon-crashed` exclusion; PGID SIGKILL escalation + pid fallback
- [x] Adjacent suites green: 45/45 across restart-marker-hygiene + lifecycle-markers + agent-manager (independently re-verified on clean cdd8fc61 worktree)
- [x] Full scrubbed suite: 6 pre-existing fails identical to base, zero new
- [x] tsc clean, build pass


drzachconner (approved):

PR #621, opened 2026-06-10
fix(watchdog): per-agent heartbeat-staleness threshold from loop_interval

grandamenium:
## Summary
`collectMetrics()` used a flat 5h staleness threshold for every agent, ignoring the per-agent `loop_interval` the heartbeat record already carries. Any agent dead less than 5 hours read as healthy — in the 2026-06-10 incident, a 1h-cadence agent dead for 3.5h counted toward `agents_healthy` while the fleet was down.

Fix: per-agent threshold derived from `loop_interval` via the existing `parseDurationMs()`:

```
threshold = max( min(3 × interval, 6h), 1.5 × interval, 15m )
```

- 5m cadence → 15m threshold (dead 30m now flags; previously silent for 5h)
- 4h cadence → 6h cap (no one-beat-late false positives)
- 6h cadence → 9h via the 1.5× floor (would otherwise flag at exactly its own beat under the bare cap)
- Missing/unparseable `loop_interval` → conservative 2h fallback, logged once per process (never silent)

New pure module `src/bus/heartbeat-staleness.ts` (`computeStaleThresholdMs` + `isHeartbeatStale`) keeps the logic unit-testable. Scope is heartbeat staleness only; cron-firing health untouched.

## Test Plan
- New `heartbeat-staleness.test.ts` (8): threshold trio for 6h cadence (dead 9.5h = stale / dead 6.5h = healthy via cap-bite guard / on-cadence = healthy), 5m and 4h cadence cases, fallback path, and the incident regression (1h cadence dead 3.5h = STALE, with an anchor asserting the old flat 5h read it healthy)
- Existing `sprint5-metrics.test.ts` staleness suite still green (fallback path) — 36/36 combined


drzachconner (approved):

PR #620, opened 2026-06-10
fix(daemon): verify spawn + bounded retry + fleet alert (spawn-failure truth)

grandamenium:
## Summary
Kills the "Bootstrap-complete lie": node-pty's `spawn()` returns a pty object even when `posix_spawnp` fails (the error surfaces async on the fd; `getPid()` holds a corpse pid), so `start()` declared `Running (pid)` for dead processes and fast-checker's bootstrap timeout then logged `Bootstrap complete` for them. In the 2026-06-09 overnight incident this kept 14 phantom agents "running" for 14.5h with zero operator signal.

Fix at all three broken stages, plus a fleet-wide alert:
- **`start()`**: verify a live pid (`isPidAlive`) before declaring running; bounded retry (3x, 1s/2s backoff); on exhaustion set new registry status `spawn-failed` (truth, not `running`/`crashed`) and record for the alerter. Legit fast exits (`onExit` nulled the pty) unchanged. Immediate pid probe, no awaited timer added to healthy starts.
- **agent-manager**: a `spawn-failed` agent gets no checker/crons/pollers (no zombie poll loop).
- **`fast-checker.waitForBootstrap`**: pid-dead aborts (no false "Bootstrap complete"); pid-alive-but-quiet proceeds as today (legit slow boots preserved).
- **`spawn-failure-alerter` (new)**: one operator alert per failure-class per 15min window naming the affected-agent count (14 failures x 3 retries collapse to a single "14 agents failed (posix_spawnp)" alert, escalating window number on persistence); drained at boot and every 60s.

Converts every future spawn-exhaustion-class outage from hours-silent to one-alert-loud.

## Test Plan
- `npx tsc --noEmit`: 0 errors (verified independently post-apply)
- New tests: `spawn-failure-alerter.test.ts` (6: dedup, affected-count, escalation, classes, format) + spawn-verify cases in `agent-process.test.ts` (2: corpse pid -> `spawn-failed` after retries; live pid -> `running`)
- Targeted suites (alerter + agent-process + codex-app-server + hermes variants): 32/32 pass
- Full env-scrubbed suite: 1686 passed, 6 fails identical to base `cdd8fc61` (zero new)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #623, opened 2026-06-10
fix(daemon): detect restart-into-wedge + exit-reason.json (#339)

grandamenium:
## Summary
A crash-recovery restart re-spawns the pty inside the existing AgentProcess and resumes the persistent poll loop blindly — nothing verifies the agent actually resumes (bootstrap verification runs once per discovery, by design). A restart into a wedged session (e.g. a `--continue` that lands on an interactive prompt and never heartbeats) is silently undetected. Observed live in sandbox soak: an agent sat 29 minutes at idle with no heartbeat after a crash-recovery restart.

Three pieces:
- **Wedge detector (new):** after each crash-recovery restart, require a heartbeat strictly newer than the restart timestamp within 5 minutes. Absent → arm `.force-fresh` + exactly one capped re-restart → still wedged → one `restart-wedged` operator alert (15-min dedup, existing operator-chat path). Healthy resumes are untouched; intentional stops excluded. Watchdog is unref'd.
- **exit-reason.json (#339):** `writeExitReasonSnapshot()` at the crashes.log append site writes a point-in-time `state/<agent>/exit-reason.json` (`{type, timestamp, reason, last_task}`, overwritten not appended) so the dashboard can distinguish planned restarts from crashes without parsing the audit log.
- **Marker-ordering (Area 4.3):** verified already satisfied in base (`sessionRefresh()` writes the marker before stop; codex session-cap rollover routes through the same generic path) — regression coverage only, no code change.

Existing notifyAgents transport tests were env-dependent (passed only with `CTX_FRAMEWORK_ROOT` unset, hiding the execPath branch); they are now env-pinned with both branches asserted explicitly.

Follow-ups once sibling PRs land (intentionally not coupled, to keep this independently mergeable): #620 swap the native alert to the spawn-failure-alerter's class-dedup pipeline; #621 derive the 5-min window from per-agent cadence; #619 add a sweep-interaction test for the armed `.force-fresh`.

## Test Plan
- Acceptance (the live wedge shape): WEDGED → detected within window → force-fresh re-restart (cap=1) → still wedged → ONE alert; RESUMES → no force-fresh, no alert
- #339 suite: schema fields, overwrite-not-append, reason fallback, best-effort on unwritable dir
- Regressions: BUG-011 (stop awaits pty exit), BUG-048 (session timer), CrashLoopPauser, marker lifecycle
- Full targeted suites green in BOTH environments: env-scrubbed and with `CTX_FRAMEWORK_ROOT` set — 43/43 each (verified independently at gate with per-command exit codes)


drzachconner (approved):

PR #624, opened 2026-06-10
fix(daemon): halted-agent midnight crash-budget reset + auto-unhalt

grandamenium:
## Summary
`resetCrashCountIfNewDay` ran only from the crash path — but a halted agent stops crashing, so on a long-running daemon its budget never reset on a new day and it stayed halted until a manual wipe (daemon restarts self-heal; the gap is the no-restart path).

- New non-incrementing `resetCrashBudgetIfNewDay()` used by `start()` and a daily check; writes only when clearing a stale date (no crash-file churn). Existing crash-path function untouched.
- `start()` clears a stale budget up front: first start of a new day = fresh budget.
- Top-of-hour daily timer (UTC-day-crossing guard, `toISOString` day semantics, no TZ logic): a crash-budget-halted agent is auto-unhalted and restarted on day-roll.
- Operator-intent guard: `.user-disable`/`.user-stop` markers mean the budget resets but the agent stays down — never auto-restarted.
- Batched notice: hour-aligned timers tick together, a module-level collector with 10s debounce emits ONE Telegram naming all auto-unhalted agents with a count (never per-agent spam).

## Test Plan
- 4 new tests: start-path stale-date clear; daily-check new-day reset + unhalt + restart; user-disabled halted agent stays down on day-roll; two halted agents produce one batched notice
- agent-process suite 17/17 green in BOTH environments (env-scrubbed and CTX_FRAMEWORK_ROOT set), verified independently at gate with per-command exit codes


drzachconner (approved):

PR #625, opened 2026-06-10
fix(daemon): surface MCP-degraded boot (broken .mcp.json)

grandamenium:
## Summary
On current claude (v2.1.170), a broken `.mcp.json` no longer crash-loops the agent — it boots degraded-but-alive with a `⚠ N setup issues: … MCP … /doctor` warning and silently runs without its MCP tools. That silent degradation was previously invisible to the operator.

New `checkMcpSetupWarningOnBoot()`: post-bootstrap scan of the ANSI-stripped boot stdout tail for the exact captured setup-issues/MCP//doctor warning structure (signature extracted from a live repro, not guessed) → one operator alert naming `.mcp.json` with the `/doctor` hint. Surface-only — no restart or backoff, the agent is alive. Once-per-incident latch clears on the first warning-free boot, so a fixed config re-arms detection naturally.

The originally-specced handleExit crash-loop detector (the 2026-05-14 incident class) is formally closed obsolete-on-current-claude: an isolated repro of both failure classes (malformed JSON, valid-but-unreachable servers) showed claude degrades instead of crashing. If a future claude version reintroduces MCP crash loops, that path can be built then against real crash signatures. Full repro finding included in the delivery notes.

## Test Plan
- 4 new tests: degraded boot → one alert naming .mcp.json; clean boot → zero alerts; false-positive guard (output mentioning mcp + /doctor without the warning structure → no alert); once-per-incident (re-degraded boot does not re-alert, warning-free boot re-arms)
- agent-process + fast-checker suites 67/67 in BOTH environments (env-scrubbed and CTX_FRAMEWORK_ROOT set), verified independently at gate with per-command exit codes


drzachconner (approved):

PR #626, opened 2026-06-11
fix(bus): list-tasks no longer truncates task ids in the table (#587)

### LINKED ISSUE #587: bug(cli): list-tasks table truncates task IDs, causing complete-task/update-task failures
## Summary

The `cortextos bus list-tasks` table display truncates task IDs in the Status column. The JSON output (`--format json`) returns full IDs, but the table display caps them, causing downstream failures when the truncated ID is passed to `complete-task` or `update-task`.

## Failures observed (Jun 3 2026)
- Two wrong-ID failures: `complete-task` crashed with truncated ID; `update-task` returned 'task not found' on a valid task
- Root cause: copy-paste from table display (truncated) vs. from `create-task` output (full ID)

## Reproduction
```bash
cortextos bus list-tasks --agent sentinel
# Status column shows: task_1779195942303_422971 (truncated at display width)
cortextos bus complete-task task_1779195942303_422971  # fails
cortextos bus complete-task task_1779195942303_4229719  # full ID works
```

## Fix
- Table display should show full task IDs (or at least enough chars to be unambiguous)
- OR: task resolution should do prefix-match if the ID is a valid unique prefix
- JSON output (`--format json`) already shows full IDs — table is the only broken surface

## Related
- paul evening-review S1 (Jun 3 2026)
- Two-failure incident logged in sentinel memory

thehoff:
Fixes #587.

`bus list-tasks` rendered the ID column at a hardcoded 26 chars, but task ids are 27 (`task_<13-digit epoch>_<8-digit rand>`), so the last digit was dropped. A truncated id copy-pasted from the table into `complete-task`/`update-task` silently hit the wrong task or none. (`--format json` was already correct — the table was the only broken surface.)

## Fix

Drive the header column and the row from one `ID_W` constant (28 = 27-char id + 1 trailing space so a full id never abuts the Assignee column) and stop truncating (`t.id.padEnd(ID_W)`). The shared constant removes the header/body drift that let the hardcoded 26 diverge from the real id length, so it can't silently re-break.

## Test

Adds an integration regression test: create a task, read the full id back out of the list-tasks **table** (not `--format json`), and confirm `complete-task` accepts it.

Verified: `tsc --noEmit` clean; `npm run build` OK; smoke shows the full id with a gap before Assignee; new test passes.


thehoff:
Note for the maintainer: #464 (by @asachs01) predates this and fixes the same issue (#587) with the same width change — I opened this before seeing it, apologies for the overlap.

This PR additionally carries a shared `ID_W` constant (so the header column and the row can't drift) and an integration regression test. No strong claim that it supersedes #464 — happy for whichever you prefer to merge. Flagging the duplication so only one lands.

drzachconner (approved):
Full diff reviewed. Fixes the list-tasks ID truncation bug (hardcoded 26-char column vs 27-char task IDs). Uses a single ID_W=28 constant to drive both the header and each row, preventing future drift. Integration test creates a task, reads the full ID from table output, and verifies complete-task accepts it. Merge - I hit this bug this session.

PR #627, opened 2026-06-11
fix(bus): manage-cycle errors name the real --cycle flag (#517)

### LINKED ISSUE #517: bug(bus): manage-cycle create error says "requires name" but CLI flag is --cycle
### Description

The `manage-cycle create` command throws an error when `--cycle` is omitted, but the error message references a flag (`name`) that does not exist in the CLI:

```ts
// src/bus/experiment.ts line 474
throw new Error('Cycle create requires name, agent, and metric')
```

The CLI defines `--cycle <name>` (not `--name`) and maps it to `options.name` internally before passing to `manageCycle()`. An operator who omits `--cycle` sees "requires name" and may spend time looking for a `--name` option that does not exist.

### Expected behavior

```
error: Cycle create requires --cycle <name>, agent, and metric
```

The error message should reference the actual CLI flag (`--cycle`) rather than the internal property name (`name`).

### Actual behavior

```bash
cortextos bus manage-cycle create <agent>
# → Error: Cycle create requires name, agent, and metric
```

`--name` is not a valid option for `manage-cycle`. The correct flag is `--cycle <name>`.

### Reproduction

```bash
cortextos bus manage-cycle create <agent-name>
# (omit --cycle)
# → Error: Cycle create requires name, agent, and metric
```

### Evidence

- `src/bus/experiment.ts` line 474: `throw new Error('Cycle create requires name, agent, and metric')`
- `src/cli/bus.ts` manage-cycle handler: `.option('--cycle <name>', 'Cycle name')` and `name: opts.cycle` in the options object passed to `manageCycle()`
- There is no `--name` option defined on the `manage-cycle` subcommand

Tested at `v0.1.1` / commit `23f589574252c81cc6862901ff8b3e654985fbbb`.

### Suggested fix direction

Update the error message in `experiment.ts` to say `--cycle <name>` instead of `name`:

```ts
throw new Error('Cycle create requires --cycle <name>, agent, and metric')
```

thehoff:
Fixes #517.

`manageCycle` validation threw "Cycle create requires name, agent, and metric" (and the modify/remove siblings "requires name"), but the CLI exposes the cycle name as `--cycle <name>` — there is no `--name` flag, so an operator who omits `--cycle` hunts for a flag that does not exist.

## Fix

The three operator-facing messages now name the real flags:

- create: `Cycle create requires --cycle <name>, an agent, and --metric <name>`
- modify / remove: `Cycle <action> requires --cycle <name>`

The issue reported `create`; `modify` and `remove` carried the identical mismatch, so all three are fixed for consistency.

## Test

Adds a regression test asserting each error names `--cycle <name>`.

Verified: `tsc --noEmit` clean; experiment suite 28/28.


drzachconner (approved):

PR #628, opened 2026-06-11
feat(daemon): F1-lite shell-engine executor for crons

fil-up:
## Why

Every cron — including ones whose prompt is just `bash script.sh` — fires by injecting the prompt into the agent's Claude PTY, burning a full model turn for the agent to re-type the command into its Bash tool. On Fable 5 (2.5-3x quota cost vs Sonnet), the pure-script crons cost 144 turns/day fleet-wide: sockeye/site-uptime (48), inari/budget-circuit-breaker (48), inari/rss-monitor (48). A curl check does not need an LLM.

This closes the dead-metadata gap: the `engine` field has existed in config.json since the Ollama-executor era but lost its runtime consumer on 2026-05-17. It now has one again.

## What

- **`src/daemon/shell-cron.ts`** (new): `executeShellCron` — runs the prompt as `bash -c <prompt>`, resolves on exit 0, throws with exit code + stderr tail on failure, 5-min timeout (SIGTERM → SIGKILL escalation). `buildShellCronEnv` — mirrors the PTY env contract (CTX_* vars, org secrets.env, agent .env precedence, CHAT_ID/TZ aliases) so scripts behave identically under either dispatch path.
- **`agent-manager.ts`**: `onFire` branches on `cron.engine === "shell"` → daemon-spawn instead of `injectAgent`. Nonzero exit rides the existing scheduler retry (1s/4s/16s) + execution-log path — failures surface exactly like failed PTY fires, no new alerting surface.
- **`types/index.ts`**: `engine?` on `CronEntry` + `CronDefinition`. Absent/other values = PTY injection (default, unchanged). Historical dead tags like `ollama` are treated as default, not errors.
- **`cron-migration.ts`**: `convertEntry` passes engine through (enabled + disabled paths); config-state-resync treats engine as config-owned like schedule/prompt — add, update, AND removal (config drops field → revert to PTY dispatch).

## Out of scope (by design)

F21 gate_script builds on this spawn path in a separate PR.

## Verification

- `npx tsc --noEmit` clean; build clean.
- **Full suite: 34 pre-existing failures identical on main and this branch (verified side-by-side), +13 new tests all passing, zero regressions.**
- **Keychain-under-daemon-context (review req):** ran the real budget-circuit-breaker via `node child_process.spawn('bash', ['-c', ...])` with only the env `buildShellCronEnv` provides — OAuth keychain read (`security find-generic-password`), usage-API query, and bus writes all work, exit 0. No env-passthrough story needed (daemon runs inside the user login session; keychain is unlocked).

## Rollout

After merge + deploy (rides the next restart window per the deploy quota playbook): set `"engine": "shell"` on sockeye/site-uptime (skipper sign-off), inari/budget-circuit-breaker, inari/rss-monitor — config.json first, then crons.json.

Audit: tezcatlipoca (implementer ≠ sole auditor).


fil-up:
Wrong target repo — internal fleet PR, re-filing against the fil-up fork.

PR #629, opened 2026-06-11
fix(metrics): check-upstream reports local_ahead instead of updates_available when ahead (#31)

### LINKED ISSUE #31: BUG-050: check-upstream reports updates_available when local branch is ahead of upstream
## Summary

`cortextos bus check-upstream` reports `"status": "updates_available"` when the local branch has commits that upstream does not. It should only report updates available when upstream has commits the local branch does not have.

## Repro

1. Commit local changes that are not yet pushed to upstream/main
2. Run `cortextos bus check-upstream`
3. Observe: reports `updates_available` with a diff showing the local-only changes

Expected: `"status": "up_to_date"` or `"status": "local_ahead"`

## Impact

The upstream sync cron fires and reports false updates. Agents waste time investigating non-existent upstream changes. The confusion is low-stakes but noisy.

## Fix

In the check-upstream script, only report `updates_available` if `git rev-list HEAD..upstream/main --count` > 0 (upstream has commits we don't). If only `git rev-list upstream/main..HEAD --count` > 0, report `local_ahead` instead.

thehoff:
Fixes #31 (BUG-050).

## Problem

`checkUpstream` returned `updates_available` whenever `HEAD !== upstream/main`, even when the only difference was **local commits upstream doesn't have**. On a branch that is ahead of (not behind) upstream, the upstream-sync cron then fired on local-only work.

## Fix

When heads differ but `git rev-list HEAD..upstream/main --count` is zero (upstream has nothing we lack), classify as:
- `local_ahead` (with the ahead count), or
- `up_to_date`

…instead of the misleading `updates_available`. The diverged case (HEAD both ahead **and** behind) still correctly returns `updates_available`, since upstream genuinely carries commits we lack.

`cortextos update` gains a matching `local_ahead` handler that exits 0 with a clear no-op message.

## Tests

New `tests/integration/check-upstream-local-ahead.test.ts` drives a real git work-repo + bare upstream remote across four cases:
1. `local_ahead` — HEAD has a commit upstream lacks
2. `up_to_date` — heads equal
3. `updates_available` — upstream genuinely ahead
4. `updates_available` — diverged (ahead AND behind)

All green; `tsc --noEmit` clean.


drzachconner (approved):

PR #631, opened 2026-06-11
fix(heartbeat): quote date format string

noogalabs:
## Summary
Quote the date format string in the Step-5 memory heredoc.

Unquoted `$(date -u +%H:%M UTC)` makes BSD/macOS `date` treat `UTC` as a separate argument, which raises `date: illegal time format` and drops the UTC timestamp from heartbeat memory entries.

This changes the affected upstream template HEARTBEAT.md files to use `$(date -u +'%H:%M UTC')`.

## Verification
- Buggy form: `date -u +%H:%M UTC` -> `date: illegal time format` on macOS/BSD date.
- Fixed form: `date -u +'%H:%M UTC'` -> `14:14 UTC`.
- Scope on upstream main: 4 template HEARTBEAT.md files.


drzachconner (approved):
Full diff reviewed. Adds quotes around the format string in date -u +%H:%M UTC across all 4 template HEARTBEAT.md files. Without quotes, shells that expand the format eagerly can mangle UTC as an env-var lookup. Correct, minimal, no risk. Merge.

PR #632, opened 2026-06-11
fix(catalog): browse-catalog detects on-disk installs, not just the registry (#295)

### LINKED ISSUE #295: bug(catalog): browse-catalog returns installed:false for all items regardless of install state
## Problem

`cortextos bus browse-catalog` returns `installed: false` for all 26 catalog items, even items that are clearly installed and running (e.g., `tasks`, `heartbeat`, `comms`, `autoresearch`).

## Steps to Reproduce

```bash
cortextos bus browse-catalog
```

Output shows `"installed": false` for every item in the catalog.

## Expected Behavior

Items that exist in the agent's `.claude/skills/` directory (or are registered in config) should be flagged `installed: true`.

## Impact

- catalog-browse cron cannot meaningfully distinguish new items from existing ones
- Recommendations logic is broken — everything looks like a candidate for installation
- Agents cannot detect genuine gaps in their skill set

## Likely Root Cause

The `browse-catalog` command probably checks for installation against a registry or path that is not being populated correctly, or the check logic does not walk the actual skills directory.

## Suggested Fix

Cross-reference catalog item names against files present in `.claude/skills/` (or the agent's config skills list) before setting the `installed` flag.

thehoff:
Fixes #295.

## Problem

`browseCatalog` derived `installed` **solely** from `.installed-community.json`, which is written only by `installCommunityItem`. Shipped/default skills (`tasks`, `heartbeat`, `comms`, `autoresearch`, …) and manual copies exist on disk under `.claude/skills/<name>` but were never recorded there — so **every** catalog item reported `installed: false`. The catalog-browse cron's new-vs-existing recommendation logic was effectively broken (everything looked like a fresh install candidate).

## Fix

Extract a shared `resolveItemTargetDir()` — `skill → .claude/skills`, `agent → templates/personas`, `org → templates/orgs` — used by **both** `installCommunityItem` and `browseCatalog`, so the install target and the install-detection path can never drift.

`browseCatalog` now flags `installed: true` when the registry **or** the on-disk target directory says so. The bus CLI threads `agentDir` through (matching `install-community-item`) so skills resolve against the agent root.

Bare `existsSync` is intentional: `installCommunityItem`'s `already_exists` guard uses the same check, so browse and install agree on what "installed" means — requiring a `SKILL.md` marker would make browse say "not installed" while install refuses with `already_exists`.

## Tests

`tests/unit/bus/catalog-installed-detection.test.ts` (8 cases): on-disk skill without registry, registry without disk, neither, agent + org on-disk detection, `agentDir`-undefined fallback, empty-dir characterization, unknown-type registry-only. Existing catalog suites still green (38 total); `tsc --noEmit` clean.


drzachconner (approved):

PR #633, opened 2026-06-11
fix(metrics): approval count agrees with list-approvals — status-filtered, deduped (#193)

### LINKED ISSUE #193: bug(metrics): collect-metrics reports stale approval count
## Description

`cortextos bus collect-metrics` reports `approvals_pending: 10` but `cortextos bus list-approvals` returns only 1 pending approval at the same point in time.

## Reproduction

```bash
cortextos bus collect-metrics | jq '.system.approvals_pending'
# → 10

cortextos bus list-approvals | jq 'length'
# → 1
```

## Observed

`collect-metrics` output:
```json
{
  "system": {
    "approvals_pending": 10
  }
}
```

`list-approvals` output: 1 item (approval_1776678186_4cvnb, created 2026-04-20T09:43:06Z)

## Expected

Both commands should agree on the count.

## Hypothesis

`collect-metrics` may be counting resolved/stale approvals that are not cleaned up from the state file, or reading from a different source than `list-approvals`.

## Impact

Misleading system health dashboard. Could cause false alerts on approval backlog.

## Environment

- cortextOS version: main
- Detected: 2026-04-20 ~13:20 UTC by sentinel during theta wave scan

thehoff:
Fixes #193.

## Problem

`collect-metrics` reported `approvals_pending` by **naively counting `*.json` files** across the root + every org `approvals/pending` dir, while `list-approvals` parsed a single org's dir. Neither filtered on the internal `status` field, so:

- a resolved-but-unmoved approval lingering in `pending/` inflated the metric, and
- the two commands counted different scopes (root + all orgs vs one org),

producing a stale backlog count no `list-approvals` invocation could reconcile — `collect-metrics` showed 10 while `list-approvals` showed 1, firing false approval-backlog alerts.

## Fix

New `collectAllPendingApprovals(ctxRoot)` is the single source of truth: scans the root + every `orgs/<org>/approvals/pending`, keeps only `status === 'pending'`, and dedups by id. Both `collect-metrics` (`approvals_pending`) and `list-approvals --all-orgs` route through it, so they **agree by construction**. (`--all-orgs` also gains the root-level approvals it previously omitted.)

`listPendingApprovals` (single-org) gains the same `status === 'pending'` filter — resolution moves files to `resolved/`, but a crash mid-move can leave a resolved approval in `pending/`, and counting it was the stale count. The sort comparator coerces a malformed `created_at` to `0` so it can never produce `NaN` (bad entries sink to the bottom in descending order).

## Tests

New `tests/unit/bus/approval-count-consistency.test.ts` (8 cases): cross-scope count, stale-status exclusion, id dedup, `collect-metrics` agreement, empty/absent dirs, malformed-date safety. The existing sprint5-metrics "counts pending approvals" test — which asserted the buggy count-any-`{}`-file behaviour — is updated to require genuinely-pending approvals. Full suite green (1856 passed); `tsc --noEmit` clean.


drzachconner (approved):

PR #635, opened 2026-06-12
fix(bus): cycle-level approval_required overrides global in create-experiment

### LINKED ISSUE #634: autoresearch: global approval_required:true overrides cycle-level approval_required:false
## Bug

In `create-experiment`, the top-level `approval_required: true` flag in an agent's `experiments/config.json` overrides per-cycle `approval_required: false` settings. The cycle-level value should take precedence.

## Observed behavior

- Skoolio `experiments/config.json` has top-level `approval_required: true`
- Cycle `first_touch_retention` has `approval_required: false`
- Running `create-experiment` for that cycle auto-creates an approval gate and leaves the experiment in `status: proposed`
- `evaluate-experiment` errors: "is proposed, expected running"
- Intent (no approval gate for this cycle) is silently overridden

## Expected behavior

Cycle-level `approval_required` overrides the global setting. If a cycle opts out, no approval is created.

## Impact

Any agent with a global `approval_required: true` header cannot run cycle-level no-gate experiments without manual intervention per run. Blocks autoresearch automation for cycles that correctly opt out of approval.

## Workaround

Manually resolve the auto-created approval after each `create-experiment` call, then call `run-experiment`. Or strip the global flag entirely.

## Found

TW51 2026-06-12 — skoolio first_touch_retention cycle. Flagged by skoolio, confirmed by sentinel.

grandamenium:
## Summary

Fixes #634. A top-level `approval_required: true` in an agent's `experiments/config.json` silently overrode a matching cycle's `approval_required: false`, auto-creating an approval gate the cycle had opted out of and stranding the experiment in `status: proposed` (`evaluate-experiment` then errors with "is proposed, expected running").

- `ExperimentCycle` now declares `approval_required?: boolean` (the field already existed in real configs, e.g. skoolio's `first_touch_retention`, but the type and code ignored it)
- New exported `isApprovalRequired(agentDir, agentName, metric)` in `src/bus/experiment.ts`: a matching cycle (same metric + agent, identical match rule to `findCycleDefaults`) wins when it sets `approval_required`; otherwise falls back to the global flag; config-read errors resolve to `false`, mirroring `findCycleDefaults` best-effort behavior
- `create-experiment` CLI handler uses the resolver instead of reading only the global flag

Not in scope (flagging, not fixing): `manage-cycle create/modify` does not yet accept an `approval_required` option, so setting it on a cycle still requires editing `config.json` directly. Follow-up if wanted.

## Test Plan

- 7 new unit tests in `tests/sprint3-experiments.test.ts` covering: global-only fallback, cycle false overriding global true (the bug), cycle true overriding global false, cycle without the field falling back to global, metric+agent match requirement, absent config, malformed config
- `npx vitest run tests/sprint3-experiments.test.ts`: 34/34 pass
- Full suite vs clean upstream/main baseline: identical 39 pre-existing environment failures, zero regressions, +7 new passing
- `npm run build` clean


drzachconner (approved):

PR #636, opened 2026-06-12
fix(bus): manage-cycle operates on the target agent's config, not the caller's

grandamenium:
## Summary

`manage-cycle <action> <agent>` resolved its working directory from the **caller's** environment (`env.agentDir`) and ignored the target agent argument entirely. A cycle "created for skoolio" by an orchestrating agent landed in the caller's own `experiments/config.json` — a second registry the target agent's autoresearch loop never reads. The result was a silent no-op: "created" cycles with zero entries in the file the skill actually loops on, no experiment, no error. (Found TW51; root of a multi-cycle autoresearch execution gap. In our deployment the orchestrating agent's config had accumulated 23 cycles targeting 7 other agents, none of which those agents could see.)

- New `resolveTargetAgentDir(env, target)` in `src/utils/env.ts`: resolves the target as a sibling of the caller's `agentDir` first (matches every deployment layout), then via the same `projectRoot` conventions `resolveEnv` uses; validates the agent name as a path-traversal guard; returns `null` when no candidate directory exists
- `manage-cycle` handler resolves the target dir and fails loudly (`exit 1` with a pointer to where cycles live) instead of writing a registry nobody reads; self-targeting keeps the previous `process.cwd()` fallback for bare invocations
- All four actions (create/modify/remove/list) now consistently operate on the target agent's config

Not in scope (flagging): reconciling already-diverged registries is an operational migration (the interim workaround hand-writes the per-agent config), and `manage-cycle create` does not dedupe same-name cycles — both follow-ups if wanted.

## Test Plan

- 7 new unit tests in `tests/unit/utils/resolve-target-agent-dir.test.ts`: caller self-resolution, sibling resolution, projectRoot org fallback, flat-layout fallback, nonexistent target returns null, rootless env returns null, path-traversal names throw
- `npx vitest run tests/unit/utils/resolve-target-agent-dir.test.ts tests/sprint3-experiments.test.ts`: 34/34 pass
- Full suite vs clean upstream/main baseline: identical 39 pre-existing environment failures, zero regressions, +7 new passing
- `npm run build` clean

### DIFF
--- a/src/cli/bus.ts
+++ b/src/cli/bus.ts
@@ -20,7 +20,7 @@ import { nextFireFromCron } from '../daemon/cron-scheduler.js';
 import { queryKnowledgeBase, ingestKnowledgeBase, ensureKBDirs } from '../bus/knowledge-base.js';
 import { checkUsageApi, refreshOAuthToken, rotateOAuth, loadAccounts, ALERT_5H, ALERT_7D } from '../bus/oauth.js';
 import { resolvePaths } from '../utils/paths.js';
-import { resolveEnv } from '../utils/env.js';
+import { resolveEnv, resolveTargetAgentDir } from '../utils/env.js';
 import { IPCClient } from '../daemon/ipc-server.js';
 import { TelegramAPI } from '../telegram/api.js';
 import { logOutboundMessage, cacheLastSent } from '../telegram/logging.js';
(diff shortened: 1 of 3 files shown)


drzachconner (approved):

PR #611, opened 2026-06-07
fix: unblock headless agents on Claude Code 2.1.x (bypass-permissions prompt) + Telegram-optional onboarding

neoturns:
## Summary
Two fixes that let cortextOS agents run **headless** (no human at the PTY) on current Claude Code, and enable **UI-only deployments** without Telegram.

### 1. Auto-accept the new "Bypass Permissions mode" prompt (`src/pty/agent-pty.ts`)
Claude Code **2.1.x** shows an interactive acceptance screen on first launch with `--dangerously-skip-permissions`:

```
WARNING: Claude Code running in Bypass Permissions mode
  ❯ 1. No, exit
    2. Yes, I accept
```

The existing auto-accept only handled the older *"trust this folder?"* prompt by sending a bare **Enter**. On the bypass screen the **default is "No, exit"**, so that Enter makes the agent **exit (code 1)** — agents crash-loop and never onboard.

**Fix:** detect the bypass screen and send **Down-arrow + Enter** to select "Yes, I accept"; keep bare Enter for the trust screen. Detection is ANSI-stripped and uses co-occurring, prompt-specific tokens (`Bypass`+`accept`, `trust`+`folder`), and the poll **stops once the session bootstraps** so no stray keystroke can reach the live session.

### 2. Telegram-optional onboarding (`src/daemon/agent-process.ts`)
When no bot is configured (`telegramApi`/`telegramChatId` are null), the boot and continue prompts no longer instruct the agent to "send a Telegram message" or run the Telegram-driven onboarding protocol. Instead the agent **self-completes onboarding** non-interactively (identity from `IDENTITY.md`, write heartbeat, `touch .onboarded`) and runs from its inbox + the dashboard. Enables fully dashboard-driven deployments; **no behavior change when Telegram is configured.**

## Testing
- `npm run build` compiles cleanly.
- Verified on Claude Code 2.1.168 + Node 22 (WSL): with these patches, fresh agents boot, auto-accept the bypass prompt, self-onboard with no Telegram, and report `status: online` in the dashboard. Without them, agents exit code 1 on first boot.

## Notes
- No new runtime dependencies. Telegram-configured deployments are unaffected.

PR #637, opened 2026-06-12
Preserve VPS OAuth and startup prompt fixes

cgossmann05:
## Summary
- refresh stored Claude OAuth accounts before calling the usage API when the token is within 5 minutes of expiry
- handle Claude Code's startup bypass-permissions prompt by selecting the affirmative option instead of accepting the default "No, exit"
- retry startup prompt handling at 5s, 8s, and 12s for slower first-run prompt rendering

## Why
These fixes were proven on the Avaron VPS after migration:
- `check-usage-api` could fail once the stored Claude profile access token expired, even though a valid refresh token was available in `accounts.json`
- daemon-managed Codex agents could exit during startup when Claude Code rendered the bypass-permissions warning with the default selection on "No, exit"

## Validation
- `npm ci`
- `npm run typecheck`
- `npm run build`

Prepared from a clean worktree based on `origin/main`; the VPS-specific `ecosystem.config.js` change was intentionally left out of this PR.


cgossmann05:
Marking this PR draft for now. It includes the current `src/bus/oauth.ts` refresh behavior, and we have since found a 400 invalid-request-format issue in that refresh path. A corrected OAuth refresh fix is incoming; this branch should be updated with that fix before the PR is ready for review.

cgossmann05:
Updated this draft PR branch with the corrected OAuth refresh implementation.

Fix summary:
- Uses Claude Code’s platform token endpoint (`https://platform.claude.com/v1/oauth/token`)
- Sends the refresh grant as `application/x-www-form-urlencoded` with the Claude Code OAuth client id
- Syncs `~/.claude/.credentials.json` when the cortextOS account was sourced from that file, so the two credential stores do not drift
- Adds unit coverage asserting the endpoint, headers, and form body

Validation:
- Live current-code failure reproduced first: old JSON request returned HTTP 400 `Invalid request format`
- Manual platform/form request succeeded and rotated the refresh token
- `npm test -- tests/unit/bus/oauth.test.ts`
- `npm run typecheck`
- `npm run build`
- `cortextos bus refresh-oauth-token --account claude-profile`
- Forced near-expiry in accounts.json, then `cortextos bus check-usage-api --json --force`; it refreshed and returned usage successfully

Leaving the PR as draft until coordinator confirms ready-for-review.

drzachconner:
## Review - conflict with #611 on agent-pty.ts

Both this PR and #611 patch the bypass-permissions prompt in `agent-pty.ts`. Recommending #611 merges first (it's more comprehensive and adds Telegram-optional mode). After #611 merges, please rebase this PR and drop the agent-pty.ts chunk — the two changes that ARE independently valuable and should survive:

1. **OAuth token refresh before expiry** (`refreshIfExpiringSoon` 5-min window + updating `checkUsageApi`) — this is a real bug where an expiring token causes API failures.
2. **Token URL update** (`https://console.anthropic.com/v1/oauth/token` → `https://platform.claude.com/v1/oauth/token`) — Claude Code 2.1.x moved the OAuth endpoint; the old URL will break.

The OAuth + URL change rebased as a standalone PR would be approved immediately.

PR #640, opened 2026-06-13
feat(templates): add post-restart source-of-truth reconciliation boot step

Honu10:
On cold boot an agent reconstructs working state from MEMORY.md, which can lag a config change — and the drift throws no error, so it surfaces only by luck or a peer catch. Adds a lightweight "On Session Start" step (ADD-ONLY, all 4 agent templates): diff MEMORY.md against the canonical files before first action — model ↔ `config.json`, crons ↔ `crons.json` (the source of truth; `config.json` is only a subset), goals ↔ `goals.json`. Canonical wins on disagreement. Keystone guardrail kept verbatim: never self-apply a change that reverts a dated user-approved decision without explicit approval. Scoped to 3 quick diffs, not a full audit. Validated on one agent first (~107ms cost, caught 3/3 seeded drifts, 0 false positives).


Honu10:
Closing — opened in error; not contributing upstream. Please disregard.

PR #663, opened 2026-06-24
fix(daemon): evaluate cron expressions in UTC, not host-local time

alexyounkin:
## Root cause

Fixed-hour cron expressions fired at the wrong time. `nextFireFromCron` in `src/daemon/cron-scheduler.ts` matched cron fields against the daemon process's **local** wall-clock (`Date.getHours/getDate/getDay`). The host's system timezone is `America/New_York` (UTC-4 / EDT), but the org timezone is UTC and all cron schedules are authored as UTC times.

Result: `morning-brief` = `0 13 * * *` (intended 13:00 UTC = 9am ET) was matched as 13:00 **EDT** = 17:00 UTC = **1pm ET — exactly 4h late**. `evening-brief` = `0 1 * * *` shifted the same 4h. Interval (`4h`) and `*/N` crons never reference a wall-clock hour, so they fired correctly — which is why only the fixed-hour daily briefs drifted.

This is **not** a scheduler eval-cadence problem: the scheduler already ticks every 30s. The lag was purely the UTC-vs-local offset.

## Fix

Evaluate cron expressions in **UTC** (`getUTC*` getters) so fire times are deterministic regardless of the host's system timezone. The minute-walk reference (`startMs`) is already epoch-based, so only the field-matching getters needed changing.

`ipc-server.computeNextFire` reuses `nextFireFromCron`, so the dashboard's next-fire display inherits the fix automatically. `cron-health` only uses interval durations (unaffected).

## Tests

- Updated `nextFireFromCron` tests to assert **UTC** fields (the new contract).
- Added two absolute-instant regression tests that pin `0 13 * * *` → 13:00 UTC and `0 1 * * *` → 01:00 UTC. These **pass even under `TZ=America/New_York`**, proving timezone-independence.
- `cron-scheduler.test.ts`: 30/30 pass (verified under both default and EDT host TZ).
- `npm run build` clean.

### Note on full suite
34 pre-existing test failures (`bus-crons`, `upgrade-cron-teaching`, `bus/hooks`, `hook-crash-alert`, `hooks` symlink) are **environmental** (daemon-not-running, fixture/permission dependent) and **identical on clean `main`**. This change introduces **zero** new failures.

### DIFF
--- a/src/daemon/cron-scheduler.ts
+++ b/src/daemon/cron-scheduler.ts
@@ -36,6 +36,14 @@ import { appendExecutionLog } from './cron-execution-log.js';
 // Cron expression parser — no external deps.
 // Supports: *, */N, comma-lists, and ranges for each of the 5 standard fields.
 // Fields: minute hour dom month dow (day-of-week: 0=Sunday … 6=Saturday).
+//
+// TIMEZONE: cron expressions are evaluated in UTC, NOT the daemon's local
+// timezone. The org timezone is UTC (see SYSTEM.md) and all cron schedules are
+// authored as UTC wall-clock times. Matching against UTC fields makes fire
+// times deterministic regardless of the host machine's system timezone — a
+// daemon running in America/New_York (UTC-4) must still fire "0 13 * * *" at
+// 13:00 UTC, not 17:00 UTC. Interpreting in local time caused fixed-hour crons
+// to fire exactly one UTC-offset late (4h on EDT).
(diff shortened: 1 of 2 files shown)


alexyounkin:
Withdrawing — opened in error. This fix should be landed at the repo owner's discretion, not via an unsolicited fork PR. The change is committed on a local branch and will be routed through the proper decision-maker.

PR #669, opened 2026-06-25
fix(dashboard): auto-populate Max plan usage widget (Claude + Codex) from usage cache

grandamenium:
## Summary

The dashboard **Max Plan Usage** widget reads `state/usage/latest.json`, which is only written by a manual `cortextos bus scrape-usage` paste. Without that one-time step the widget shows "Plan usage tracking not configured" indefinitely — even though the `usage-monitor` cron already maintains a fresh OAuth usage cache at `state/usage/api-cache.json`.

`getPlanUsage()` now falls back to `api-cache.json` when `latest.json` is absent, mapping its `five_hour` / `seven_day` / `seven_day_sonnet` utilization onto the widget's `session` / `week_all_models` / `week_sonnet` fields (friendly reset dates; cache file mtime as the timestamp). `latest.json` stays primary, so the manual scrape path and the daily history series are unchanged — this only fixes the empty default state.

## Test Plan

- Verified the mapping against a real `api-cache.json` (5h 1% → session, 7d 44% → week-all-models, 7d-sonnet 7% → week-sonnet, reset dates formatted).
- `npx tsc --noEmit` (dashboard) → no errors.
- Backward compatible: `latest.json` still takes precedence; absent/invalid cache returns `null` (same empty state as before).

### DIFF
--- a/dashboard/src/app/(dashboard)/analytics/page.tsx
+++ b/dashboard/src/app/(dashboard)/analytics/page.tsx
@@ -13,7 +13,7 @@ import { AgentEffectiveness } from '@/components/analytics/agent-effectiveness';
 import { CostTracking } from '@/components/analytics/cost-tracking';
 import { GoalProgress } from '@/components/analytics/goal-progress';
 import { FleetHealth } from '@/components/analytics/fleet-health';
-import { getFleetHealth, getLatestSnapshot, getPlanUsage, getUsageHistory } from '@/lib/data/reports';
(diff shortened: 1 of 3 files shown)


grandamenium:
Added a second commit: a **Codex Plan Usage** card on the same analytics widget, read from `codex-wham-cache.json` (the usage-monitor's Codex cache). Shows 7d + 5h utilization + plan type, omitted when the cache is absent. Verified mapping vs real cache (pro, 5h 15%, 7d 22%) + dashboard tsc clean.

PR #688, opened 2026-06-30
fix(daemon): inject verbatim live tail + mission anchor on resume

clearworks-ai:
## Problem
Resume prompts force-fed the **stale handoff doc** as an imperative, while the verbatim **live tail** (`conversation-buffer.jsonl`) was only a skippable AGENTS.md bash step — and `loadBuffer()` was dead code (zero callers). The mission file was only used to *defer* restarts, never surfaced or auto-created. Net effect: agents resumed on stale task state instead of Josh's newest instruction.

Root cause (architect-reviewed): **authority asymmetry** between the doc (imperative) and the live tail (optional).

## Fix — 7 changes (spec: `.agent/one-big-feature/handoff-mission-live-tail/`)
1. Wire `loadBuffer()` into **both** `buildStartupPrompt` + `buildContinuePrompt` via a shared `buildResumeContextBlocks()` — `--continue` restarts previously got none of this (the biggest hole).
2. Path roots never mixed: mission = `agentDir/state/current-mission.txt`, live tail = `loadBuffer(ctxRoot, name)`.
3. Precedence encoded in prompt text: **newest inbound > mission anchor > handoff doc**; the mandated "back —" message now derives from doc + newest inbound, newest wins.
4. Auto-derive `current-mission.txt` at handoff-fire time from the trailing **inbound telegram** burst (both Tier-2 handoff + Tier-3 force-restart paths) — never overwrite, never anchor on the agent's own monologue.
5. Handoff-generation prompt now leads with `## LIVE PRIORITY` quoting the newest inbound verbatim, above stale task state.
6. Doc-freshness guard: only adopt handoff docs with `mtime >= ctxHandoffFiredAt`.
7. New `recall-conversation` / `search-transcripts` bus command (scoped by grep/days/limit over buffer + archive).

## Verification
- `npm run build` (tsup) clean; `tsc --noEmit` exit 0.
- Targeted: **42/42 pass** (`agent-process.test.ts`, `context-monitor.test.ts`).
- Full suite: only the **pre-existing** `hooks.test.ts:154` symlink failure is red — confirmed identical on `origin/main` (unrelated to this diff).
- All buffer/mission reads wrapped in try/catch (never throw into the prompt path). No `any`; no new `console.log` outside CLI output.

⚠️ **Not yet restart-proven.** This code only takes effect once merged → `dist` rebuilt → daemon reloaded. A real restart proving an agent resumes on the newest instruction will be captured immediately after merge — will not claim "fixed" before that.


clearworks-ai:
Wrong base — clearworks fork main has diverged from grandamenium upstream; this integrates on clearworks-ai/cortextos:main where the fleet runs. Reopening there.

PR #690, opened 2026-06-30
fix(goal-progress-probe): track codex-runtime agents on long-running in_progress tasks

revopsglobal:
## Summary

Fleet-health infra fix. `hasEngagedOrchTask` now keeps codex-runtime agents in the goal-progress probe whenever any orchestrator-issued task assigned to them is currently `in_progress`, regardless of `created_at` age. The previous `hasOrchIssuedTaskInLast24h` gate silently dropped agents from the probe once their task aged past 24h, hiding actively-engaged long-running work (e.g. a 13-day in_progress assignment) and skipping the stamp-stale classification that would otherwise flag "agent is fine, just re-stamp the goals.json."

Active status is the engagement signal. Creation timestamp alone isn't sufficient. The 24h-since-created window is preserved as a secondary signal so freshly-assigned tasks of any status still register.

## Changes

- `src/bus/goal-progress-probe.ts`: replace `hasOrchIssuedTaskInLast24h` with `hasEngagedOrchTask` (status=`in_progress` ∨ created<24h).
- `tests/unit/bus/goal-progress-probe.test.ts`: explicit `status` in `writeTask` helper; existing "old non-active task → suppressed" case clarified; new regression test for "long-running in_progress task → tracked as stamp-stale, not suppressed."

## Test plan
- [x] `npx vitest run tests/unit/bus/goal-progress-probe.test.ts` → 10/10 pass (was 9, +1 regression case)
- [x] `npm run build` clean
- [x] Live probe verification post-build: 0 stalled, 0 stamp-stale across analyst / blocked-uat-escalator / director / orchestrator
- [x] Confirmed `python3` check that codex has a real in_progress orchestrator task that the new gate keeps tracked

Closes the live-state half of `task_1781658696612_98511388`. Do not merge — opening as review gate per orchestrator direction; merge waits for sign-off.


revopsglobal:
Closing — opened against the read-only public mirror by mistake. Per policy that just landed in CLAUDE.md, RevOps-Global-GIT/cortextos is the integration repo. Reopening the focused probe fix there.

PR #695, opened 2026-07-01
fix(bus): recover orphaned inbox locks + log lock-acquire failures

asachs01:
## Root cause

\`acquireLock()\` treats a \`.lock.d\` whose \`pid\` file is missing or empty/corrupt as "holder mid-acquire — retry", with **no staleness escape**. The PID-liveness stale check only runs after reading a *valid* pid — which a holder killed between \`mkdirSync(.lock.d)\` and \`writeFileSync(pid)\` never wrote. Result: the lock wedges **permanently**, and \`checkInbox()\` compounds it by silently returning \`[]\` on lock failure.

Observed live 2026-07-01: 8 agent inboxes across the \`default\` and \`wyre-gateway\` instances wedged for 2–4 days (~84 stranded messages). The agents kept answering Telegram and crons, so the fleet looked healthy — this was the "inconsistent agent replies" bug. Diagnosis fingerprint: messages stuck in \`inbox/\` (not \`inflight/\`), \`processed/\` mtime frozen, zero \`Injected\` log lines since a daemon restart, \`.lock.d\` present with missing/0-byte pid.

## Fix

- **\`src/utils/lock.ts\`** — \`tryStealStaleLock()\`: the missing-pid and corrupt-pid branches now steal the lock once \`.lock.d\` is older than \`STALE_LOCK_MS\` (30s — ~6 orders of magnitude above the real mkdir→write gap). Fresh mid-acquire locks and live-holder locks are untouched; dead-valid-pid recovery unchanged.
- **\`src/bus/message.ts\`** — \`checkInbox()\` logs a rate-limited warning (once per inbox per minute) when it cannot acquire the lock, so a wedged inbox can never be silent again.

## Tests (TDD — watched fail first)

7 new lock tests: steal on old missing/empty/corrupt pid; NO steal on fresh missing/empty pid (mid-acquire protection); NO steal on old lock held by live pid; dead-valid-pid steal regression guard. 1 new bus test: lock-failure warning + rate limiting.

\`npm run build\` clean. Full suite: 1714 passed; the 6 failures (hooks symlink #18, agent-manager multi-org ×3, add-agent validation, dashboard workflows poll) fail identically on \`origin/main\` — pre-existing, untouched by this change.

## Ops note

The 8 live stale locks were already removed by hand and all stranded messages delivered. Running daemons still execute the old code until rebuilt/restarted.


asachs01:
Superseded by #696 — same fix rebased onto grandamenium:main directly, so the diff no longer drags in the wyre-technology/main divergence (which is what made this one CONFLICTING).

PR #707, opened 2026-07-02
fix(bus): stop truncating task IDs in list-tasks display (conflict-free; supersedes #626)

asachs01:
## What
Stops `list-tasks` from truncating task IDs in the text table. `src/cli/bus.ts` rendered the ID column as `t.id.substring(0, 26).padEnd(26)`, but task IDs grew from 22 to **27 chars** (`task_<13ms>_<8-digit>`, per the createTask same-ms collision fix) without bumping the display column — so the **last digit of every ID was silently dropped**. Operators (and agents) copy the visible 26-char ID into `update-task`/`complete-task`/`save-output`, which then 404 with "not found in any org" because the on-disk filename has the full 27-char ID.

Fix: widen the ID column to 28 (27 max + 1 headroom) and **never substring an identifier in display** (same defensive change applied to the assignee column). Display-layer only — no data drift.

## Provenance / why a new PR
This is a clean cherry-pick of **`a8d318fd`** (authored-by murph, co-investigated-by boss). The prior PR for this fix (**#626**) is currently `mergeable=CONFLICTING` / `DIRTY` against main and cannot merge. This branch is cut fresh off `origin/main` (`cf799265`) — the current code matches the fix's pre-image exactly, so it applies **conflict-free and is immediately mergeable**. Supersedes #626 (close #626 in favor of this, or merge whichever you prefer).

## Risk
Minimal — 1 file, +7/-3, display formatting only. Build & Type Check + Unit Tests pass on the identical change in #626.

## Impact
Clears the fleet-wide `update/complete 404` class that dev, boss, and I all hit this week. Interim workaround (already in use): read IDs from `list-tasks --format json`.

Needs @asachs01 review/merge (CODEOWNERS) — not self-merging.


asachs01:
Closing — wrong base. Branched off the wyre-technology fork main which has diverged from grandamenium upstream, so this shows the whole fork delta (651 files). The fix targets the wyre-technology deploy fork, not upstream. Re-opening there.

PR #723, opened 2026-07-05
fix(dispatch): opencode silent-drop — inject-then-ACK, dedup rollback, retry, dead-letter

clearworks-ai:
## Problem
Opencode dispatch could **silently drop** a queued work item. The ACK/dedup marker was recorded *before* the injection was confirmed, so a failed injection left the item marked handled but never delivered — no error, no retry, gone.

## Fix (self-contained, injection-path only)
- **Inject-then-ACK**: only mark the item handled after injection is confirmed.
- **Dedup rollback**: on injection failure, roll back the dedup marker so the item is eligible again.
- **Retry** transient failures.
- **Dead-letter** after 3 consecutive fails — nothing disappears without a trace.

This is a clean re-cut that drops the earlier draft's reboot-noise coupling; it touches only the injection path.

## Verification (on clean base 06f3388)
- `npm run build` — OK
- `tsc --noEmit` — clean
- affected unit tests — **86/86 pass** (`agent-process-opencode`, `inject`, `fast-checker`)

## Scope
8 files: `src/daemon/agent-process.ts`, `src/daemon/fast-checker.ts`, `src/pty/agent-pty.ts`, `src/pty/inject.ts`, `src/pty/opencode-pty.ts` + 3 matching test files.


clearworks-ai:
Wrong base — recreating against clearworks-ai/cortextos (our origin main).

PR #725, opened 2026-07-06
fix(cli): make dashboard remote access via Cloudflare tunnel work end-to-end

johnreitano:
## Problem

Following the documented flow — `cortextos tunnel start` + `cortextos dashboard` — does not produce a working remote dashboard:

1. `tunnel start` prints `https://<id>.cfargotunnel.com` as the "Dashboard URL", but Cloudflare never serves those addresses directly; a hostname CNAME'd to it is required. Every user following the output gets a dead link.
2. Once a hostname is routed, login fails: NextAuth builds redirect URLs from the `Host` header it sees (`localhost:<port>` behind the tunnel), so successful logins bounce the browser to a dead `https://localhost:<port>/`.
3. The PM2 entry generated by `cortextos ecosystem` runs `next dev`. Dev-mode HMR/streaming stalls behind the tunnel and pages never finish hydrating — the login button sits disabled on "Loading…" because the CSRF fetch effect never runs. (`dashboard.ts` already notes production is required for tunnel use.)
4. If `~/.cloudflared/config.yml` exists from any other tunnel, it silently shadows the `config.yaml` cortextos writes (the launchd service passes no `--config`), and the tunnel serves 404s.

## Changes

- **ecosystem.ts** — PM2 dashboard entry uses `next start` when a production build exists (`.next/BUILD_ID`), falling back to `dev` on fresh installs.
- **dashboard.ts** — generated `.env.local` now includes `AUTH_URL` (precedence: env > `dashboard.env` > `tunnel.json`, ignoring cfargotunnel addresses).
- **tunnel.ts** — new `--hostname` flag creates the DNS route during `tunnel start` and records the real URL; without it, prints honest guidance instead of an unreachable URL. The launchd plist now passes `--config` explicitly.
- **templates/orchestrator** — adds the `Stop` hook (`cortextos bus hook-idle-flag`) that the agent and analyst templates already have; without it the fast-checker cannot detect turn completion for orchestrator agents (same class as the 2026-05-16 analyst incident noted in `doctor.ts`).

## Testing

- `npm run build` clean.
- Full suite green (2013 passed) with these changes applied in a working checkout; on this branch's worktree, the only failing tests also fail on pristine `upstream/main` (dashboard test env / pre-existing), none related to these files.
- Verified live end-to-end: production dashboard behind a named tunnel hostname, login page hydrates, credentials sign-in redirects to the correct public origin.

https://claude.ai/code/session_01PXXy538vLWnZpUe6feJcCU


johnreitano:
Cross-reference for whoever reviews these: this PR's change to `templates/orchestrator/.claude/settings.json` (adding the `Stop` hook) is the same fix I had duplicated in #853. I have removed it from #853 so the two no longer conflict — this PR is the one that carries it.

Context for why it matters beyond the tunnel work: `doctor`'s drift check compares every agent against `templates/agent` (`src/cli/doctor.ts:334`) unless `config.json` records a template, and `templates/agent` does define `Stop`. So every orchestrator scaffolded from the current template fails `cortextos doctor` from birth with `Missing hook block(s): Stop`. This PR fixes that as a side effect.

PR #727, opened 2026-07-06
fix(cli): error boundary so action failures don't crash the process

khxlidhindxwi:
## Problem
`src/cli/index.ts` ended with a synchronous `program.parse()` and **no error boundary**. Any `Error` thrown by an action handler — e.g. `Task <id> not found` from `bus complete-task` / `bus update-task` — escaped as an **uncaught exception**, crashing with a full JS stack trace + `Node.js vX` footer (louder on Node v25) and a non-zero exit. On the happy path the commands work, so it looked intermittent; on the error path a task's status transition would appear "stuck" (reported on the dashboard).

Repro (before):
```
$ cortextos bus complete-task bad-id --result x
/…/dist/cli.js:4004
    throw new Error(
    ^
Error: Task bad-id not found in any org …
    at completeTask (…)  ← stack trace, crash
Node.js v25.5.0
```

## Fix
Switch the entry point to `program.parseAsync(process.argv).catch()`:
```ts
program.parseAsync(process.argv).catch((err: unknown) => {
  console.error(err instanceof Error ? err.message : String(err));
  process.exit(1);
});
```
commander v14 awaits action results, so a synchronous throw in any action surfaces here as a rejection → clean one-line stderr + `exit(1)` instead of a crash. Covers **all** subcommands (incl. `bus complete-task` / `update-task`).

After:
```
$ cortextos bus complete-task bad-id --result x
Task bad-id not found in any org under /…/orgs/
$ echo $?
1
```

## Tests
`tests/integration/cli-error-boundary.test.ts` — spawns the real source via `tsx` (no build step needed, so it runs in the CI test job): unknown-id `complete-task` and `update-task … completed` each assert `exit 1` + a clean `not found` message + **no** stack frame / `Node.js vX` footer. 2/2 pass locally. `npm run typecheck` clean.

## Scope
One entry-point change (+10/-1) + one test file. Low blast-radius; no behavior change on success paths (verified `bus --help` and normal commands unaffected).


khxlidhindxwi:
Closing — this branch was mis-based (cut from a stale local upstream/main ~246 commits behind; diff exploded to 605 unrelated files and did not cleanly contain the intended change). Recreated on a fresh branch off current upstream/main with ONLY src/cli/index.ts + the test. New PR incoming.

PR #734, opened 2026-07-07
fix(hooks): customer-box guard on permission + plan-review Telegram hooks

JBODE-mhhs:
## Summary
- Closes scope (b) of the customer-Telegram-leak incident: `hook-permission-telegram.ts` and `hook-planmode-telegram.ts` could still post a raw system prompt straight to a customer's own `CHAT_ID` — only `hook-crash-alert.ts` had the `CTX_IS_CUSTOMER_BOX` redirect-or-suppress guard. `hook-ask-telegram.ts` is untouched (legitimately customer-facing by design — it forwards the agent's own question, not a system prompt).
- Extracted the routing decision into `resolveCustomerBoxTarget()` in `src/hooks/index.ts` so it's unit-testable without subprocess spawning.
- Fail-closed decisions on the suppress path are immediate (no 30-min wait): permission-telegram denies, plan-review auto-approves — matching each hook's own existing timeout-fallback direction.

## Design notes from two codex review passes
- `OPERATOR_CHAT_ID` equal to the customer's own `chatId` is treated as unset (fails closed) rather than trusted — a misconfiguration here could otherwise route straight back to the customer.
- The token is always pinned to the primary bot (`allowOperatorBotToken: false`) for these two interactive hooks specifically: Approve/Deny callback queries are only ever received by this agent's own daemon-managed bot poller, which the daemon starts only when primary `BOT_TOKEN`+`CHAT_ID` are both present (`agent-manager.ts`'s `if (botToken && chatId) { telegramApi = new TelegramAPI(botToken) }`). A distinct `OPERATOR_BOT_TOKEN` has no poller listening on it, so swapping it in would silently break the callback path.
- The primary-credential check runs first, unchanged from the pre-existing behavior — redirecting to an operator can never functionally work without a running poller, regardless of customer-box status.

## Test plan
- [x] `npm run build`
- [x] `npm test` (2008 passed, 1 skipped, pre-existing) via the repo's pre-push hook
- [x] New unit tests for `resolveCustomerBoxTarget` covering: pass-through when not a customer box, operator redirect, `OPERATOR_BOT_TOKEN` fallback, fail-closed when no operator configured, non-`"1"` `CTX_IS_CUSTOMER_BOX` values, the chatId-collision guard, and `allowOperatorBotToken: false`
- [x] Two codex cross-model review passes (read-only, scoped diff) — both real findings fixed, third pass confirmed clean


JBODE-mhhs:
Opened against the wrong repo by mistake (gh pr create picked the upstream remote instead of origin/JBODE-mhhs/ZeusOS). Closing — no merge intent into the public repo. Correct PR: see JBODE-mhhs/ZeusOS.

PR #772, opened 2026-07-14
fix(bus): a locked inbox must never look like an empty one

libanh824-prog:
An agent went **silently deaf for three days**. A hard shutdown killed a process between `mkdirSync` and `writeFileSync` in `acquireLock`, leaving an abandoned lock with an empty pid file. Two defects compounded:

1. **`lock.ts` could never recover it.** A lock whose pid is empty/corrupt/absent was refused forever, deferring to "a future stale-detection pass" that can never run — a dead process never writes its pid.
2. **`checkInbox` returned `[]` when it could not ACQUIRE the lock** — the same value it returns when the inbox is genuinely EMPTY. Every caller (fast-checker poll loop, CLI, heartbeats) reads `[]` as "no messages."

So the inbox reported clean and empty while **26 messages queued behind the lock**, including an urgent approved dispatch. The failure was on disk the whole time and rendered nowhere anyone looks.

## Fixes

- **Age-bounded recovery** for an unreadable-pid lock (`LOCK_STALE_MS`, 30s). A live holder occupies that window for microseconds, so a legitimate mid-acquire holder cannot be robbed; a killed one self-heals.
- **`checkInbox` throws `InboxLockedError`** instead of returning `[]`. Not-read and nothing-to-read are different outcomes and must not share a representation. CLI exits non-zero; fast-checker logs loudly and still processes Telegram.
- **The steal is now atomic and identity-checked.** Review found the naive `rmSync(force)` + `mkdirSync` steal **broke mutual exclusion** — two processes could both hold the lock, and the same pattern was already present in the pre-existing dead-pid branch. Now: a claim serialises stealers, the lock is judged **in place** (never moved, so it is never absent for a third process to walk into), and an inode check ensures a stale verdict can only destroy the object it actually examined.
- **Tests pin that the inbox is already fenced by `rename()`** — every mutation is a compare-and-swap, so a stale holder's write *fails* rather than corrupting. That is why an epoch token is unnecessary, and why nobody may quietly rewrite a rename into copy+unlink.

Tests cover the exact production deadlock, the mid-acquire holder that must not be robbed, live/dead pid behaviour, claim recovery, and `checkInbox` throwing rather than reporting empty. Verified RED against the buggy versions.

### DIFF
--- a/src/bus/message.ts
+++ b/src/bus/message.ts
@@ -4,9 +4,34 @@ import { createHmac, timingSafeEqual } from 'crypto';
 import type { InboxMessage, Priority, BusPaths } from '../types/index.js';
 import { PRIORITY_MAP } from '../types/index.js';
 import { atomicWriteSync, ensureDir } from '../utils/atomic.js';
-import { acquireLock, releaseLock } from '../utils/lock.js';
+import { acquireLock, releaseLock, inspectLock, type LockHeldInfo } from '../utils/lock.js';
 import { randomString } from '../utils/random.js';
 import { validateAgentName, validatePriority } from '../utils/validate.js';
+
+/**
+ * Thrown when the inbox lock cannot be taken, so the inbox could not be READ.
+ *
+ * Distinct from an empty inbox on purpose: callers must not be able to mistake
+ * "I could not look" for "there was nothing there". See checkInbox.
+ */
+export class InboxLockedError extends Error {
+  readonly inbox: string;
+  readonly lock: LockHeldInfo | null;
+
+  constructor(inbox: string, lock: LockHeldInfo | null) {
+    const held = lock
+      ? `held ${Math.round(lock.ageMs / 1000)}s` +
+        (lock.pidUnreadable ? ' by an unreadable PID (holder likely died mid-acquire)' : ` by pid ${lock.pid}`)
+      : 'lock released during inspection';
+    super(
+      `Inbox "${inbox}" is LOCKED and was NOT read (${held}). ` +
+      `Messages may be queued and undelivered — this is NOT an empty inbox.`,
+    );
+    this.name = 'InboxLockedError';
+    this.inbox = inbox;
+    this.lock = lock;
+  }
+}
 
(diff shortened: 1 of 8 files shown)


storypixel:
Independent reproduction, on a longer timeline and with a variant worth recording.

Same failure mode, five weeks instead of three days. An agent inbox was poisoned on 2026-06-20 and `check-inbox` returned a clean empty array on every call until 2026-07-25, while **213 messages** accumulated unread on disk. Clearing the lock drained all 213 immediately.

**Variant:** the lock dir had **no pid file at all**, rather than a 0-byte one. The process died between `mkdirSync` and `writeFileSync` before the write started rather than during it. That lands on the `readFileSync` catch branch instead of the `isNaN` branch, so it is a second entry point into the same permanent deadlock. Your age-bounded recovery covers both, which is the right call — worth noting explicitly since a fix that only handles the empty-pid case would miss this one.

**Corroborating the design choice:** before finding this PR I implemented the same approach independently — age-bounded reclaim on an unreadable pid, and I landed on the same 30s threshold for the same reason, that the legitimate `mkdir`→`writeFileSync` window is sub-millisecond so it cannot false-positive on a live holder. Two regression tests were worth having: a *fresh* pid-less dir must still be refused (the interleaving race the original comment warns about), and an old dir with a *live* pid must still be refused (age must not override liveness). Both pass against your threshold logic too.

**On the second defect being the more important one:** agreed, and the field evidence supports it. `checkInbox` returning `[]` is what made this invisible. Every health surface I have reported "inbox empty" truthfully-as-far-as-it-knew for five weeks, so nothing escalated. The lock bug caused the outage; the `[]` return caused the *silence*. Separating not-read from nothing-to-read is the fix that makes the next occurrence loud instead of invisible.

**Downstream cost, since impact is easy to under-weight here:** one agent sat blocked for six days waiting on a decision it had escalated six times. Those escalations were in the queue the whole time. It concluded it was being ignored and stopped asking.

Happy to test a merge candidate against the real poisoned lock if that is useful.

PR #776, opened 2026-07-14
fix(security): isolate Codex fleet credentials

SallyRS:
## Summary

- replace forced danger-full-access with a read-only base and explicit per-seat writable, read-only, credential-deny, socket-deny, environment, MCP, network, and native-search capabilities
- enable native Codex web research only when the reviewed seat sets `codex_web_search_enabled: true`
- keep raw shell networking disabled; native search does not grant browser, app, arbitrary HTTP, environment, or MCP access
- keep built-in local tools and same-task multi-agent delegation available inside the reviewed work plane
- verify the active profile across start, resume, and turns, with bounded MCP readiness and no permissive fallback
- retain the hash-reviewed stopped-seat migration and fail-closed lifecycle, discovery, and atomic-writer protections

## Verification

- root TypeScript typecheck: passed
- dashboard TypeScript typecheck: passed
- production build: passed
- focused changed-surface capability suite: 168 passed
- dashboard suite: 112 passed
- real Codex permission-profile probes: 14 passed, including native search enabled while loopback network remains denied
- full repository suite: 2048 passed, 17 skipped, with one unchanged pre-existing macOS tmp versus private-tmp assertion failure in untouched hook code
- final adversarial review: no remaining P1 or P2 findings

## Scope

Source only at reviewed commit 7a730ec5bfc7e44fd9fb142193d441140436aee7. This PR does not install a package, migrate an installed seat, start or restart an agent, change fleet state, or advance Stage 3. Future activation requires a separate authorized installation plus host-trusted runtime collectors and candidate-bound canaries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,89 @@ Every agent's `config.json` carries an explicit `runtime` field that the daemon 
 
 Pass `--runtime <kind>` on `add-agent` to set it at scaffold time, or edit the field in `config.json` and restart the agent. The default is `claude-code`. Today only `--template agent` (and the alias `--template agent-codex`) supports `--runtime codex-app-server` — pairing the codex runtime with `--template orchestrator`/`analyst`/`m2c1-worker`/`hermes` errors with a clean message until codex variants of those templates ship.
 
+### Codex role-capability profile
+
+Unattended `codex-app-server` agents require an explicit capability profile in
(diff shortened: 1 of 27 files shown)


SallyRS:
Closed because this change belongs on the personal SallyRS/cortextos fork. Superseded by https://github.com/SallyRS/cortextos/pull/2.

PR #778, opened 2026-07-14
fix(cron): the fleet health monitor was blind — a cron dead 30 days reported HEALTHY

libanh824-prog:
## The bug

`cron-health.ts` derives `expectedIntervalMs` by parsing the **schedule**:

```ts
const expectedIntervalMs = Math.max(0, parseDurationMs(cron.schedule) || 0);   // :119
...
if (expectedIntervalMs > 0 && gapMs > WARNING_MULTIPLIER * expectedIntervalMs) { … }  // :158
```

`parseDurationMs` is `^(\d+)(m|h|d|w)$`. **A cron expression does not match** → `NaN` → `NaN || 0` → `0` → the staleness rule is gated behind `> 0` and **never runs**.

Not "warns late". **Never warns.** Compiled the module standalone and ran it — a cron that has not fired **in 30 days**:

| cron | schedule | health |
|---|---|---|
| `heartbeat` | `4h` | **WARNING** ← the only one that works |
| `heartbeat-tick` | `*/27 * * * *` | **HEALTHY** |
| `safetystack-daily-check` | `3 15 * * *` | **HEALTHY** |
| `weekly-security-sweep` | `7 15 * * 1` | **HEALTHY** |
| `safetystack-handoff-check` | `13 14 * * *` | **HEALTHY** |

## Blast radius: 13 of 24 crons across the fleet

Including:
- **`weekly-security-sweep`** — the security monitor.
- **`axeltire-uptime-check`** — watches a **live business site**. If it dies, the site can go down and health reports HEALTHY. The owner finds out from a customer.
- **`stale-heartbeat-check`** — *the monitor that catches dead agents, itself unwatched.* The watcher that can die silently, already in production.

The 11 that *did* work all happen to use duration strings (`4h`, `15m`) — **by luck, not design**.

A monitor that reports healthy because it cannot see is worse than no monitor: a missing one is a known gap; this one **manufactures confidence**.

## Why not just teach `parseDurationMs` about cron expressions

That arms the 2×-gap rule, whose blind window **scales with the period** — the weekly sweep would still be invisible for a **fortnight**. Gap-since-last-fire is the wrong primitive.

The right question is not *"how long since it last ran?"* but **"did it run when it said it would?"** — and the schedule already answers that. Detection latency then tracks the **grace period, not the cadence**: a weekly cron is caught as fast as a 27-minute one. `expectedIntervalMs` stays `0`; no interval is fabricated.

## Why the bug survived: the tests encoded it as correct

```ts
// a cron dead 100 HOURS:
expect(['healthy', 'warning']).toContain(result.state);   // ← CANNOT FAIL
```
That assertion accepts the bug and the fix equally. Its own comment said *"should be healthy (or warning if we add logic later)"*.

And worse — a second test was **named `handles daemon-was-down scenario`** and asserted the monitor reports **`healthy`**. The bug wasn't overlooked; it was written down as a **requirement** and pinned with an assertion, so any real fix would have looked like the regression.

Both replaced with assertions that **fail against the old code**.

## The second death mode: ABSENT, not late

A cron that was never registered has no fire-gap to measure. Every fire-based check reports nothing wrong **forever**, because there is nothing to report on.

Real: `safetystack-handoff-check` was live in `crons.json` and **missing from `config.json`** — a rebuild-from-config would have silently deleted it. Found by hand, which is not a strategy.

`reconcileAgentCrons()` (pure) compares **declared** (`config.json`, the documented source of truth) against **live** (`crons.json`): `declared-not-registered` / `registered-not-declared` / `registered-but-disabled`.

**It runs in the daemon, not in a cron.** A cron that watches crons can die the same silent death it exists to catch. The daemon's own death is *already loud* — if it's gone, nothing fires and the agent goes visibly STALE. **Put the guard where its own failure is impossible to miss.**

Zero coverage is not an all-clear: it returns `examined`, and the daemon warns `examined 0 crons — CANNOT confirm any cron is registered` rather than reporting clean over an empty comparison.

## Verification

- Typecheck clean.
- **Negative controls fire.** Reverting only the source fix, the behavioural tests fail with the *wrong answer* (`healthy` for a cron dead 100h; `healthy` for a daily cron silent 3 days). A liveness check that cannot be shown to **fire** is a vacuous gate — the last place to ship one.
- No false alarms: a cron that *did* run its slot stays healthy; inside the grace window we do not cry wolf.
- **Zero regressions**, measured against a clean `upstream/main` worktree: baseline `14 failed / 28 failed / 1750 passed` → branch `14 failed / 28 failed / 1758 passed`. Identical failure set; +8 passing tests. The 28 failures are pre-existing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,3 +67,6 @@ WINDOWS_INSTALL_REPORT.md
 *INSTALL_REPORT.md
 /PHASE*-REPORT.md
 docs/phase-reports/
+
+# also ignore a symlink/file named node_modules (the dir pattern node_modules/ does NOT — that let PR #778 commit a build-machine symlink)
+node_modules
--- a/node_modules
+++ b/node_modules
@@ -1,1 +1,0 @@
-/home/liban/cortextos/node_modules
--- a/node_modules
+++ b/node_modules
@@ -1,0 +1,1 @@
+/home/liban/cortextos/node_modules
--- a/src/daemon/agent-manager.ts
+++ b/src/daemon/agent-manager.ts
@@ -23,6 +23,7 @@ type LogFn = (msg: string) => void;
  * Manages all agents in a cortextOS instance.
  */
 import { reconcileAgentCrons } from '../utils/cron-registration.js';
+import { logEvent } from '../bus/event.js';
 import { readCrons } from '../bus/crons.js';
 
 export class AgentManager {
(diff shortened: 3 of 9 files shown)


libanh824-prog:
Context for the node_modules symlink removed here: it caused **two** harms, not one.
1. **CI:** `node_modules -> /home/liban/cortextos/node_modules` (a build-machine absolute path, self-referential) does not exist in CI, so Dashboard Build failed on it.
2. **Local dev env:** checking out this branch *materializes* that self-referential symlink over a real node_modules, breaking dependency resolution (`Cannot find module 'commander'`) — it broke the cortextos CLI on checkout and needed `npm ci` to recover.

Fix removes the tracked symlink and adds a **slashless** `node_modules` to .gitignore — the existing `node_modules/` (trailing slash) only ignores *directories*, which is exactly why a symlink *file* of that name slipped through. Do not reintroduce a committed node_modules in any form.

PR #803, opened 2026-07-23
fix(telegram): self-heal poller on persistent network failure

notanothergitaccount:
## Problem

The Telegram poll loop self-healed on a **409 Conflict** (exit → supervisor sleeps 30s → fresh connection) but treated network errors (`fetch failed`, DNS, socket resets, 15s timeouts) as **log-and-continue**. That reuses a wedged keep-alive socket in the process-global `fetch` pool indefinitely, so a transient network blip could strand **all inbound Telegram** until the process was restarted.

Observed as a ~2-day outage (2026-07-22): the daemon stayed "online" while every `getUpdates` threw `TypeError: fetch failed`. It only recovered on a full reboot (fresh process = fresh undici pool). The underlying cause was never diagnosable because the error wrapper stringified the error and dropped `err.cause`, logging bare `fetch failed` with no code.

## Fix

- **`src/telegram/poller.ts`** — after **5 consecutive** network-class failures, exit with `lastExitReason = 'network-self-die'`; any clean poll resets the budget. Added an `isNetworkError()` classifier so application errors (bad token, malformed request) — which a restart won't fix — don't count.
- **`src/daemon/agent-manager.ts`** — supervisor handles `network-self-die`: sleep 30s (lets the wedged idle socket close), then reconnect, **retrying indefinitely**. Unlike Conflict's 5-min give-up, a network outage must never become permanent silence.
- **`src/telegram/api.ts`** — preserve the underlying cause (`ECONNRESET`/`ENOTFOUND`/…) in the error message and chain via `{ cause }`, so the next incident is diagnosable.
- **`tests/unit/telegram/poller.test.ts`** — 4 new tests: threshold self-die, sub-threshold recovery (clean poll resets budget), Conflict path unchanged, non-network errors excluded from the budget.

## Testing

- `npm run build` — clean
- `npm test` — 112 files, 1895 passed, 1 skipped


drzachconner:

PR #807, opened 2026-07-24
fix(daemon): disabled agents no longer respawn via handleExit

fil-up:
## Summary
- Adds `isEnabled()` to `AgentProcess` — reads `enabled-agents.json` fresh each call so runtime disables are caught immediately
- Gates all 3 restart paths in `handleExit()` (planned restart, image-poison recovery, crash backoff): disabled agents now set `status='stopped'` and return instead of respawning
- Fixes the accretion pattern where `kill_disabled()` in `iris-session-cap.sh` had to clean up re-accreted PTYs every 10 min after each `--continue` restart

## Root cause
`agent-manager.ts` already gated initial spawns on `enabled=true`, but `handleExit()` had no such check. Every exit (crash, planned restart, session refresh) unconditionally called `this.start()`, causing disabled agents to respawn indefinitely.

## Test plan
- [ ] Build passes in worktree (`tsup` clean)
- [ ] Disabled agent exits → status=stopped, no respawn in daemon logs
- [ ] Enabled agent crash → still recovers normally
- [ ] Re-enabling an agent → next `startAgent()` call from manager brings it back

## Deploy
Daemon restart at 06:00 UTC 2026-07-25 alongside SWL re-enable window.


drzachconner:

PR #824, opened 2026-07-29
fix(dashboard): stop writeGoals from deleting north_star on every save

ryanbrazzell:
`writeGoals` rebuilt the goals object without carrying `north_star` through, so every save silently dropped it.

**Scope:** `dashboard/src/lib/data/goals.ts` + unit test. Single-purpose, independent of the other open PRs.
**Verified:** unit test covering round-trip preservation, run in an isolated worktree.

---


ryanbrazzell:

PR #826, opened 2026-07-29
fix(lock): close check-then-delete races in release and failed-claim cleanup

ryanbrazzell:
`releaseLock` and the failed-claim cleanup both did check-then-delete on the lock directory, so a stale-lock reaper and the owner could race and delete a lock the other had just legitimately re-acquired.

**Scope:** `src/utils/lock.ts` + `tests/unit/utils/lock.test.ts`.
**Honest caveat:** the tests are single-process. `acquireLock` stores `process.pid` and liveness is `process.kill(pid, 0)`, so a live pid does hold a genuine non-reentrant lock and the contention is real — but this exercises restructuring under contention, not multi-process locking. Further work on this file (release-fallback removal, and an ABA in the held-records cache) is in progress and will stack on this branch.

---


ryanbrazzell:

PR #827, opened 2026-07-29
fix(oauth): usage utilization is percentage points, not a fraction

ryanbrazzell:
Utilization from the usage API is already in percentage points; two independent call sites treated it as a 0-1 fraction, under-reporting usage by 100x.

**Scope:** `src/bus/oauth.ts` + `dashboard/src/lib/quota.ts` (the dashboard carries its own parser with the same bug) + oauth unit tests. One commit per call site.
**Verified:** unit tests. This fix has been running live in our deployment, which is what surfaced the dashboard-side duplicate.

---


ryanbrazzell:

PR #829, opened 2026-07-29
fix(dashboard): getAgentDir resolves to the framework root unconditionally

ryanbrazzell:
`getAgentDir` resolved relative to the process cwd, so the same agent name resolved to different directories depending on where the dashboard was launched from.

**Scope:** `dashboard/src/lib/config.ts` + test pinning both the determinism and the discovery asymmetry it exposed.

---


ryanbrazzell:

PR #830, opened 2026-07-29
fix(dashboard): lock the enabled-agents.json RMW and make agent creation atomic

ryanbrazzell:
Two related races: the lifecycle routes did an unlocked read-modify-write of `enabled-agents.json`, and agent creation did not atomically claim the name, so two concurrent creates could both succeed.

**Scope:** dashboard lifecycle + agents routes, `dashboard/src/lib/enabled-agents.ts`, a file-lock shim, and `src/utils/package.json` marking `src/utils` as an ESM subtree so bundlers can consume `lock.ts`.
**Verified:** concurrent-create race test.
**Honest caveat:** this closes dashboard-vs-dashboard contention only. **No CLI or daemon writer of these files takes a lock**, so dashboard-vs-CLI remains open — lockable, not locked.

---


ryanbrazzell:

PR #831, opened 2026-07-29
fix(cli): lock the read-modify-write of enabled-agents.json

ryanbrazzell:
CLI commands that enable/disable/add agents each did an unlocked read-modify-write of `enabled-agents.json`; concurrent invocations could lose an entry.

**Scope:** `src/utils/enabled-agents.ts` and its CLI callers (`add-agent`, `enable-agent`, `import-agent`, `install`, `start`) + an integration test under real process contention.
**Note:** a missing roster entry means ENABLED, so a lost write fails open — practical impact is a re-enabled agent, not a silently disabled one.

---


ryanbrazzell:

PR #832, opened 2026-07-29
fix(bus): ack-inbox reported success for message ids that do not exist

ryanbrazzell:
`ackInbox()` returned void and silently no-opped when no inflight message matched the given id. The CLI printed `ACK'd <id>` unconditionally and logged an `inbox_ack` event, so a typo'd or stale id produced a confident success plus false telemetry — while the real message stayed inflight and redelivered five minutes later as an apparent duplicate.

**Scope:** `src/bus/message.ts` (`ackInbox` now returns whether it matched) + `src/cli/bus.ts` (reports the miss on stderr, exits non-zero, skips the `inbox_ack` event) + unit tests.
**Verified:** unit tests covering both the hit and the miss path.
**Behaviour change worth flagging:** `bus ack-inbox` now exits non-zero on an unmatched id. Any script that acks defensively with an id it is not sure is live will start seeing a failure exit where it previously saw success. That is the point of the change, but it is a change.

Found while acking a real message with a mistyped id.

---

### DIFF
--- a/src/bus/message.ts
+++ b/src/bus/message.ts
@@ -166,8 +166,12 @@ export function checkInbox(paths: BusPaths): InboxMessage[] {
 /**
  * Acknowledge a message by moving it from inflight to processed.
  * Identical to bash ack-inbox.sh behavior.
+ *
+ * Returns true if a matching inflight message was found and acknowledged,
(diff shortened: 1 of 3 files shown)


ryanbrazzell:

PR #833, opened 2026-07-29
fix(bus): checkInbox must not report a wedged lock as an empty inbox

ryanbrazzell:
`checkInbox` returned `[]` from two states the caller could not tell apart: the inbox was read and held nothing, and the lock was busy so the inbox was never read at all. `acquireLock` is single-shot and declines on ordinary transient contention on the assumption that the caller retries — `checkInbox` never did, so a skipped inbox round needed no genuinely wedged lock at all. Messages sat unnoticed and nothing anywhere said so.

Routes through `withFileLockSync` instead: transient contention resolves by backing off, and a genuinely stuck lock throws rather than impersonating a quiet inbox.

**Scope:** `src/bus/message.ts`, `src/cli/bus.ts`, `src/daemon/fast-checker.ts`, `src/daemon/agent-manager.ts` + 18 tests.

**The two call sites handle the throw deliberately differently**, and the asymmetry is the design, not an oversight:
- **Daemon** catches locally and continues as "no messages", logging an error-category event. The catch must be local: `pollCycle` shifts queued Telegram messages off `this.telegramMessages` into `messageBlock` *before* `checkInbox` runs, so a throw escaping to the poll loop's handler would discard them while already dequeued — trading a silent inbox gap for silent Telegram message loss, which is strictly worse.
- **CLI** exits non-zero and prints nothing to stdout. Printing `[]` would hand every shell caller the same silent outage one level up.

Also bounds the cost of a persistent wedge: a circuit breaker (5 consecutive failures, 30s backoff) stops paying a blocking retry every cycle, since `checkInbox` blocks the calling thread and every agent's poll loop shares one event loop. Outage events are throttled to one per 60s, with suppressed occurrences counted and reported in the next event so the throttle never makes an outage look smaller than it was.

**Verified:** every new test carries mutation evidence — each was run against a mutant built to differ from the real code in one dimension, with the expected failure named before the run. Notably the breaker's next-attempt time is pinned finite and reachable: setting it to `Infinity` wedges an agent's inbox for the life of the process and previously survived the entire suite.

**Depends on #832** — both touch `src/bus/message.ts` and `src/cli/bus.ts`. #832 should land first; this will need a trivial rebase if it does not.

---

### DIFF
--- a/src/bus/message.ts
+++ b/src/bus/message.ts
@@ -4,7 +4,7 @@ import { createHmac, timingSafeEqual } from 'crypto';
 import type { InboxMessage, Priority, BusPaths } from '../types/index.js';
 import { PRIORITY_MAP } from '../types/index.js';
 import { atomicWriteSync, ensureDir } from '../utils/atomic.js';
-import { acquireLock, releaseLock } from '../utils/lock.js';
+import { withFileLockSync, type FileLockOptions } from '../utils/lock.js';
 import { randomString } from '../utils/random.js';
 import { validateAgentName, validatePriority } from '../utils/validate.js';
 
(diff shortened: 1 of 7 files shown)


ryanbrazzell:

PR #834, opened 2026-07-29
fix(oauth): serialize accounts.json read-modify-write to stop cross-account clobber

ryanbrazzell:
`refreshOAuthToken` loaded the **whole** accounts store, awaited the token endpoint, then wrote the **whole** store back. Two refreshes for *different* accounts that overlap on that await both hold a pre-refresh snapshot; the later writer reverts the earlier account to its already-spent `refresh_token`. That account's next refresh then fails and it is stranded with no way back.

The atomic rename in `saveAccounts` made each individual **write** safe but did nothing for the load → mutate → save **sequence** around it.

**Scope:** `src/bus/oauth.ts` + unit tests. Reuses the existing `withFileLockSync` (`src/utils/lock.ts`) rather than adding a new primitive — atomic mkdir, PID-based stale detection, fails closed. A new `withAccountsLock()` helper re-reads `accounts.json` *inside* the lock so callers can never apply a change to a snapshot taken before it. Applied at all three sites with the same shape, not just the reported one: `refreshOAuthToken`, `checkUsageApi`, `rotateOAuth` phase-1 commit.

`refreshOAuthToken` now throws if it cannot persist after a successful refresh. The old refresh token is spent by then and the response holds the only copy of the new one, so reporting success would strand the account and call it a win.

**Limits, stated deliberately:**
- Same-account double-spend (two concurrent refreshes of the *same* account spending one one-time token) is **not** fixed here. It needs the lock held across the network call, which is a real availability tradeoff.
- The lock-application tests assert the lock is **applied**, via a delegating spy on `withFileLockSync`. They do **not** prove cross-process exclusion; that cannot be observed from a single process, where the re-read alone already makes behaviour correct. The test comments say so.

**Verified:** unit tests, every mutation pinned by a test named for it (A: fix removed, B/D/E: lock removed per site, C: stale spread, F: persistence throw removed).

---

### DIFF
--- a/src/bus/oauth.ts
+++ b/src/bus/oauth.ts
@@ -16,6 +16,7 @@ import { existsSync, readFileSync, chmodSync } from 'fs';
 import { join } from 'path';
 import { homedir } from 'os';
 import { atomicWriteSync, ensureDir } from '../utils/atomic.js';
+import { withFileLockSync } from '../utils/lock.js';
 
 // --- Types ---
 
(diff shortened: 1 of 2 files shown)


ryanbrazzell:

PR #835, opened 2026-07-29
fix(oauth): rotateOAuth must not commit a rotation decision that lost the race

ryanbrazzell:
> **Stacked on #834.** This branch contains commit `38ef144` from that PR as well as its own `52af6bd`. Review/merge #834 first; this diff will shrink to just the second commit once it lands.

`rotateOAuth` snapshots `active` and the candidate list, then awaits a token refresh and a preflight before committing phase 1. The locked callback checked only that the **destination** still existed — never that `active` was still the account it decided to rotate *away from*.

So when a competing rotation (or an admin edit) landed inside that window, this call overwrote the newer selection with one chosen before it existed, and appended a `rotation_log` entry describing a from→to transition that never happened. Phase 2 then pushed the superseded account's access token into every agent `.env`, which is the user-visible half of the bug.

**Scope:** `src/bus/oauth.ts` + unit tests (27 → 30).

**Aborts rather than recomputes.** Recomputing the destination inside the lock would select an account this call never refreshed and never preflighted — trading a race that had already resolved correctly for an unverified token on every agent, which is the invariant rotation exists to protect. The winning rotation preflighted its own destination, so declining is safe by construction.

The two `!committed` paths now report distinctly: a lost race reads as "superseded", a missing destination still reads as "disappeared". Collapsing them would make a benign race look like store corruption.

**Behaviour change, scoped deliberately:** the sole caller is the `rotate-oauth` CLI, which prints `rotated:false` as "No rotation needed" and exits 0. No cron, daemon, or alerting path consumes it.

**Verified:** mutation-tested with the prediction named before each run — deleting the guard kills the abort and `.env` tests and leaves the vanished-destination test alive, as predicted; collapsing each message branch kills exactly its own test.

---

### DIFF
--- a/src/bus/oauth.ts
+++ b/src/bus/oauth.ts
@@ -458,7 +458,24 @@ export async function rotateOAuth(
     seven_day_util: current.seven_day_utilization,
   };
 
+  // Everything decided above — rotate AWAY from `currentName`, TO `nextName` —
+  // rests on a snapshot read before the refresh and preflight awaits. If another
+  // rotation (or an admin edit) moved `active` during those awaits, that premise
+  // is void: someone else has already rotated, to a destination THEY preflighted.
+  // Committing here would demote a verified newer selection to a staler one and
+  // append a from->to entry describing a transition that never happened.
+  //
+  // Abort rather than recompute. Recomputing the destination inside the lock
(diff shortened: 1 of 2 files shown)


ryanbrazzell:

PR #836, opened 2026-07-29
fix(cli): lock the org context.json and install/dashboard credential read-modify-writes

ryanbrazzell:
> **Shares `src/cli/add-agent.ts` with #831 — that one should land first.**

Two related unlocked read-modify-writes in the CLI.

**1. Org `context.json`.** `cortextos init` and `cortextos add-agent --template orchestrator` both did an unlocked RMW on `<frameworkRoot>/orgs/<org>/context.json`. Run concurrently, both read the whole object, mutate a few fields, and write the whole object back — the later write wins wholesale. `init` additionally had an `existsSync`-then-write TOCTOU between its create and upgrade branches.

New `src/utils/org-context.ts`: `mutateOrgContext()` does read+decide+write inside one `withFileLockSync` closure, modelled on `src/bus/crons.ts`. Plain reads stay unlocked, matching that precedent. Writes go through `atomicWriteSync` — all three old call sites used bare `writeFileSync`, so readers could observe a truncated file, and every reader of this file swallows parse errors and silently keeps defaults. A malformed `context.json` is backed up and refused rather than overwritten, because both writers are fill-missing-fields passes and treating a parse failure as `{}` would rewrite a recoverable file down to bare defaults and report success.

**2. `install`/`dashboard` credential writes.** `install.ts` did four bare check-then-write sequences on ctxRoot state files, and `dashboard.ts` independently generated and persisted the same credentials. `enabled-agents.json` + `bus-signing-key` now write under one lock on `<ctxRoot>/config`, re-checking `existsSync` **inside** the lock. Clobbering a populated roster with `{}` is not cosmetic: a missing entry means ENABLED (`src/bus/agents.ts:90-91`), so it silently restarts explicitly-disabled agents. New `src/utils/dashboard-env.ts` gives both writers a shared locked, merge-preserving RMW for `dashboard.env` — beyond the race, `install` rewrote that file from a fixed five-key list and so dropped any other key on every run, with no concurrency required.

**Scope:** `src/cli/{add-agent,dashboard,init,install}.ts`, `src/utils/{org-context,dashboard-env,env}.ts` + two real-process contention tests.

**Verified:** both tests spawn real processes, because vitest is single-threaded and an in-process lock is uncontended on every acquire — such a test goes green whether or not the lock exists. Mutation-tested with a **rate**, not a single failure: bypassing `withFileLockSync` makes the context.json arm fail 10/10 runs (`orchestrator: expected "boss", got ""`) and the dashboard.env arm 10/10; restored, 5/5 and 3/3 green. Each has a control arm replicating the old unlocked RMW, so a green cannot mean the detector sees nothing. Suite reconciled against an `a15baad` baseline re-run in the same session with failing test **names** diffed, not just counts.

**Scope limits, stated:**
- This does **not** close CLI↔dashboard. The dashboard resolves the same path and re-exports the same lock module, so the rendezvous is real, but `api/org/config/route.ts` is still an unlocked RMW. Already-installed older binaries are never excluded.
- The dashboard.env test runs the shared helper via `tsx`, not the production binary (`install` runs `npm link`, `dashboard` starts a server), so it does not re-prove that those two call sites route through it. The file header says so.
- The sibling per-agent `config.json` is deliberately untouched: the dashboard's `getAgentDir()` falls back to `<CTX_ROOT>/…` when the framework-root path is missing while the CLI never looks there, so the two sides can resolve *different* directories for the same agent precisely during creation. A lock there would exclude nothing in the window where the race lives. Tracked separately.

---

### DIFF
--- a/src/cli/add-agent.ts
+++ b/src/cli/add-agent.ts
@@ -4,6 +4,7 @@ import { join, resolve } from 'path';
 import { homedir } from 'os';
 import { OrgContext } from '../types';
 import { validateAgentName, validateOrgName } from '../utils/validate';
+import { mutateOrgContext } from '../utils/org-context';
 
 const VALID_RUNTIMES = ['claude-code', 'hermes', 'codex-app-server', 'opencode'] as const;
 type RuntimeKind = typeof VALID_RUNTIMES[number];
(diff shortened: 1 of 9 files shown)


ryanbrazzell:

PR #837, opened 2026-07-29
fix(dashboard): migrate cron CRUD off the config.json fossil and delete the dead route

ryanbrazzell:
Both of the dashboard's cron editors read and wrote the agent's `config.json`. The daemon migrates `config.json` crons into `crons.json` exactly once, guarded by a `.crons-migrated` marker, and **never reads `config.json` again**. So the editors were not merely stale, they were actively misleading:

```
config.json (fossil):  [heartbeat 4h]                      <- what the UI showed
crons.json  (live):    [heartbeat 4h,
                        nightly-review-sbc-edge 0 8 * * *]
```

One plausible-but-wrong cron rendered, the other hidden entirely, and every edit written into a file nothing reads — silently discarded.

**Scope:** three commits, in this order deliberately (migrate first, delete second, so there is never a window where the UI points at a route that is gone):
1. Agent-detail Crons tab → `/api/workflows/crons`.
2. Workflows page "Manage Crons" accordion → same. Its per-agent-fetch-in-a-loop (an N+1) collapses to one GET; `agents` is now derived from that single source, so an edit in the accordion also refreshes the status table above it.
3. **Deletes** `dashboard/src/app/api/agents/[name]/crons/route.ts` (144 lines) once a repo-wide grep finds no remaining caller. Two prose comments that referenced it are updated rather than left stale.

**Three shape changes this forces:** `interval` → `schedule` (legitimately holds either `4h` or `0 8 * * *`; the old `/^\d+[smhd]$/` model would have rejected our own live `nightly-review-sbc-edge` cron); whole-list `PUT` → per-cron `POST`/`PATCH`/`DELETE`, which go through IPC so the daemon owns the write, the lock, and the scheduler reload; and validation reuses `@/lib/cron-utils` rather than reimplementing it. Crons key by name now, so cron name is immutable while editing.

**Verified** against a dev server with a real browser, run from this worktree with the shell `CTX_*` vars unset so config resolution matches the pm2 dashboard. **Control arm first, because the app is behind an auth wall** and a logged-out blank tab is indistinguishable from an inert fix: unauthenticated `/api/workflows/crons` returns 401, and the session was asserted live (`/api/auth/session` returns a user) before reading any UI.

The browser checks are proven non-vacuous by reverting only the changed file and re-running: 21/21 pass → 8 fail → 21/21 pass. The 8 are exactly the 8 predicted in advance. An earlier draft matched text against the whole page body — a weak test, since the already-correct status table above the accordion renders the same strings and those assertions passed pre-fix too. They are kept, labelled `[weak]`, only to show the page rendered; every load-bearing assertion is scoped to the agent card element itself.

The route deletion is verified with an **authenticated** request, because middleware returns 401 before routing and an anonymous probe cannot tell a deleted route from a live one: logged in, the deleted path returns 404, a sibling `/api/agents` route still returns 200, and a made-up path under the same prefix also returns 404 — so the 404 is the route being gone, not the whole prefix failing.

`tsc --noEmit` clean and confirmed non-vacuous (injected type error reproduces failure at exit 2; restored clean).

---


ryanbrazzell:

PR #838, opened 2026-07-29
docs(kb): correct the collection model — two collections, and no --collection flag

ryanbrazzell:
Agent-facing docs across `templates/` and `community/` described a **three**-collection knowledge base (`memory-{agent}`, `private-{agent}`, `shared-{org}`) and instructed agents to pass `--collection memory-$CTX_AGENT_NAME` when ingesting. Both are wrong, and the combination is the damaging part: the flag files an agent's memory under a collection that `kb-query` can never read back.

**Scope:** 29 doc files (`templates/`, `community/`). Doc-only — no source or behaviour change.

**Verified against source, not inferred:**
- `src/bus/knowledge-base.ts:284-286` computes the name as `` `agent-${agent}` `` for private scope and `` `shared-${org}` `` otherwise. There is no `memory-{agent}` and no `private-{agent}` — an agent's memory files and its outputs share **one** collection.
- `src/bus/knowledge-base.ts:162-169` shows `kb-query` only ever searches `` `shared-${org}` `` and `` `agent-${agent}` ``. So anything ingested under a custom collection name is unreachable by the documented query path.
- `cortextos bus kb-ingest` has no `--collection` option at all; passing it exits with `unknown option '--collection'` — the exact string now quoted in the docs.
- `bus/kb-ingest.sh:11,40` **does** accept `--collection` and overrides the name directly.

That last asymmetry is why the old instructions looked like they worked: whichever entry point an agent reached decided whether the flag was a hard error or a silent misfile.

The docs now say two collections, do not pass `--collection` at either entry point, and narrow a search by wording the question rather than by picking a collection. The consequence agents most need is stated explicitly — **you cannot query your memory separately from your outputs**, because they are the same collection.

---

### DIFF
--- a/community/agents/agent/AGENTS.md
+++ b/community/agents/agent/AGENTS.md
@@ -278,25 +278,29 @@ This is the knowledge you have synthesised over time. Not a log — a living doc
 
 Also update GUARDRAILS.md when you identify a pattern of behaviour that should be explicitly prohibited or corrected — not just for yourself but as a guardrail for future sessions.
 
-Update on every heartbeat and at session end. When you update MEMORY.md, ingest it to your `memory-{agent}` KB collection so it is semantically searchable.
+Update on every heartbeat and at session end. When you update MEMORY.md, ingest it to your `agent-{agent}` KB collection so it is semantically searchable.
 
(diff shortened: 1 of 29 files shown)


ryanbrazzell:

PR #840, opened 2026-07-29
fix(dashboard): atomic-write the config.json PATCH

ryanbrazzell:
**Scope:** the dashboard's config PATCH ended its read-modify-write in a bare `writeFileSync`. Swap it for `atomicWriteSync` (temp file + rename), via a re-export shim, and cover it with a test that fails on the reverse mutation. Two commits: the shim, then the route + test.

Why it matters: `writeFileSync` truncates and then refills, so a concurrent reader can observe config.json half-written — and every reader parses it inside a `try/catch` that falls back to defaults. A torn read therefore does not surface as an error, it surfaces as an agent silently running with **default settings**.

**Verified:**
- `dashboard/src/app/api/agents/[name]/__tests__/config-atomic-write.test.ts` — 4 tests, green. Full dashboard suite: 123 passed / 10 files. `tsc -p dashboard/tsconfig.json --noEmit` clean.
- **Mutation-tested, both directions.** Reverting the route to `writeFileSync(... + '\n')` turns the inode test and the mode test red (`2 failed | 2 passed`). Keeping `atomicWriteSync` but letting the caller append its own `'\n'` turns the trailing-newline test red (`1 failed | 3 passed`).
- The decisive test hard-links config.json before the PATCH. A hard link is a second name for the same inode, so an in-place rewrite is visible through it and a whole-file replacement is not — which is exactly the property "no reader observes a torn file" rests on. It carries a control arm (the new value must actually have landed), or "the old inode is untouched" would be trivially true of a handler that wrote nothing.

**Honest caveats:**
- **This PR used to include a lock, and it was cut on review.** The earlier version wrapped the RMW in `withFileLockSync`, justified as making the dashboard's *two* config.json writers mutually exclusive. Tracing it through what #837 actually removes, that justification does not survive: (a) the lock does **not** close an in-process race — this RMW is synchronous and after the handler's last `await`, and Node is single-threaded, so it was already safe; the crons PUT was the genuine in-process hazard precisely because it read *before* its `await request.json()`, and #837 deletes it; (b) it would only exclude other **processes** locking the same directory, and that set is empty today — `src/cli/add-agent.ts:176,193,299` and `src/cli/import-agent.ts:132` are plain `writeFileSync`, and the CLI's crons lock keys on the **state** dir, not `orgs/<org>/agents/<agent>/`. It made config.json lock**ABLE**, not locked. If a real second writer appears, the lock is one `withFileLockSync` wrap away and belongs in that change.
- **config.json goes 0644 → 0600** on its first write through this path: `atomicWriteSync` creates its temp file `0o600` and rename preserves the mode. Nothing in the fleet reads config.json as another user, so this is safe today — it is asserted in the test so it is a recorded decision rather than a surprise, and that test is where it surfaces if that ever changes.
- The guarantee is "no torn read", **not durability**. `atomicWriteSync` does not fsync the file or the directory, so post-crash the rename may not have reached disk. The shim's header says so explicitly.
- The leftover-`.tmp.*` assertion is **vacuously green** under a plain `writeFileSync` — it guards the atomic path's own failure mode, it does not distinguish the two writers. The test file says that inline so nobody counts it as evidence.
- Pre-existing, not touched here: `src/utils/atomic.ts:35` uses `require('fs')` for temp-file cleanup, which is `undefined` under the `{"type":"module"}` marker this crossing relies on. It sits in a swallowing `try/catch` and the original error is rethrown, so the only consequence is a leaked `.tmp.*` on an already-failing write. Worth a separate one-line fix, not this PR.

**Ordering:** branch is based on `fix/dashboard-unlocked-rmw` (#830) — it needs that PR's `src/utils/package.json` (`{"type":"module"}`) and `dashboard/tsconfig.json` target bump for the cross-boundary import. **Land #830 first.** No conflict with #837.

### DIFF
--- a/dashboard/src/app/api/agents/[name]/__tests__/config-atomic-write.test.ts
+++ b/dashboard/src/app/api/agents/[name]/__tests__/config-atomic-write.test.ts
@@ -1,0 +1,144 @@
+/**
+ * The config PATCH must REPLACE config.json, not rewrite it in place.
+ *
+ * The bug this guards is invisible to the obvious test.  A test that PATCHes
+ * and then asserts the resulting contents passes identically against
+ * `writeFileSync` and against `atomicWriteSync` — both leave the same bytes on
+ * disk.  What differs is only observable DURING the write: `writeFileSync`
+ * truncates the target and then refills it, so a concurrent reader can see a
+ * half-written file, while temp-file-plus-rename leaves the old inode intact
+ * and swaps a complete new one into its place.
+ *
+ * So the first test below asserts that swap directly, by hard-linking the file
+ * before the write.  A link is a second name for the SAME inode: if the handler
+ * rewrote the file in place, the link sees the new bytes; if it replaced the
+ * file, the link still holds the whole old contents.  That is exactly the
+ * property "no reader observes a torn file" rests on, and it is what fails if
+ * someone swaps `atomicWriteSync` back for `writeFileSync` — verified by making
+ * that mutation and watching this test go red.
+ *
+ * HONEST SCOPE.  This is a single-process test.  It proves the write is a
+ * whole-file replacement; it does NOT run a reader concurrently with a writer,
+ * and it says nothing about durability — `atomicWriteSync` does not fsync, so
+ * the guarantee is "no torn read", not "survives power loss".
+ */
+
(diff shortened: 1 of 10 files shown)


ryanbrazzell:

PR #841, opened 2026-07-29
fix(lock): releaseLock only ever touches an acquisition we currently hold

ryanbrazzell:
**Based on #826** (`fix/lock-stale-detection` @ `e571bf1`) — that commit introduces `heldRecords`, which this one fixes. Review #826 first; against `main` this PR shows both commits, only `3631b22` is new.

## Scope

`heldRecords` is meant to answer "do we hold this lock, and with what record". Both halves were broken.

**1. A map miss fell back to `?? ownerRecord()`.** Releasing a lock we never took recomputed an identity and proceeded anyway — taking custody of a healthy *foreign* lock: renamed it aside, read it, renamed it back. That enters the restore window documented on `removeWithCustody`, where a third acquirer's fresh claim can be destroyed, and it logs `data guarded by it may be torn. Investigate` — a false alarm someone has to chase. Recomputing was never sound as identity either: `readStartTime` can fail transiently, so the recomputed record need not match the one we wrote. A miss is now a no-op.

**2. The entry survived a failed release**, so the map named a *process* rather than an *acquisition*. The entry is now cleared as soon as the acquisition ends.

Clearing it on *every* failure would be a new bug, so `Custody`'s `lost` is split:

| outcome | condition | do we still hold it? | entry |
|---|---|---|---|
| `vanished` | ENOENT | no — path is free | cleared |
| `stuck` | any other errno | **yes** — dir still there, still carrying our record | **kept** |

`stuck` must keep the entry. Dropping it would make every later release the no-op from (1) and **wedge the lock for the life of the process**, since a live holder is judged alive forever.

## Verified

`tsc --noEmit` clean. 32/32 (`lock.test.ts` 28 + 4 new).

**By mutation.** The 28 existing tests in `lock.test.ts` pass both before and after this change and went red on **none** of the five mutations — they are not evidence for it. Each mutation is caught only by the new file:

| mutation | tests that go red |
|---|---|
| restore `?? ownerRecord()` | 3 — no-op / duplicate / vanished |
| clear on every outcome (`if (true)`) | 1 — **STUCK** |
| clear only on `'removed'` (old behaviour) | 2 — duplicate / vanished |
| split collapsed to always `'stuck'` | 1 — **VANISHED** |
| split collapsed to always `'vanished'` | 1 — **STUCK** |

The last two pin each direction of the split independently.

An earlier draft of the vanished test re-acquired the lock to show the path was free. It **passed even when the record was wrongly kept** — re-acquiring repopulates the map with a byte-identical record, so cleared-or-not was unobservable. It now hands the freed path to a foreign holder instead, so the entry's absence is what the assertion rests on. That draft is why the "always `'stuck'`" row above exists.

The STUCK test forces EACCES by `chmod 0o555` on the parent dir, and asserts it actually hit that path — as root the chmod would not bite and the test would otherwise pass vacuously.

## Caveats — what this does NOT close

**Double-release is not safe in general.** If the process **re-acquires** the same path, `claimLock` writes a byte-identical record and repopulates the map, so a delayed release still matches and removes the new acquisition. Closing that needs a **generation nonce** in the record — a format migration, since the record is parsed by a fixed regex (`/^(\d+)(?: (\d+))?$/`, ~line 427). Deliberately not attempted here; worth its own issue.

The invariant actually established is the narrower **"`releaseLock` only ever touches an acquisition this process currently holds"**, and no more.

**A note on the filed task text.** The task this came from prescribed "clear `heldRecords` on declined/lost release". Taken literally that is *harmful*: `lost` covered two opposite situations, and clearing on a non-ENOENT failure is exactly the wedge described above. Hence the split. Flagging it because the prescription, not just the bug, needed correcting.

`removeWithCustody`'s decline path still does an unverified `renameSync(tmp, lockDir)` restore — pre-existing, untested, out of scope here.

---

### DIFF
--- a/src/utils/lock.ts
+++ b/src/utils/lock.ts
@@ -1,21 +1,397 @@
-import { mkdirSync, rmdirSync, writeFileSync, readFileSync, rmSync } from 'fs';
+import { mkdirSync, writeFileSync, readFileSync, rmSync, renameSync, statSync } from 'fs';
 import { join } from 'path';
+
+/**
+ * How long a lock directory may sit with NO pid file before we treat it as
+ * abandoned.  The window between `mkdirSync` and `writeFileSync` in
+ * `claimLock` is sub-millisecond, so 30s is ~30,000x the exposure: for this to
+ * steal from a live holder, that holder must freeze *precisely* inside the gap,
+ * stay frozen past the threshold, and then resume.
+ */
+const PIDLESS_STEAL_AFTER_MS = 30_000;
+
+/**
+ * How long a lock may be held by a PID we can see but cannot *identify* before
+ * we treat it as abandoned (no /proc, or a legacy pid-only record).  Longer
+ * than PIDLESS_STEAL_AFTER_MS because we have no corroborating evidence at all
+ * here — only the passage of time.  Every caller holds this lock for a
+ * synchronous JSON read-modify-write measured in milliseconds, so a five-minute
+ * tenure is already pathological.
+ */
+const UNVERIFIABLE_STEAL_AFTER_MS = 300_000;
+
+/**
+ * Read a process's start time (field 22 of /proc/<pid>/stat, in clock ticks
+ * since boot).  Together with the PID this forms an identity that survives PID
+ * reuse: the OS can recycle the number, but not the number *and* the start
+ * tick.
+ *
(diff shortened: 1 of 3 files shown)


ryanbrazzell:

PR #846, opened 2026-07-30
fix(bus): filter .DS_Store and non-agent entries from list-agents

fil-up:
## Summary

- `cortextos bus list-agents` was returning `.DS_Store` and any non-directory filesystem entry as an `enabled=true` agent
- Root cause: `src/cli/bus.ts` scanned `orgs/*/agents/` with no name filter and no `isDirectory()` check
- Also: the `enabled-agents.json` loop had no name validation (exploitable if the file is corrupted or hand-edited)

## Changes

Added `AGENT_NAME_RE = /^[a-z0-9_-]+$/` constant and `statSync().isDirectory()` guards at both scan sites:

1. **`enabled-agents.json` loop** (line ~1390): reject keys that don't match agent name pattern
2. **`orgs/*/agents/` filesystem walk**: reject non-directory org entries AND non-directory/non-name agent entries

Belt + suspenders: regex filter catches `.DS_Store`, numeric names, dotfiles; `isDirectory()` catches symlinks and files.

## Test plan

- [x] `npm run build` — clean
- [ ] `npm test` — 35 pre-existing failures (bus-crons, dashboard, hooks, integration) unrelated to this change; verified same failures on main


drzachconner:

PR #853, opened 2026-07-31
fix: two fresh-clone correctness bugs (macOS .claude guard, pre-push gate report)

### LINKED ISSUE #852: Three fresh-clone correctness bugs: macOS .claude hook guard, false pre-push gate report, orchestrator template missing Stop
Found while setting up a fresh clone and a new org. All three are independent; happy to split into separate PRs if preferred.

## 1. `isClaudeDirOperation()` fails on every macOS host

`src/hooks/index.ts` canonicalizes the agent dir but compares it to a target resolved only lexically. When the install path contains a symlink, the two are in different namespaces:

```
base            /var/folders/xy/T/hookperm-abc/agent
canonAgentDir   /private/var/folders/xy/T/hookperm-abc/agent   (realpath)
claudeRoot      /private/var/.../agent/.claude
file_path       /var/folders/.../agent/.claude/settings.json   (absolute)
target          /var/folders/.../agent/.claude/settings.json   (resolve() ignores
                                                                the base for an
                                                                absolute arg)

target.startsWith(claudeRoot + sep)  ->  false
```

macOS symlinks `/var` to `/private/var`, so this is every macOS host. Fail-closed, so not exploitable — legitimate `.claude` writes just stop being auto-approved and fall through to the human gate.

Secondary effect: the same mismatch makes the symlink-component walk unreachable (`hasSymlinkComponent()` returns early when the target is not under the root), so the dangling-symlink guard is dead code on macOS rather than defense in depth.

**This already fails in-tree:** `tests/unit/hooks/hooks.test.ts:154` fails on macOS at `a15baad`. CI stays green because `/tmp` on Linux is a real directory, so base and target already agree.

```
$ npm test
Test Files  1 failed | 115 passed | 1 skipped (117)
     Tests  1 failed | 1952 passed | 3 skipped (1956)
```

## 2. `cortextos init` reports a pre-push gate that was not installed

`src/cli/init.ts:225` runs `setup-hooks.sh` with `stdio: 'ignore'` and prints `Installed git pre-push hook (build + test gate)` whenever it exits 0.

`setup-hooks.sh` is deliberately non-clobbering (#704 / 5a0882d) and exits 0 in all three outcomes — installed, already installed, and skipped-because-a-different-hook-exists. It prints an accurate line for each; `init` discards it.

On any machine that already has a `pre-push` hook (a dotfiles hook, a local leak-guard), `init` asserts a build+test gate is active when it is not. A false "your security control is installed" is worse than silence: it stops the operator from checking. This happened on my machine — `.git/hooks/pre-push` was an existing symlink, so the gate never ran and I did not find out until I tested it explicitly.

## 3. Orchestrator template is missing its `Stop` hook

`templates/orchestrator/.claude/settings.json` has no `Stop` block, so it never wires `cortextos bus hook-idle-flag`.

`doctor`'s drift check defaults an agent's comparison template to `'agent'` (`src/cli/doctor.ts:334`) unless `config.json` records one, and `templates/agent` *does* define `Stop`. So every orchestrator scaffolded from this template fails `cortextos doctor` from birth:

```
[FAIL] Template drift: <org>/<orchestrator>: Missing hook block(s): Stop
```

`templates/agent` and `templates/analyst` already have it; `orchestrator` is the only gap.

PR follows.

johnreitano:
Closes #852 (items 1 and 2 — see the note on item 3 below).

Two independent fixes, one commit each. Rebased onto current `main`.

### 1. `fix(hooks)` — canonicalize the target before the `.claude` containment check

`isClaudeDirOperation()` canonicalized the agent dir but compared it against a lexically-resolved target. When the install path contains a symlink the two live in different namespaces and every comparison fails:

```
base            /var/folders/xy/T/hookperm-abc/agent
canonAgentDir   /private/var/folders/xy/T/hookperm-abc/agent   (realpath)
claudeRoot      /private/var/.../agent/.claude
file_path       /var/folders/.../agent/.claude/settings.json   (absolute)
target          /var/folders/.../agent/.claude/settings.json   (resolve() ignores
                                                                the base for an
                                                                absolute arg)
target.startsWith(claudeRoot + sep)  ->  false
```

That is every macOS host, where `/var` is a symlink to `/private/var`. Fail-closed, so not exploitable — legitimate `.claude` writes simply stopped being auto-approved and fell through to the human gate. The same mismatch also made the symlink-component walk unreachable (`hasSymlinkComponent()` returns early when the target is not under the root), so the dangling-symlink guard was dead code on macOS rather than defense in depth.

**This fixes an existing in-tree failure.** `tests/unit/hooks/hooks.test.ts:154` fails on macOS at `a15baad`; CI stays green because `/tmp` on Linux is a real directory, so base and target already agree.

No test changes were needed — the dangling-symlink and symlinked-ancestor cases are already pinned at `:196` and `:211`. Worth noting for reviewers that those two pass *both before and after*: with the bug, containment rejected everything, so a `false` expectation was satisfied for the wrong reason. The `:154` sanity assertion is the only one that distinguishes the states.

### 2. `fix(init)` — do not report a pre-push gate that was not installed

`init` ran `setup-hooks.sh` with `stdio: 'ignore'` and printed "Installed git pre-push hook" on exit code alone. The script is non-clobbering (#704) and exits 0 in all three outcomes — installed, already installed, and skipped-because-a-different-hook-exists — printing an accurate line for each, which `init` discarded. On any machine with a pre-existing hook it therefore claimed a build+test gate was active when it was not. Now it surfaces the script's own outcome and warns explicitly when the gate is absent.

### Note on the orchestrator `Stop` hook (item 3 of #852)

Dropped from this PR. #725 already adds the byte-identical `Stop` block to `templates/orchestrator/.claude/settings.json`, and it is the older PR with green CI — merging both would conflict. Item 3 of #852 is resolved by #725; no action needed here.

## Verification

```
$ npm run build
Build success

$ npm test
Test Files  116 passed | 1 skipped (117)
     Tests  1953 passed | 3 skipped (1956)
```

Green with no bypass. Before this branch the same command reported `1 failed | 1952 passed` — the macOS failure in (1).

Also confirmed against a live install: `cortextos init` now prints `Skipped: .git/hooks/pre-push already exists (leaving your hook in place)` plus a "gate is NOT active" warning on a machine with a pre-existing hook, where it previously claimed success.

### DIFF
--- a/src/cli/init.ts
+++ b/src/cli/init.ts
@@ -222,8 +222,29 @@ export const initCommand = new Command('init')
         existsSync(join(projectRoot, '.git')) &&
         existsSync(join(projectRoot, 'scripts', 'setup-hooks.sh'))) {
       try {
-        execSync('bash scripts/setup-hooks.sh', { cwd: projectRoot, stdio: 'ignore' });
-        console.log('  Installed git pre-push hook (build + test gate)');
+        // Capture stdout rather than discarding it: setup-hooks.sh is
+        // non-clobbering and exits 0 whether it installed the hook, found it
+        // already installed, or LEFT AN EXISTING HOOK IN PLACE. Reporting
+        // "Installed" on exit code alone claimed a build+test gate was active
+        // when a pre-existing hook (e.g. a dotfiles pre-push) meant it was not
+        // — a false assurance about a security control, which is worse than
+        // saying nothing because it stops the operator from looking.
+        const out = execSync('bash scripts/setup-hooks.sh', {
+          cwd: projectRoot,
+          stdio: ['ignore', 'pipe', 'ignore'],
+        }).toString();
+
+        if (/^\s*Skipped:/m.test(out)) {
+          const line = out.split('\n').find(l => /^\s*Skipped:/.test(l))?.trim();
+          console.log(`  ${line}`);
+          console.log('  WARNING: the build + test pre-push gate is NOT active in this clone.');
+          console.log('           Chain it from your own hook, or install it with:');
+          console.log('             bash scripts/setup-hooks.sh   (after moving your hook aside)');
+        } else if (/^\s*Already installed:/m.test(out)) {
+          console.log('  Git pre-push hook already installed (build + test gate)');
+        } else {
+          console.log('  Installed git pre-push hook (build + test gate)');
+        }
       } catch {
         console.log('  Skipped git pre-push hook install (run: bash scripts/setup-hooks.sh)');
       }
--- a/src/hooks/index.ts
+++ b/src/hooks/index.ts
@@ -252,18 +252,35 @@ export function isClaudeDirOperation(
   // Canonicalize the agent dir first (resolves legitimate symlinks on the install
   // path, e.g. /tmp -> /private/tmp), so the .claude subtree below it is the only
   // thing left to vet.
-  const canonAgentDir = canonicalizePath(resolve(base));
+  const origBase = resolve(base);
+  const canonAgentDir = canonicalizePath(origBase);
   const claudeRoot = join(canonAgentDir, '.claude');
-  const target = resolve(canonAgentDir, filePath);
 
-  // Lexical containment within the agent's own .claude/.
+  // Canonicalize the TARGET the same way as the base before comparing. Without
+  // this, a symlink anywhere on the install path defeats the check: on macOS
+  // the agent dir canonicalizes to /private/var/... while an absolute
+  // file_path keeps its /var/... prefix, so containment failed for legitimate
+  // writes and every .claude operation fell through to the human gate.
+  const lexicalTarget = resolve(origBase, filePath);
+  const target = canonicalizePath(lexicalTarget);
+
+  // Containment within the agent's own .claude/.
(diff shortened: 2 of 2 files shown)


johnreitano:
**Updated — scope reduced, force-pushed.** Recording what changed so the diff and the description do not silently disagree with what reviewers saw earlier.

**Was:** three commits — the macOS `.claude` guard, the `init` gate report, and the orchestrator `Stop` hook.
**Now:** two commits, touching `src/hooks/index.ts` and `src/cli/init.ts` only.

1. **Dropped the orchestrator `Stop` hook commit.** #725 already adds the byte-identical block to `templates/orchestrator/.claude/settings.json`. It is the older PR and has green CI, so merging both would have conflicted on that file. I missed the overlap when I opened this. Item 3 of #852 is resolved by #725 — I have noted that on both.

2. **Rebased onto current `main`**, picking up the four commits merged since this branch was cut.

3. **The suite is now green without a bypass.** The earlier revision disclosed a `--no-verify` push, because the pre-push gate failed on the pre-existing macOS bug that commit 1 fixes. Rebased and deduplicated, `npm test` reports `1953 passed / 0 failed` and the push went through the gate normally.

The title and description have been rewritten to match. No change to the substance of either remaining fix.

PR #863, opened 2026-08-01
fix(daemon): resolve symlinks and dots when locating agent conversation history

### LINKED ISSUE #862: Agents silently start fresh sessions when the agent dir is a symlink (or contains a dot): shouldContinue() mis-resolves the Claude project dir
## Summary

`shouldContinue()` (`src/daemon/agent-process.ts`) decides whether the daemon passes `--continue`. It derives the Claude Code project-history directory from the raw launch path:

```js
const launchDir = this.config.working_directory || this.env.agentDir;
const convDir = join(homedir(), '.claude', 'projects', launchDir.split(sep).join('-'));
try {
  const files = readdirSync(convDir);
  return files.some(f => f.endsWith('.jsonl'));
} catch {
  return false;   // <-- silent
}
```

Two things make that name wrong, and both fail identically: `readdirSync` throws on a non-existent directory, the `catch` returns `false`, and the agent starts a **fresh** session. Nothing is logged. The agent looks perfectly healthy — you find out only when you notice it has forgotten everything, and it happens again on every restart.

## 1. Symlinks are not resolved

Claude Code keys history by the **resolved** path. Any agent directory reached through a symlink computes a name that never exists:

```
launch path   /Users/j/cortextos/orgs/acme/agents/bob     (orgs -> elsewhere)
computed      -Users-j-cortextos-orgs-acme-agents-bob     0 sessions (missing)
Claude uses   -Users-j-data-orgs-acme-agents-bob          35 sessions
```

Hit on a live 13-agent install after relocating the `orgs/` tree behind a symlink — a reasonable thing to do, since this repo gitignores `orgs/` wholesale, so backing it up means putting it in its own repo. Every agent restarted into a fresh session; the daemon had intended to continue the whole time.

## 2. Dots are not replaced

Claude Code also mangles `.` to `-`. Separator-only mangling breaks any dotted component, including agents launched under `~/.cortextos/<instance>/...` — e.g. an orchestrator's worker sessions:

```
computed      -Users-j-.cortextos-default-state-boss-workers-domain
Claude uses   -Users-j--cortextos-default-state-boss-workers-domain
```

You can see both forms side by side in a real `~/.claude/projects` listing:

```
-Users-jreitano--cortextos-default-state-hsom-orchestrator-workers-domain
-Users-jreitano-cortextos-orgs-leadio-agents-analyst
```

## Reproduce

```bash
mv ~/cortextos/orgs /somewhere/else
ln -s /somewhere/else ~/cortextos/orgs
pm2 restart cortextos-daemon
grep "Starting in" ~/.pm2/logs/cortextos-daemon-out.log   # every agent: "fresh mode"
```

## Suggested fix

`realpathSync()` the launch dir (best-effort — a dir that does not exist yet should keep its literal spelling) and replace dots as well as separators. Worth noting the `catch` is load-bearing here: a genuinely missing directory and a *mis-computed* directory name are indistinguishable, so a wrong name degrades silently. A debug log when the computed dir is absent would have made this obvious immediately.

PR follows, with tests that fail against the current implementation.

johnreitano:
Closes #862.

`shouldContinue()` derived the Claude Code project-history directory from the raw launch path, so it computed a name that does not exist whenever the agent directory is reached through a symlink, or contains a dotted component. `readdirSync` throws, the `catch` returns `false`, and the agent starts a **fresh** session with nothing logged — it looks healthy while having lost its conversation, on every restart.

Observed on a live 13-agent install after relocating the `orgs/` tree behind a symlink: every agent restarted fresh, though the daemon had intended to continue.

Extracted into `claudeProjectDir()`, which:

- `realpathSync()`es best-effort — a launch dir that does not exist yet keeps its literal spelling, preserving current behaviour for that case
- replaces dots as well as separators, matching how Claude Code actually names these directories

## Tests

Four new cases in `tests/unit/daemon/claude-project-dir.test.ts`. I verified they are meaningful by reverting the helper to the previous one-line implementation:

```
against the old implementation:   Tests  2 failed | 2 passed (4)
against this change:              Tests  4 passed (4)
```

The two that fail are the symlink case and the dot case.

## Verification

```
$ npm run build
Build success

$ npm test
Test Files  1 failed | 116 passed | 1 skipped (118)
     Tests  1 failed | 1956 passed | 3 skipped (1960)
```

**Disclosure:** pushed with `--no-verify`. The single failure is `tests/unit/hooks/hooks.test.ts` — a pre-existing macOS-only bug on `main` (#852, fixed by #853), unrelated to this branch, which touches only `src/daemon/agent-process.ts` and adds a test file. With #853 applied the suite is fully green.

## Aside for reviewers

The `catch { return false }` is load-bearing: a genuinely missing history directory and a *mis-computed* directory name are indistinguishable, which is why both bugs degraded silently for so long. A debug log when the computed dir is absent would surface this class of problem immediately. Happy to add it here if wanted.

### DIFF
--- a/src/daemon/agent-process.ts
+++ b/src/daemon/agent-process.ts
@@ -1,4 +1,4 @@
-import { appendFileSync, existsSync, readFileSync, statSync, unlinkSync, writeFileSync } from 'fs';
+import { appendFileSync, existsSync, readFileSync, realpathSync, statSync, unlinkSync, writeFileSync } from 'fs';
 import { join, sep } from 'path';
 import { homedir } from 'os';
 import type { AgentConfig, AgentStatus, CtxEnv } from '../types/index.js';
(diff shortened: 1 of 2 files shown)


johnreitano:
**Force-pushed — removed a file that should never have been in this PR.**

The first revision included `knowledge-base/chromadb/chroma.sqlite3` (188 KB binary). That is a local ChromaDB store generated at runtime, not source. It got swept in by a careless `git add -A` on my side and has nothing to do with this fix.

The diff is now the two files it should always have been:

- `src/daemon/agent-process.ts`
- `tests/unit/daemon/claude-project-dir.test.ts`

No change to the fix itself or to the tests. Apologies for the noise.

Two things possibly worth a separate look, since I hit them both: `knowledge-base/chromadb/` is not in `.gitignore`, so it is easy to commit by accident from a working install; and this branch's CI will stay red until #853 lands, because it is based on `main` where `tests/unit/hooks/hooks.test.ts` fails on macOS (#852). That failure is unrelated to this change — this branch touches only the daemon and adds a test file.

PR #874, opened 2026-08-04
fix(daemon): don't count a missing agent binary as an agent crash

asachs01:
## Summary

An agent binary that can't be exec'd was being counted as an agent crash, which walked agents to `max_crashes_per_day` and HALTED them — for a condition no restart could fix, and which heals by itself.

Every agent runs under its own `CLAUDE_CONFIG_DIR`, so each Claude Code instance believes it's a standalone install and independently schedules its own auto-update. But all agents share **one** binary (`~/.local/bin/claude -> versions/<version>`): N private updaters, one shared file, no lock.

On a 9-agent fleet two updaters fired 250ms apart, both reported `install_failed`, and left the symlink dangling at an already-deleted version for ~12 minutes:

```
ruby:  14:05:30.564Z  2.1.220 → 2.1.221  install_failed
pearl: 14:05:30.812Z  2.1.220 → 2.1.221  install_failed
```

node-pty hands back a pid for a dangling symlink, and the child then exits 1 having written **zero** bytes. Verified directly against a dangling symlink:

```
pid assigned: 69035
exitCode: 1  signal: 0
output bytes: 0 ""
```

That's byte-for-byte what the daemon logs — `Running (pid: N)` → `Exited with code 1 signal 0`, nothing in `stdout.log`. Indistinguishable from a crash at that layer, so the daemon charged the daily budget with exponential backoff. Two agents hit the cap and HALTED. Only agents that *restarted* during the window were affected, so the fleet looked half-healthy and it read as agent-specific. It had also fired silently ~13h earlier and gone unnoticed because the hang-detector happened to rescue it.

**Prevention** — `getBaseEnv()` sets `DISABLE_AUTOUPDATER=1`. Updating the runtime is an operator action taken when the fleet is quiet, not something N unsupervised agents attempt against one shared file.

**Resilience** — `handleExit()` gains a binary-unavailable exemption beside the existing image-poison block. Exit code 1 plus a runtime that isn't executable on `PATH` right now is retried *without* charging `max_crashes_per_day`, logged as `BINARY_UNAVAILABLE_RECOVERY`. The reasoning for those two conditions being sufficient: if the binary can't be exec'd, a restart cannot succeed regardless of why the process exited, so charging the budget only guarantees a permanent HALT instead of recovery when the binary returns. This also covers any other cause of a vanished runtime — botched upgrade, unmounted volume, bad `PATH`.

Retry cadence is two-tier on outage age: 30s for the first 15 minutes (real install windows are minutes, and a failed exec costs ~1ms so polling is free), then 5min — that long an outage is a broken runtime needing a human, and the slower tier bounds `restarts.log` growth without ever giving up, since nothing else would bring the agent back. Age comes from timestamps rather than an attempt counter, so no reset has to stay in sync with `start()`.

`agentBinaryName()` mirrors the runtime switch in `start()` and covers all four runtimes (claude / hermes / opencode / codex).

## Test plan

- `npm run typecheck` — clean
- `npm run build` — clean
- `npx vitest run tests/unit/daemon/ tests/unit/pty/` — **508/508 passing**
- New tests: exemption doesn't touch `.crash_count_today`; an ordinary crash with the binary present still counts (regression guard); a clean `exit(0)` is not exempted even while the binary is missing; retry tiers fast → slow → reset-on-new-outage
- **Mutation-tested**: disabling the exemption (`if (false)`) makes the primary test fail, so it isn't a vacuous assertion
- The `pid/exit-1/zero-bytes` signature above was reproduced against a real dangling symlink via node-pty before writing the fix

## Checklist

- [x] `npm run build` passes
- [x] `npm test` passes (daemon + pty suites; 508/508)
- [x] No new secrets or credentials committed
- [x] **Agent Awareness:** no command/endpoint/hook surface change — daemon-internal crash classification plus one env var on the agent PTY
- [x] **Migration Parity:** no agent-installed files change; existing agents pick both halves up on next daemon restart

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,56 @@
 
 ## [Unreleased]
 
+### Fixed — a missing agent binary was counted as an agent crash, halting agents
+
+Every agent runs under its own `CLAUDE_CONFIG_DIR`, so each Claude Code instance
+believes it is a standalone install and independently schedules its own
+auto-update — but all agents share ONE binary
+(`~/.local/bin/claude -> versions/<version>`). N private updaters, one shared
+file, no lock.
+
+Observed on a 9-agent fleet: two updaters fired 250ms apart, both reported
+`install_failed`, and left the symlink dangling at an already-deleted version
+for ~12 minutes. node-pty hands back a pid for a dangling symlink, and the child
+then exits 1 having written zero bytes — verified directly against one:
+
+```
+pid assigned: 69035
+exitCode: 1  signal: 0
+output bytes: 0 ""
+```
+
+The daemon could not distinguish that from an agent crash, so it charged the
+daily crash budget with exponential backoff. Two agents walked to
+`max_crashes_per_day` and HALTED — for a condition no restart could fix, which
+healed on its own the moment the binary reappeared. Only agents that RESTARTED
+during the window were affected, so the fleet looked half-healthy and the
+failure masqueraded as agent-specific.
+
+- **`src/pty/agent-pty.ts`** — `getBaseEnv()` now sets `DISABLE_AUTOUPDATER=1`
+  for every agent PTY. Updating the runtime is an operator action taken when the
+  fleet is quiet, not something N unsupervised agents each attempt against one
+  shared file. *(Prevention.)*
+- **`src/pty/agent-pty.ts`** — new exported `isBinaryAvailable(binary)`:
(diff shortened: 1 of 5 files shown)


drzachconner:
Thanks for this contribution. This pull request currently has merge conflicts with current main, so we cannot complete the review in its present state. Could you please rebase onto current main and resolve the conflicts? Once it is clean, we will take another look.

PR #889, opened 2026-08-08
feat(autoresearch): G1 zero-delta gate in evaluate-experiment

Amoasii:
## What

Enforces the autoresearch skill's **zero-delta rule** in the CLI so agents cannot chain keep/discard cycles on a dead surface (theta 58, G1).

When an `evaluate-experiment` call is the **3rd consecutive flat (non-improving) window on the same surface**, the gate:
- forces a `discard` decision,
- records a tamper-proof re-hypothesize note in the experiment `learning` and `learnings.md`,
- prints a warning banner to stderr.

A genuine improvement (`keep`) resets the streak — the rule is explicitly about not chaining keeps on flat results.

## Changes
- `src/bus/experiment.ts`: `ZERO_DELTA_GATE_LIMIT` (=3), `countConsecutiveFlatOnSurface()`, optional `guardrail_note` field (back-compat), gate logic in `evaluateExperiment`.
- `src/cli/bus.ts`: surface `guardrail_note` to stderr on `evaluate-experiment`.
- `tests/sprint3-experiments.test.ts`: 3 cases — 3rd flat flags, cross-surface no-accumulate, keep resets streak.

## Verification
- `npm run build` clean.
- `npm test` green: 106 files, 1723 passed / 1 skipped.

## Scope
Part of theta 58 autoresearch CLI guardrails (task_1786212275853). G3 (baseline drift detect+warn) lands in a follow-up PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #890, opened 2026-08-08
feat(autoresearch): G3 baseline-drift detection + baseline suggestion

Amoasii:
## What

Second half of theta 58 autoresearch CLI guardrails (**G3**). Detect-and-warn only — **never auto-mutates the baseline**, so a real regression stays distinguishable from a drifted baseline (analyst directive).

### create-experiment
- New `--baseline <n>` to set an explicit starting baseline (default 0).
- When the surface has prior results and no `--baseline` is given, prints a **suggested baseline = median of the last 5 results** to stderr (stdout stays the bare id for capture). Median (not min) resists outliers; the agent may override.

### evaluate-experiment
- Warns when the current result plus recent windows cluster within `DRIFT_DELTA` (=1) of the baseline for `DRIFT_WINDOW` (=5)+ in a row — the baseline is pegged and no longer discriminating. Advisory note in `learning` + `learnings.md` + stderr; **decision and baseline unchanged**.

## Changes
- `src/bus/experiment.ts`: `DRIFT_DELTA`, `DRIFT_WINDOW`, `median()`, `suggestBaselineFromSurface()`, `driftRunLength()`, optional `drift_note` field, `ExperimentCreateOptions.baseline`.
- `src/cli/bus.ts`: `--baseline` option + suggestion print; `drift_note` to stderr on evaluate.
- `tests/sprint3-experiments.test.ts`: explicit baseline, median suggestion (null when none), drift warn at 5, no warn at 4, out-of-band result breaks the run.

## Verification
- `npm run build` clean.
- `npm test` green: 106 files, 1728 passed / 1 skipped.

## Note
Stacks on #889 (G1) — **merge #889 first**; this branch contains the G1 commit too. Completes theta 58 task_1786212275853.


Amoasii:
Superseded by #903 — recreated clean off upstream/main per develop process improvement 2026-08-12 (original carried the full ~485-file fleet-branch divergence, making it CONFLICTING and un-reviewable; #903 is the same 3 real files, MERGEABLE).

PR #896, opened 2026-08-11
fix(cron): persist next_fire_at to stop restart drift (theta 61)

Amoasii:
## Problem

Daemon-managed crons that have **never fired successfully** re-anchor their next fire to `restart_time + interval` on every daemon restart, drifting up to one full interval. Weekly (168h) crons are worst-hit — a restart can skip a whole cadence. Confirmed 2026-08-11: designer's `experiment-user-acceptance` (168h, never fired) slid **+36h** across two restarts.

## Root cause

`loadCrons()` computes `referenceMs` from `last_fired_at` / `last_fire_attempted_at` / cron-state; for a never-fired cron all are absent, so it falls back to `now` → `next_fire_at = restart_time + interval`, sliding forward every restart.

## Fix

Persist `next_fire_at` in crons.json and treat it as the authoritative future anchor on cold-start load.

- **types**: optional `next_fire_at` on `CronDefinition` (backwards-compatible; absent = compute from scratch).
- **loadCrons**: cold-start-only branch honouring a *future* persisted value; a stale (past) value falls through to the existing catch-up + `last_fire_attempted_at` path. Skipped on reload — reload already preserves unchanged crons via the changeKey guard, and a modified cron must recompute (so a schedule change isn't pinned to a stale anchor).
- **write sites**: post-fire (Site 1) and fresh-schedule (Site 2, *future anchors only* — catch-up `now` values are left to the imminent fire's Site-1 write). Both non-fatal on write failure.

## Tests (6, per the design doc)
no-drift across restart · fired-cron preserved · stale→catch-up · persisted-on-every-fire · write-failure-safe · schedule-change-recomputes.

## Verification
- `npm run build` clean.
- `npm test` green: 106 files, 1726 passed / 1 skipped. (One pre-existing retry test's exact `updateCron` count updated +1 for the new future-anchor persist, with an explanatory comment.)

## Rollout
Zero migration — optional field, null on existing files; fleet self-populates on first fire/load. Implements `docs/architecture/cron-persistent-next-fire.md`. Independent of #889/#890. capitan owns final review.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #901, opened 2026-08-12
fix(cron): persist next_fire_at to stop restart drift (theta 61)

Amoasii:
Clean recreation of #896 — branched off `upstream/main` so the diff is only the 4 real files, not the ~485-file fleet-branch divergence that made the original CONFLICTING.

## Problem
On daemon restart, never-fired (or crashed-before-first-fire) crons re-anchor to `restart_time + interval` every time, so a 12h cron can drift up to +12h per restart.

## Fix
Persist `next_fire_at` (ISO-8601 UTC) in `crons.json`. `loadCrons()` treats a present, still-future value as the authoritative anchor; a past value falls through to the existing catch-up policy. Written after each fire and on fresh schedule. Backwards-compatible: absent = compute-from-scratch (legacy).

Cold-start-only (`!isReload`) guard so a config reload that changes a schedule doesn't honor a stale-but-future persisted anchor.

## Files
- `src/daemon/cron-scheduler.ts` — persist + honor `next_fire_at`
- `src/types/index.ts` — `next_fire_at?: string | null` on `CronDefinition`
- `tests/unit/daemon/cron-scheduler.test.ts` — coverage (34/34 pass)
- `docs/architecture/cron-persistent-next-fire.md` — design (force-added past `docs/.gitignore`)

Supersedes #896.

### DIFF
--- a/docs/architecture/cron-persistent-next-fire.md
+++ b/docs/architecture/cron-persistent-next-fire.md
@@ -1,0 +1,201 @@
+# Cron Scheduler — Persistent `next_fire_at`
+
+**Status:** Design — not yet implemented
+**Priority:** Medium (silent bug affects weekly/daily crons; short crons unharmed)
+**Owner:** develop (implementation), analyst (design)
+**Last updated:** 2026-08-11
+**Related:** theta 61, `project_cron_reanchor_drift`
+
+---
+
+## Problem
+
+Daemon-managed crons drift on restart when they have never fired successfully, or when they were crashed mid-fire. The drift equals up to one full interval.
+
+Observed 2026-08-11 (designer, theta 60 aftermath):
+- Cron `experiment-user-acceptance`, schedule `168h`, last_fired_at = `-` (never fired)
+- Pre-restart next fire: `2026-08-17 06:09 UTC`
+- After config-reload cascade at 11:59:14Z + daemon reload: next fire slipped to `2026-08-18 12:02 UTC` (+30h)
+- After another restart at ~12:02Z: next fire slipped to `2026-08-18 18:02 UTC` (+36h from original)
+
+Each restart re-anchors next fire to `restart_time + interval`.
+
+## Root cause
+
+`src/daemon/cron-scheduler.ts:392–397`:
+
+```typescript
+const stateFire = stateLastFireByName.get(def.name);
+const candidates: number[] = [];
+if (def.last_fired_at) candidates.push(new Date(def.last_fired_at).getTime());
+if (def.last_fire_attempted_at) candidates.push(new Date(def.last_fire_attempted_at).getTime());
+if (stateFire) candidates.push(new Date(stateFire).getTime());
+const referenceMs = candidates.length > 0 ? Math.max(...candidates) : now;
+
+let nextFireAt = computeNextFireAt(def, referenceMs);
+```
+
+For a cron with **no** `last_fired_at`, no `last_fire_attempted_at`, and no cron-state entry, the `candidates` array is empty and `referenceMs` falls back to `now` (i.e. daemon-restart time). Result: `next_fire_at = restart_time + interval`, sliding forward on every restart.
+
+The `last_fire_attempted_at` field mitigates crash-mid-fire (avoids double-firing the same slot), but when the daemon crashes before ANY successful fire, `attempted_at` becomes an anchor for the same drift.
+
+## Blast radius (by cron interval)
+
+| Interval    | Drift per restart | Practical impact                                   |
+|-------------|-------------------|----------------------------------------------------|
+| 10m         | ≤ 10 min          | Invisible — restarts settle within one tick        |
+| 4h          | ≤ 4 h             | Barely noticeable; heartbeats catch up             |
+| 12h         | ≤ 12 h            | Autoresearch windows slip by half a day            |
+| 24h         | ≤ 24 h            | Nightly reports miss a day                         |
+| **168h**    | **≤ 7 days**      | **Weekly cycles can skip a whole cadence**         |
+
+## Proposed fix
+
+Persist `next_fire_at` in crons.json whenever it is computed, and treat it as the authoritative future anchor on load.
+
+### Data model change
+
+Add optional `next_fire_at: string | null` (ISO-8601) to `CronDefinition` in `src/types/index.ts`. Backwards-compatible: absent field means "compute from scratch" (current behavior).
+
+### Load path change (`loadCrons()`)
+
+Insert a branch before the current referenceMs computation at cron-scheduler.ts:392:
+
+```typescript
+// PERSISTED next_fire_at — authoritative when present and future.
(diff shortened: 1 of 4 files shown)


drzachconner (changes requested):
# PR #901 Review — `fix(cron): persist next_fire_at to stop restart drift (theta 61)`

## Verdict: DO-NOT-MERGE

The normal restart-drift case is implemented correctly: a future `next_fire_at` is persisted on initial scheduling and after a successful fire, and a cold start reuses that timestamp. However, two cold-start paths are incorrect and can skip or mistime live fleet crons. The first directly contradicts the PR's stated catch-up behavior and was reproduced with a focused test.

Reviewed commit: `7fe9a28b092d1d755cb1ccfb7cfa89a92fb5d29f` on `fleet/cron-persist-next-fire-clean`, against merge base `86729341012dc7b2865f0e21c968ce1fa564c4cc` (`origin/main`). The complete four-file diff was reviewed.

## Findings

### High — A stale persisted due time does not catch up for a never-fired cron

`src/daemon/cron-scheduler.ts:397-425` recognizes only a future persisted timestamp. When `next_fire_at <= now`, it discards that timestamp and falls through. If the cron has never fired, `last_fired_at`, `last_fire_attempted_at`, and `cron-state.json.last_fire` are all absent, so line 423 selects `now` as the reference and line 425 schedules `now + interval`. The catch-up check at lines 434-442 therefore never sees an overdue time.

Concrete example: a never-fired weekly cron persists August 19 as its due time; the daemon is down over that time and starts August 20. PR #901 schedules it for August 27 instead of firing once immediately. This can skip an entire weekly cadence—the exact class of defect the PR aims to fix.

A temporary review-only Vitest probe reproduced this result:

```text
FAIL  tests/unit/daemon/cron-scheduler.theta61-review.test.ts > PR 901 review probe > catch-up fires a never-fired cron whose persisted next_fire_at passed while stopped
AssertionError: expected "vi.fn()" to be called 1 times, but got 0 times

 Test Files  1 failed (1)
      Tests  1 failed (1)
```

The probe file was removed after execution; the PR checkout is clean. The added PR test at `tests/unit/daemon/cron-scheduler.test.ts:857-870` does not cover this case because it supplies an old `last_fired_at`; that field alone drives the existing catch-up path, so the test does not demonstrate that stale `next_fire_at` works for a never-fired cron.

Required change: when a valid persisted timestamp is stale and there is no later successful/attempted fire record that suppresses it, carry the persisted due time into the existing one-shot catch-up clamp instead of recomputing from `now`. Add the never-fired/stopped-past-due regression test.

### High — A schedule changed while the daemon is down retains the old schedule's future anchor

`src/daemon/cron-scheduler.ts:397-407` treats any future `next_fire_at` as authoritative on cold start, but the stored anchor has no schedule/version key. Both update paths preserve unrelated fields: `src/bus/crons.ts:250-265` shallow-merges patches, while the CLI (`src/cli/bus.ts:2105-2124`) and IPC handler (`src/daemon/ipc-server.ts:426-442`) update `schedule` without clearing `next_fire_at`.

If an operator changes a weekly cron to hourly while the daemon is stopped (or its reload signal cannot be delivered), restart still waits until the old weekly anchor—potentially almost seven days. Changing hourly to weekly can instead cause one unexpectedly early fire. This is a new behavior introduced by trusting the persisted value.

The PR test at `tests/unit/daemon/cron-scheduler.test.ts:898-916` covers only `reload()` in a live scheduler, where `changeKey` detects the modified schedule; it does not exercise a cold start after an offline update.

Required change: atomically invalidate `next_fire_at` whenever `schedule` changes, or persist enough schedule identity with the anchor to validate it at cold start. Add CLI/IPC/store coverage plus a fresh-scheduler regression test.

## Correctness and scope evidence

- The intended future-anchor path is sound for an unchanged cron restarted before its due time: initial load writes the computed timestamp, post-success writes it in the same fire-state update, and cold start reuses it.
- Post-success computation reuses the same `next` value for disk and memory (`src/daemon/cron-scheduler.ts:548-569`), avoiding disk/memory divergence in the success path.
- Existing reload behavior remains unchanged for unchanged `name|schedule` keys; modified schedules recompute while the daemon remains alive.
- No package manifest or lockfile changed, so the PR introduces no runtime dependencies.
- `git diff --check origin/main...HEAD` passed with no whitespace errors.
- The design document still says `Status: Design — not yet implemented`; this should be updated if the implementation is merged.

## Atomic-write discipline and live-store risk

The new writes use the established path rather than direct filesystem writes: `updateCron()` takes the per-agent file lock, rereads and shallow-merges the current definition, calls `writeCrons()`, and `writeCrons()` calls `atomicWriteSync(..., keepBak=true)`. Individual rewrites therefore use temp-file-plus-rename atomicity and preserve the normal `.bak` behavior.

Fleet risk is nevertheless material because this changes the persistent cron store, **not `config.json`**. On first scheduler load after rollout, every enabled cron missing a future anchor causes a separate locked read-modify-write of its agent's `crons.json`; a store with N such crons is rewritten N times, with `.bak` advancing through intermediate migration states. A write failure is logged and leaves the in-memory schedule active, but that cron can drift again after another restart. The two correctness findings above can also persist a wrong future schedule into this live store. Rollout should wait for both fixes and their cold-start tests.

## Gates

### `npm install` — PASS

```text
added 146 packages, and audited 147 packages in 3s

46 packages are looking for funding
  run `npm fund` for details

4 vulnerabilities (1 low, 3 high)

To address all issues, run:
  npm audit fix

Run `npm audit` for details.
```

### `npm run typecheck` — PASS

```text
> cortextos@0.1.1 typecheck
> tsc --noEmit
```

Exit code: 0. TypeScript strict compilation passed.

### `npm test` — FAIL

Verbatim Vitest summary:

```text
 Test Files  11 failed | 106 passed | 4 skipped (121)
      Tests  1 failed | 1872 passed | 58 skipped (1931)
   Start at  22:50:54
   Duration  16.65s (transform 24.68s, setup 0ms, import 29.95s, tests 65.08s, environment 13ms)
```

All 11 failures are missing dashboard packages after the requested root `npm install`: imports of `next/server` fail in eight suites/one test, and imports of `better-sqlite3` fail in three suites. Those packages are declared only in `dashboard/package.json`; the PR does not change either package manifest or lockfile. This appears to be a repository/root-install gate issue rather than a regression introduced by PR #901, but the requested full-suite gate is still formally red.

Focused relevant suites passed:

```text
 Test Files  3 passed (3)
      Tests  78 passed (78)
   Start at  22:52:51
   Duration  294ms (transform 252ms, setup 0ms, import 250ms, tests 262ms, environment 0ms)
```

Those suites were `tests/unit/daemon/cron-scheduler.test.ts`, `tests/unit/bus/crons-io.test.ts`, and `tests/integration/phase5-audit.test.ts`. The PR's scheduler suite alone also passed:

```text
 Test Files  1 passed (1)
      Tests  34 passed (34)
   Start at  22:51:32
   Duration  292ms (transform 134ms, setup 0ms, import 165ms, tests 32ms, environment 0ms)
```

PR #916, opened 2026-08-13
fix: silent failures in auto-commit, credential scanning, and the experiment approval gate

iMoveforwardd:
Four related fixes for failures that report success. Every one of these was found because a job
exited 0 having produced nothing — the common thread is a success signal decoupled from the work.

Each fix carries its reasoning in the commit body; this is a summary.

## 1. `auto-commit` targeted the framework repo and could never see agent changes (`2aa12cd`)

`resolveEnv` synthesises `agentDir` without an existence check, and the code fell back to
`projectRoot` — which is pinned equal to `frameworkRoot`. Since the framework repo gitignores
`orgs/`, agent changes were invisible while the caller's own WIP got staged instead.

`git status --porcelain` also emits repo-root-relative paths, so pointing the command at a
subdirectory made every `existsSync()` guard miss — silently skipping **both** the credential scan
and the 10 MB check while still returning success.

## 2. Credential scanning had three separate holes (`decf2ec`, `3a46542`, `61ebd2c`)

- **Unanchored `sk-`** matched inside ordinary words (`disk-`, `task-`, `risk-`), blocking almost
  every file containing those substrings.
- **Case-sensitivity** meant `API_KEY=`, `TOKEN=`, `PASSWORD=` — the most common real leak shape —
  passed straight through. The existing test only passed because the `.env` *filename* rule fired
  first, which gave false confidence that content scanning worked at all.
- **`\s*=\s*` matched across newlines**, so the value could begin on the *next line* and the match
  consumed the following variable's **name** as the value. A blank-valued template like
  `BOT_TOKEN=\nCHAT_ID=` matched on `BOT_TOKEN=\nCHAT_ID`. Measured across a 389-file corpus this
  produced 19 false positives, all ordinary documents that merely mention a variable name.
- Untracked directories collapse to a single `?? dir/` entry without `-uall`. The directory was
  then staged as one unit, and because `statSync(dir).isFile()` is false, **both** the credential
  scan and the size check were skipped — a new folder containing a key would have been committed
  entirely unscanned.

Also: `src/bus/catalog.ts` held a **second, divergent** copy of the credential regex under a comment
reading "If you change one, change the other." It was changed alone anyway, more than once. This PR
exports the pattern from `system.ts` and imports it — a comment cannot couple two constants.

## 3. The experiment approval gate told a one-shot worker to block (`3a46542`, `61ebd2c`)

The `autoresearch` skill instructed the agent to create an approval and then **block until
approved**. A one-shot worker cannot block: its session ends, the run is lost, and the cron still
exits 0, so the failure is invisible. The gate now records the approval, notifies, and **stops** to
be re-dispatched — with an idempotency check so a re-dispatched run does not create a duplicate
experiment.

The same skill also used an **invalid approval category**, which `process.exit(1)`s before creating
anything, so `APPR_ID=$(...)` captured an empty string and the whole stop-and-resume chain rested on
a command that never once succeeded.

These fixes are applied across all 11 copies of the skill, including the `templates/` and
`community/` copies that **seed new agents** — those had been left behind by earlier partial fixes,
so the defect was repaired for existing agents while still shipping to every new one.

## 4. Failures are now visible

`autoCommit` returns a distinct `not_a_repo` status instead of `clean`, so a broken invocation is
distinguishable from a healthy no-op, and the CLI exits non-zero on it. Previously both produced
identical output and exit 0.

## Testing

`tests/unit/bus/system.test.ts` goes from 22 to 31 tests. Each new test was **mutation-verified** —
the corresponding fix was reverted individually and the test confirmed to fail — because a test that
passes either way documents nothing. One earlier test that passed with its fix reverted was found
this way and rewritten.

Includes regression tests for: repo-root resolution from a subdirectory, `not_a_repo` vs `clean`,
paths with spaces, renames, the `-uall` untracked-directory gap, cross-line credential matching, and
the `*_KEY=` shapes.

## Notes for review

- The `[ \t]` separator (rather than `\s`) around `=` is a deliberate tradeoff: it also stops
  matching values on a continuation line. A credential does not legitimately begin on the line after
  its `=`, but it is a real narrowing and worth a second opinion.
- The scanner does **not** detect YAML, JSON, or PEM/private-key credential shapes. That predates
  this PR and is left alone rather than expanded here.
- Tested on Windows. The broad suite has pre-existing platform-sensitive failures unrelated to these
  changes; a HEAD-vs-branch comparison showed **no test file failing on this branch that does not
  also fail at HEAD**.


iMoveforwardd:
Closing at the author's request - withdrawing the fork this was raised from. The fixes remain available if the maintainer wants them; happy to re-raise from a different arrangement.

PR #929, opened 2026-08-16
fix(daemon): detect silent agent dormancy on `cortextos status` with per-agent cadence threshold

grandamenium:
Detects a silent/dormant enabled agent (heartbeat stale beyond an additive, per-agent-cadence-derived threshold) and surfaces it on `cortextos status`. Threshold = the agent's own heartbeat cadence + a fixed jitter margin (additive, not multiplicative). Touches src/utils/dormancy.ts, src/daemon/agent-manager.ts, src/cli/status.ts, src/types/index.ts + tests.

STACKED on #923 (`fix/startagent-idempotent-start`), which is stacked on #895 (`fix/daemon-map-entry-race`) — all three modify agent-manager.ts in a dependency order. Base auto-retargets to main as its parents merge. Do NOT merge before #923 and #895.

### DIFF
--- a/src/cli/status.ts
+++ b/src/cli/status.ts
@@ -95,10 +95,12 @@ export function displayStatuses(statuses: AgentStatus[]): void {
   console.log(separator);
 
   let anyAwaiting = false;
+  let anyDormant = false;
   for (const s of statuses) {
     const name = s.name.padEnd(18);
     let label: string = s.status;
     if (s.awaitingConfirmation) { label = 'unhealthy*'; anyAwaiting = true; }
+    if (s.dormant) { label = 'dormant†'; anyDormant = true; }
     const status = label.padEnd(12);
     const pid = (s.pid?.toString() || '-').padEnd(10);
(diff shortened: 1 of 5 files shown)


grandamenium:
HELD as DRAFT (machine-enforced hold) pending the root heartbeat-refresh fix. This dormancy detector reads heartbeat staleness, but heartbeat is currently spoofable: the daemon refreshes a target agent's last_heartbeat on inbound-message delivery (logEvent side-effect), so a wedged agent that keeps receiving messages never trips this detector — it fails OPEN (proven live with opencode). The root fix (caller-identity opt-in gate on heartbeat refresh, branch fix/heartbeat-refresh-caller-gate) must merge to main FIRST. Sequencing per #929-hold owner (paul): Option A — root-fix merges, then this PR rebases onto the fixed write-path and reads a truthful heartbeat. Kept as draft so it cannot merge out of order. Will be marked ready after the root fix is on main and this is rebased.

PR #979, opened 2026-08-28
Buzz relay adapter: fix reconnect hang and gate WebSocket on unsupported Node

noogalabs:
Two fixes to the existing Buzz (NIP-29 / Nostr) relay adapter, found while running it in production.

### 1. Reconnect hang when the socket closes before AUTH
If the relay or network drops the connection *before* sending the NIP-42 AUTH challenge, `rejectAllPending()` cleared the auth-wait timer but never settled the promise returned by `waitForAuthChallenge()`. `connectAndRun()` then stayed suspended forever, so the reconnect loop never ran again and Buzz inbound went silently offline until the daemon was restarted. The pending auth-wait now carries a reject callback that fires on close. New test: close the socket before the challenge and assert a second connection attempt happens.

### 2. Unguarded `WebSocket` on older Node
Native `WebSocket` is only available by default from Node 22, but the adapter called `new WebSocket()` unguarded. On older Node it threw `ReferenceError` inside the reconnect loop and retried forever while Buzz stayed offline. `start()` now checks for `WebSocket` up front and logs a clear "requires Node 22+" message instead of looping; the CLI throws the same clear error. New test covers the gated path.

All existing Buzz tests still pass, plus the two new ones. Scoped to the adapter — no changes to the transport's design or public surface. Thanks for the adapter; these are just the rough edges we hit putting it under load.

### DIFF
--- a/src/buzz/relay-client.ts
+++ b/src/buzz/relay-client.ts
@@ -62,7 +62,7 @@ export class BuzzRelayClient {
   private eoseSeen = new Set<string>();
 
   private pendingOks = new Map<string, PendingOk>();
-  private pendingAuthChallenge: { resolve: (challenge: string) => void; timer: ReturnType<typeof setTimeout> } | null = null;
+  private pendingAuthChallenge: { resolve: (challenge: string) => void; reject: (reason: string) => void; timer: ReturnType<typeof setTimeout> } | null = null;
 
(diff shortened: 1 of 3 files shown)


grandamenium:
Landed on main via the #991 connector-cluster merge-commit 8ad6c279 (crew-deploy reconciliation). Commits are on main; closing as merged.

